PR #33, opened 2014-10-18
Support expiring ops from oplog

devongovett:
This goes along with https://github.com/share/livedb-mongo/pull/12.  It handles emitting `Missing operations` errors when operations requested from the oplog are not there.  This happens in `submit` and `getOps` (and by extension `fetch`, since it calls `getOps`).

This error should occur if a client has been offline longer than the ttl and tries to submit/fetch some ops.  It should only occur if other edits have taken place while the user was offline, which have also expired.  Specifically:

1. user 1 goes offline
2. user 1 edits
3. user 2 edits
4. ops are dropped (user 2’s ops expire)
5. user 1 comes back online and submits its ops
6. ERROR

If 3 and 4 are swapped (i.e. user 2 doesn’t edit anything before the ops are dropped), the error shouldn't occur. Basically, it should be fairly rare.

When the error happens, ShareJS sends it back to the client, which has an opportunity to notify the user that their changes could not be merged.  Not ideal, but necessary to keep the data size in mongo from growing forever.

One case I thought of but I'm not sure about is whether we need to get the snapshot version in `getOps` when `to` is null to determine whether ops are expected from the oplog.  Right now, the check for missing operations only occurs when there are ops returned (check `ops[0].v === from`).  If `to` is null, we may also need to check if ops were expected if `from < snapshot.v`, which means fetching the snapshot there.  I don't really like that, but it may be necessary.  In `submit` this is handled already.  What do you think?

### DIFF
--- a/lib/redisdriver.js
+++ b/lib/redisdriver.js
@@ -322,7 +322,7 @@ RedisDriver.prototype._oplogGetOps = function(cName, docName, from, to, callback
   this.oplog.getOps(cName, docName, from, to, function(err, ops) {
     if (err) return callback(err);
     if (ops.length && ops[0].v !== from) 
-      return callback('Oplog is returning incorrect ops');
+      return callback('Missing operations');
 
     for (var i = 0; i < ops.length; i++) {
       ops[i].v = from++;
--- a/lib/redisdriver.js
+++ b/lib/redisdriver.js
@@ -172,7 +172,7 @@ RedisDriver.prototype.atomicSubmit = function(cName, docName, opData, options, c
           // In this case, we should write a no-op ramp to the snapshot
           // version, followed by a delete & a create to fill in the missing
           // ops.
-          throw Error('Missing oplog for ' + cName + ' ' + docName);
+          return callback('Missing oplog for ' + cName + ' ' + docName);
         }
         self._redisSubmitScript(cName, docName, opData, dirtyData, version, callbackWrapper);
       });
@@ -321,7 +321,8 @@ RedisDriver.prototype._oplogGetOps = function(cName, docName, from, to, callback
   var self = this;
   this.oplog.getOps(cName, docName, from, to, function(err, ops) {
     if (err) return callback(err);
-    if (ops.length && ops[0].v !== from) throw Error('Oplog is returning incorrect ops');
+    if (ops.length && ops[0].v !== from) 
+      return callback('Oplog is returning incorrect ops');
 
     for (var i = 0; i < ops.length; i++) {
       ops[i].v = from++;
--- a/test/oplog.coffee
+++ b/test/oplog.coffee
@@ -98,15 +98,10 @@ module.exports = (create) ->
         check = (error, ops) ->
           throw new Error error if error
           assert.deepEqual ops, []
-          done() if ++num is 7
+          done() if ++num is 2
 
         @db.getOps @cName, @docName, 0, 0, check
-        @db.getOps @cName, @docName, 0, 1, check
-        @db.getOps @cName, @docName, 0, 10, check
         @db.getOps @cName, @docName, 0, null, check
-        @db.getOps @cName, @docName, 10, 10, check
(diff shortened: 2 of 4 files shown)


devongovett:
Just merged master back into this branch. Any chance of reviewing this and https://github.com/share/livedb-mongo/pull/12 soon? FWIW, we've been running this in production at Storify for a few months now.

nmai:
@devongovett Without the ability to purge historical data, the official LiveDB is not a good option for me unless I had unlimited resources. Are you still using your `expire-ops` branch in production?

devongovett:
Yes, still running this in production.

nmai:
Sweet, I'll start using your fork. It's a shame this feature never got merged.

josephg:
Reviewing? I did review it. Its missing the things I said, and missing tests.

devongovett:
Uhh... I made changes since you reviewed and moved a bunch of stuff into the mongo backend instead of doing it here. There are a few tests here, but the real meat of the logic was moved into the mongo backend, so most of the tests are there.

lib/redisdriver.js
```diff
@@ -172,7 +172,7 @@ RedisDriver.prototype.atomicSubmit = function(cName, docName, opData, options, c
           // In this case, we should write a no-op ramp to the snapshot
           // version, followed by a delete & a create to fill in the missing
           // ops.
-          throw Error('Missing oplog for ' + cName + ' ' + docName);
+          return callback('Missing oplog for ' + cName + ' ' + docName);
```
josephg:
```javascript
return callback('Missing oplog');
```

... so people can match on the error string.

lib/redisdriver.js
```diff
@@ -323,7 +323,8 @@ RedisDriver.prototype._oplogGetOps = function(cName, docName, from, to, callback
   var self = this;
   this.oplog.getOps(cName, docName, from, to, function(err, ops) {
     if (err) return callback(err);
-    if (ops.length && ops[0].v !== from) throw Error('Oplog is returning incorrect ops');
+    if (ops.length && ops[0].v !== from) 
+      return callback('Missing operations');
```
josephg: It might be easier to just call `self.getOps(...)` instead of replicating the logic here.

test/oplog.coffee
```diff
@@ -98,15 +98,10 @@ module.exports = (create) ->
         check = (error, ops) ->
           throw new Error error if error
           assert.deepEqual ops, []
-          done() if ++num is 7
+          done() if ++num is 2
```
josephg: This test doesn't do the thing it says anymore.

test/test.coffee
```diff
@@ -69,6 +69,36 @@ describe 'livedb', ->
               throw new Error err if err
               assert.equal m.language, null
               done()
+              
+    it 'errors when oplog is missing all operations', (done) -> @create =>
```
josephg: Rename the test

test/test.coffee
```diff
@@ -69,6 +69,36 @@ describe 'livedb', ->
...
+        getOps = @client.driver.getOps
+        @client.driver.getOps = (cName, docName, from, to, fn) ->
+          fn null, []
+        
+        @collection.submit @docName, {v:1, op:['test']}, (err, v) =>
+          assert.equal err, 'Missing operations'
+          @client.driver.getOps = getOps
+          done()
+          
+    it 'errors when oplog is missing some operations', (done) -> @create =>
```
josephg: This is bad behaviour. If the oplog throws, that implies something has gone wrong. We shouldn't just silently eat the exception. (Exceptions ruin optimization, they shouldn't ever be used as part of the expected behaviour of a program)

josephg:
Hah - I never took the money out of gittip and a week ago they quietly started refunding it :/

> The code looks good. Can you also add tests to see what happens if you call subscribe at an old version of the document.

nmai:
If a document is deleted, do the ops get purged as well? Not sure if this is implemented in the standard behavior already.

PR #40, opened 2015-01-09
Moving all unit tests from coffee script to javascript.

noansknv:
Moving all unit tests from coffee script to javascript to remove unnecessary complexity and simplifying the project to only one language.

@nateps Pls review whenever you have time. All unit tests are passing (doh).


josephg:
I really appreciate all the work you've done here, but I'm not convinced this is a net win. Coffeescript is a great language for mocha tests, just because of how nested and callbacky they end up being. You can see that in the diff stats - and the difference is *bigger* than I expected: `6,279 additions and 2,561 deletions.`

Moving the main code from coffee->JS makes it so people can read & understand livedb without knowing coffeescript. Moving the tests across finishes the loop, that people can also contribute tests to any changes they make without knowing coffeescript. Is this a serious blocker? Is it enough of a big deal to justify doing twice the typing when we make unit tests? I'm not looking forward to debugging those big chains of `}); }); }); }); });.....`.

Let me know if I'm missing something. Otherwise I'm going to close this PR.

wmertens:
-1, please keep tests coffeescript

PR #47, opened 2015-03-09
Registers rich-text support in livedb.

Bryan-Turek:
Registers rich-text support in livedb. ottypes/rich-text must include .type property on export for this to work.

### DIFF
--- a/lib/ot.js
+++ b/lib/ot.js
@@ -18,7 +18,7 @@ var registerType = exports.registerType = function(type) {
 registerType('ot-json0');
 registerType('ot-text');
 registerType('ot-text-tp2');
-
+registerType('rich-text');
 // Returns an error string on failure. Rockin' it C style.
 exports.checkOpData = function(opData) {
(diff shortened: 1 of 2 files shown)


pedrosanta:
@Bryan-Turek I think ottypes/rich-text already exports a `type` property. Probably you'll need to update `rich-text` dependency to `^2.0.0` for the build to pass, no?

Cheers.

pedrosanta:
I'm not sure if we should register all the available OT types, or leave that decision to the user/coder. At the moment I'm kinda leaning to the last one, but I'm open to arguments.

This all needs better documentation though, some community love on this could be appreciated. :smiley: 

Cheers.

Bryan-Turek:
Yeah, I think I submitted this and the other exporting type property at the same time, so the one was accepted because it worked, but this one did not because the other wasn't accepted yet.

Also it would be great if we allowed the user to select which ones he/she wants to use in the application, but at this time that requires modifying this source? I haven't checked in on this project in a while.

PR #50, opened 2015-05-05
avoid breaking browserify

enjalot:
the dynamic way that ot.js is using requires breaks browserify when requiring livedb directly on the client side for ot functions.


nateps:
addressed in the refactor to sharedb

PR #51, opened 2015-05-06
Prevent oplog error from crashing the server

cray0000:
When you delete the oplog completety it was throwing an error 'Oplog is returning incorrect ops' crashing the server. Instead it should just return a callback with an error crashing only the client.


nateps:
Thanks!

PR #59, opened 2015-10-21
Add 'index' to 'doc' event

zag2art:
We should definitely know 'index' in the 'doc' event for access control purpose. Source collection name is not sufficient.

Also we maybe should do the same change for 'op' event.

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -66,6 +66,7 @@ Backend.prototype.addProjection = function(name, collection, type, fields) {
   }
 
   this.projections[name] = {
+    name: name,
     target: collection,
     type: ot.normalizeType(type),
     fields: fields
@@ -135,7 +136,7 @@ Backend.prototype._sanitizeOpsBulk = function(agent, projection, collection, ops
   }, callback);
 };
 
-Backend.prototype._sanitizeSnapshot = function(agent, projection, collection, id, snapshot, callback) {
+Backend.prototype._sanitizeSnapshot = function(agent, index, projection, collection, id, snapshot, callback) {
   if (projection) {
     try {
       projections.projectSnapshot(projection.fields, snapshot);
@@ -143,18 +144,18 @@ Backend.prototype._sanitizeSnapshot = function(agent, projection, collection, id
       return callback(err);
     }
   }
-  this.trigger('doc', agent, {collection: collection, id: id, snapshot: snapshot}, callback);
+  this.trigger('doc', agent, {index: index, collection: collection, id: id, snapshot: snapshot}, callback);
 };
-Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, snapshots, callback) {
+Backend.prototype._sanitizeSnapshots = function(agent, index, projection, collection, snapshots, callback) {
   var backend = this;
   async.each(snapshots, function(snapshot, eachCb) {
-    backend._sanitizeSnapshot(agent, projection, collection, snapshot.id, snapshot, eachCb);
+    backend._sanitizeSnapshot(agent, index, projection, collection, snapshot.id, snapshot, eachCb);
   }, callback);
 };
-Backend.prototype._sanitizeSnapshotBulk = function(agent, projection, collection, snapshotMap, callback) {
+Backend.prototype._sanitizeSnapshotBulk = function(agent, index, projection, collection, snapshotMap, callback) {
   var backend = this;
   async.forEachOf(snapshotMap, function(snapshot, id, eachCb) {
-    backend._sanitizeSnapshot(agent, projection, collection, id, snapshot, eachCb);
+    backend._sanitizeSnapshot(agent, index, projection, collection, id, snapshot, eachCb);
   }, callback);
 };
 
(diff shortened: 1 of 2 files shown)


ovvn:
+1

zag2art:
I think the PR is still relevant.

nateps:
Sorry for not providing much context here. Was in a rush to get ShareDB out originally and just now trying to catch back up.

I would recommend moving away from projection names as much as possible, which is why I closed this. Projection names are obviously still required now for querying, but I'd rather not expose them in other places so people don't have to change as much code later. In a later major revision (ShareDB 2.0+), I intend to get rid of the entire idea of named projections so that you can pass in a unique projection along with each query.

Alternatively, what I'd recommend is actually checking the fields on the document and ensure that the document does not contain fields that the user should not have access to. For true secrets (like hashed passwords or API keys) I recommend storing them in a different Mongo collection where the id is the same. For example, users.XYZ would have a separate user_secrets.XYZ document so you can fetch them at the same time in parallel when you know the userId. Completely blocking access to a collection is easier to ensure and it is a good signal to other developers on the team that this is stuff to be careful exposing.

Another step we'll be doing soon at lever is adding a middleware that generates the actual query sent to mongo on the server so that users cannot create arbitrary queries from the client. Will write more about this later. Quick example:

```
model.query('tasks', {userId: userId});

**Comment by zag2art** [comment]
> Another step we'll be doing soon at Lever is adding a middleware that generates the actual query sent to mongo on the server so that users cannot create arbitrary queries from the client.

We use the idea in https://github.com/dmapper/server-query

By the way, why don't you make the middleware API synchronous? Most of the hooks should be definitely synchronous. It looks like the async nature of the hooks is from the old sharejs code. Take a look for example at 'connect' hook. If I write something like:
```
backend.use('connect', function(request, next) {
  process.nextThick(next);
});
```
This breaks tests.

Don't see the reason to have all the hooks async: 'commit', 'after submit', 'submit', 'op', 'connect', 'doc', 'query', 'apply'. What do you think?

**Comment by nateps** [comment]
The question about the middleware API is a separate issue. In the future, please open a new issue for a new question so the discussion is easier to follow.

All middleware work whether you call next synchronously or asynchronously. At any of these hooks you might want to fetch additional data to do validation, make sure you have tracked the event in an audit log, etc before calling next. And even if some don't need to be async but some or most do, it is much easier to learn a consistent API.

What tests are broken by adding a next tick in connect?

**Comment by zag2art** [comment]
> In the future, please open a new issue for a new question so the discussion is easier to follow.

Ok

> What tests are broken by adding a next tick in connect?

The 3 sharedb tests:

1. updates after connect and connection.close()
2. subscribed query gets simultaneous insert and remove after reconnecting
3. submits fail above the backend.maxSubmitRetries threshold

**Comment by zag2art** [comment]
> At any of these hooks you might want to fetch additional data to do validation, make sure you have tracked the event in an audit log, etc before calling next. And even if some don't need to be async but some or most do, it is much easier to learn a consistent API.

Actually, the hook i'm interested in is 'query'. Let me try to explain what I try to achieve. 
I'm implementing 'qr' (query resubscribe) message. You already have "qs" and "qu" (query subscribe/unsubscribe) messages. Something like:

```
msg = {
  a: 'qr', 
  id: 5, 
  q: { type: 'new' } // or something
}
```

The main part of implementation is not hard - I'm just adding the new query into queryPoll queue of the corresponding query-emitter. But there is bad situation in case when a client send the two messages one right after another: 'qs' and (without delay) 'qr'. A kind of race condition.

For now, queryEmitter object for a 'qs' message is created at the very end of a chain of async functions. So, in my case, if I send 'qr' message right after the 'qs' in the 'qr'-handler the queryEmitter still doesn't exist.

To solve this I'm trying to refactor 'qs' handling chain in order to create the queryEmitter object synchronous (or at least as early as it possible). A dual nature of the 'query' hook make the refactoring much harder.

Maybe you can suggest another way to solve this or maybe the it's a bad idea to implement the 'qr' event at all?

**Comment by nateps** [comment]
There should be no need for a query resubscribe command, as subscribe already has this optimized. You can pass in a results option that will cause the query to diff against the client's results. This is already done automatically on reconnect. You should definitely not add more complexity here.

**Comment by zag2art** [comment]
But I need to change mongo query. F.e. initially it was {type: 'new', $orderby: {date: -1}}, and in the resubscribe message it's {type: 'old', $orderby: {date: -1}}

**Comment by nateps** [comment]
That is a completely new query, so it does not make sense as a resubscribe.

**Comment by zag2art** [comment]
I want to have liveQueries on the racer level. Something like pathQueries but more powerful. 
For example, I have the object in the model (_page.query):
```
{
  type: 'items'
  $skip: 10,
  $limit: 20
}
```

and the model query object, like

```
q = model.query('items', '_page.query')
model.subscribe(q)
```

and then after a while if I change the path (_page.query) reactively it should resubscribe my query on the sharedb server.

something like this:
```
model.set('_page.query', {type: 'items', $skip: 20, $limit: 20})
```
it should resubscribe the query and give me a diff from the server.

PR #64, opened 2015-12-16
memory queries with mingo

vmakhaev:
Basic implementation of in-memory queries. Tests are passes


vmakhaev:
Gonna close it in favor of [sharedb-memory-mongo](https://github.com/vmakhaev/sharedb-memory-mongo)

PR #66, opened 2016-01-17
Added example for client/server communication through websockets

geakstr:


examples/websockets/client.js
```diff
@@ -0,0 +1,18 @@
+var sharedb = require("sharedb/lib/client");
+
+var connection = new sharedb.Connection(new WebSocket("ws://localhost:8888"));
+
+var doc = connection.get("collection", "document");
+
+doc.subscribe(function(error){
+  if (!doc.type) {
```
geakstr: I'm not sure that this is correct method to verify of doc availability

lubaochuan:
hello geakstr, you example is very helpful. Do you know how to use the client in a browser (like a sharejs client side script)?

geakstr:
@lubaochuan yes, I use sharedb in my project in browser. `client.js` file in this pull request works in browser as well

lubaochuan:
@geakstr Thanks! I used browserify to create a bundle (client.js) of node code for browser use. The client side code runs but it didn't send operations to the server via the websocket correctly (sever.js doesn't print anything when index.html is loaded). Here is my repo for this experiment https://github.com/lubaochuan/sharedb_examples Did I miss something?

geakstr:
@lubaochuan I submit PR into your repo, check this https://github.com/lubaochuan/sharedb_examples/pull/1

examples/websockets/server.js
```diff
@@ -0,0 +1,49 @@
...
+});
+var share = sharedb();
+
+wss.on("connection", function(ws, req) {
+  var stream = new Duplex({
+    objectMode: true
+  });
+
+  stream.headers = ws.upgradeReq.headers;
+  stream.remoteAddress = ws.upgradeReq.connection.remoteAddress;
```
nornagon: Are these (headers / remoteAddress) necessary for the correct functioning of sharedb? Would be nice to add a comment here if so.

examples/websockets/server.js
```diff
@@ -0,0 +1,49 @@
...
+  stream._read = function() {};
+  stream.on("error", function(msg) {
+    ws.close();
+  });
+  stream.on("end", function() {
+    ws.close();
+  });
+
+  ws.on("message", function(op) {
+    // console.log('op', JSON.stringify(JSON.parse(op), null, 2));
```
nornagon: Can't merge this with random commented-out code lying around :)

examples/websockets/client.js
```diff
@@ -0,0 +1,18 @@
...
+    doc.create({x: "hello"});
+  }
+
+  if (doc.type) {
+    // `doc` ready, make ops
+    setTimeout(function() { // timeout need for example (to delay after create)
```
nornagon: When I tried this, this branch didn't hit in the first connection (the first pageload), so I don't think the timeout is necessary.

examples/websockets/client.js
```diff
@@ -0,0 +1,18 @@
...
+      doc.submitOp([{p: ["x", 5], si: " world"}]);
+    }, 1000);
+  }
+});
```
nornagon: Would be great to add a listener for `doc.on('op')` to show how that event works.

(Also, mega nitpick: add newlines at the end of the files!)

nornagon:
This was super helpful for getting started with sharedb. Despite being a member of the share org, I haven't used sharejs in years and it's changed a lot. I'd love to have this or something like it in the documentation. @nateps, what do you think about merging this?

nateps:
Hey, Jeremy! Great to hear from you here, and thanks for jumping in with helpful comments.

Having a simple example is a great idea. Thanks, @geakstr!

I think the server code can be simplified a lot. I imagine there should be a library that will create a server-side stream for a websocket connection. Have you tried https://www.npmjs.com/package/websocket-stream?

If we are not using a library, the best way to create a custom stream is typically to inherit from the Duplex class instead of attaching methods to each stream instance. See the implementation of ServerStream in ShareDB: https://github.com/share/sharedb/blob/master/lib/stream-socket.js.

Please add a package.json file in this directory so people can npm install in the example directory and get up and running quickly!

geakstr:
@nateps Hello. I will explore your links a bit later and update pull request. 
@nornagon Thank you for the comments, I will fix it.

avital:
Hi @geakstr, we're aiming to get an simple websocket example ready for the ShareDB 1.0 release, to be released in a few weeks. Do you think you'll have time to respond to the comments? I'd love to use your example.

geakstr:
@nornagon @nateps @avital sorry for delay, guys, I took note of your comments and updated PR

geakstr:
Ah, I see examples folder already exist in repo, unfortunately I did not make it in time :)

avital:
Yeah :)

PR #68, opened 2016-02-09
Pass index to "doc" middleware

psirenny:
This is helpful if you want to restrict access to a collection but permit access to one of its projections.


nateps:
Thanks for the contribution, @psirenny!

Closed as duplicate of #59 

Please see duplicate issue for explanation of my reasoning. In the long run, we're moving away from named projections, so I'd like to minimize their impact on the API in the meantime.

PR #71, opened 2016-03-24
Initialize op objects with all properties

avital:
This way, op objects won't change form later when properties
are assigned. This makes the objects be optimized as structs
in V8.


nateps:
:-)

nateps:
Actually I just realized the issue with this (and probably why I didn't do it originally) is that this causes Mongo to persist the keys set to null. That could end up taking some more space in Mongo, and we are only adding these properties for pub sub. Instead of this, we should probably copy the appropriate properties from the op and make a new object instead of mutating this object in publish.

PR #72, opened 2016-03-24
Fix flaky pollDebounce tests

avital:
The pollDebounce tests insert 10 documents and make sure
that debouncing leads to two separate chunks of documents
being passed through the 'insert' callback on the query.

The problem was that we weren't waiting for the document
creation to be submitted before sending the next document.

Even though the original test behavior was a more accurate
representation of what a real client would do, it lead to
unpredictable results. The timing of when the op was submitted
could have lead to all of the documents being added in one single
chunk, rather than two.

While at it, made the test more strict -- it now checks the sizes
of the two chunks passed through the 'insert' callback, rather
than just the number of chunks.

### DIFF
--- a/test/client/query-subscribe.js
+++ b/test/client/query-subscribe.js
@@ -214,10 +214,11 @@ describe('client query subscribe', function() {
       }
     });
     function createDoc(count) {
-      connection.get('items', count.toString()).create({});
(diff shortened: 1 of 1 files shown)


nateps:
Yep, looks great. Thanks!

PR #84, opened 2016-04-29
Use sharedb-mingo-memory instead of custom MemoryDB in tests

avital:
See https://github.com/avital/sharedb-mingo-memory/blob/master/index.js for the implementation of the Mingo database adapter

### DIFF
--- a/test/client/query-subscribe.js
+++ b/test/client/query-subscribe.js
@@ -397,7 +397,7 @@ describe('client query subscribe', function() {
   });
 
   it('changing a sorted property moves in a subscribed query', function(done) {
-    if (!makeSortedQuery) this.skip();
+    if (!makeSortedQuery) return this.skip();
 
     var connection = this.backend.connect();
 
--- a/test/client/query-subscribe.js
+++ b/test/client/query-subscribe.js
@@ -2,12 +2,15 @@ var expect = require('expect.js');
 var async = require('async');
 var util = require('../util');
 
-module.exports = function(makeSortedQuery) {
+module.exports = function(options) {
+var getQuery = options.getQuery;
+
 describe('client query subscribe', function() {
+  var nullDbQuery = getQuery({query: {}});
 
   it('creating a document updates a subscribed query', function(done) {
     var connection = this.backend.connect();
-    var query = connection.createSubscribeQuery('dogs', {}, null, function(err) {
+    var query = connection.createSubscribeQuery('dogs', nullDbQuery, null, function(err) {
       if (err) return done(err);
       connection.get('dogs', 'fido').create({age: 3});
     });
@@ -28,7 +31,7 @@ describe('client query subscribe', function() {
       function(cb) { connection.get('dogs', 'spot').create({age: 5}, cb); }
     ], function(err) {
       if (err) return done(err);
-      var query = connection.createSubscribeQuery('dogs', {}, null, function(err) {
+      var query = connection.createSubscribeQuery('dogs', nullDbQuery, null, function(err) {
         if (err) return done(err);
         connection.get('dogs', 'taco').create({age: 2});
       });
@@ -51,7 +54,7 @@ describe('client query subscribe', function() {
       function(cb) { connection.get('dogs', 'spot').create({age: 5}, cb); }
     ], function(err) {
       if (err) return done(err);
-      var query = connection.createSubscribeQuery('dogs', {}, null, function(err) {
+      var query = connection.createSubscribeQuery('dogs', nullDbQuery, null, function(err) {
         if (err) return done(err);
         connection.get('dogs', 'fido').del();
       });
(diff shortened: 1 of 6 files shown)


curran:
Awesome!

zag2art:
Cool, looks very close to https://github.com/derbyparty/sharedb-mingo and https://github.com/vmakhaev/sharedb-memory-mongo

avital:
Yeah, I was trying to use MemoryDB directly and keep the implementation as short as possible.

test/client/query-subscribe.js
```diff
@@ -2,7 +2,7 @@ var expect = require('expect.js');
 var async = require('async');
 var util = require('../util');
 
-module.exports = function() {
+module.exports = function(makeSortedQuery) {
```
nateps: Make this an options object for future proofing. Good chance we'll need other generic query building methods in the future and will be better if passed in as a config object than a bunch of arguments.

```

File: test/client/query-subscribe.js [hunk 2]
```diff
@@ -391,18 +391,24 @@ describe('client query subscribe', function() {
...
     ], function(err) {
       if (err) return done(err);
-      var query = connection.createSubscribeQuery('dogs', {$orderby: {age: 1}}, null, function(err, results) {
-        if (err) return done(err);
-        expect(util.pluck(results, 'id')).eql(['fido', 'spot']);
-        expect(util.pluck(results, 'data')).eql([{age: 3}, {age: 5}]);
-        connection.get('dogs', 'spot').submitOp({p: ['age'], na: -3});
-      });
+      var query = connection.createSubscribeQuery(
+        'dogs', makeSortedQuery({}, [['age', 1]]), null,
```
nateps: this is a bit busy for a single statement. Better to build object on a previous line. Annoying we are using the term `query` to mean two different things (historical nomenclature problem i've never figured out). So sometimes I use the term `expression` to mean the raw object query. You could also call it `sortedQuery` in this context or `dbQuery` if we used that consistently.

```
  var dbQuery = getSortedQuery({}, [['age', 1]]);
  ...
```

test/client/query-subscribe.js
nateps: Actually, `getSortedQuery` is really just the generic query method. We also need to call it for all cases where we don't need a sort. If the query language were SQL, then we'd call it for a simple filtering query. We should update all the tests that make queries to go through the single function so that any format could be implemented. Should probably just use the name `getQuery`

test/client/query-subscribe.js
```diff
@@ -391,18 +391,24 @@ describe('client query subscribe', function() {
   });
 
   it('changing a sorted property moves in a subscribed query', function(done) {
+    if (!makeSortedQuery) return this.skip();
+
     var connection = this.backend.connect();
+
     async.parallel([
       function(cb) { connection.get('dogs', 'fido').create({age: 3}, cb); },
```
avital: call `getQuery` on all queries

avital: wrap `getQuery` so that tests don't try using special Mongo operators

test/client/query-subscribe.js
```diff
@@ -2,12 +2,15 @@ var expect = require('expect.js');
 var async = require('async');
 var util = require('../util');
 
-module.exports = function() {
+module.exports = function(options) {
+var getQuery = options.getQuery;
+
 describe('client query subscribe', function() {
+  var nullDbQuery = getQuery({query: {}});
```
avital: matchAllDbQuery

test/client/query-subscribe.js
```diff
@@ -2,12 +2,15 @@ var expect = require('expect.js');
 var async = require('async');
 var util = require('../util');
 
-module.exports = function() {
+module.exports = function(options) {
+var getQuery = options.getQuery;
```
avital: check getQuery -- allow db adapters that don't support queries

test/client/query.js
```diff
@@ -2,13 +2,16 @@ var expect = require('expect.js');
 var async = require('async');
 var util = require('../util');
 
-module.exports = function() {
+module.exports = function(options) {
+var getQuery = options.getQuery;
+
 describe('client query', function() {
+  var nullDbQuery = getQuery({query: {}});
```
avital: ditto

test/client/query.js
```diff
@@ -2,13 +2,16 @@ var expect = require('expect.js');
 var async = require('async');
 var util = require('../util');
 
-module.exports = function() {
+module.exports = function(options) {
+var getQuery = options.getQuery;
```
avital: ditto

test/db.js
```diff
@@ -3,7 +3,15 @@ var expect = require('expect.js');
 var Backend = require('../lib/backend');
 var ot = require('../lib/ot');
 
-module.exports = function(create) {
+module.exports = function(options) {
+  var create = options.create;
+  function getQuery(queryOptions) {
+    for (var key in queryOptions.query) {
+      if (key[0] === '$') throw new Error('unsupported in tests: ' + key);
+    }
```
avital: also check for dots

PR #85, opened 2016-05-02
Let MemoryDB instances return "extra" from _querySync

avital:
In case like queries with $count, in addition to the array of snapshots
returned from a query, an additional "extra" argument is passed in
the callback to the query method.

This change lets implementors of MemoryDB still only implement
_querySync even when exposing the "extra" field. This is done by
adding an "extra" property to the returned array, which in turn
is stripped out when MemoryDB calls the callback passed to the
query method.

We use this functionality in https://github.com/avital/sharedb-mingo-memory
to implement $count.


avital:
Folding this into #84, which can no longer run without this.

PR #90, opened 2016-07-07
README.md database adapter discrepancy

caleblloyd:
Readme.md references MemoryPubSub as a database adapter, I think it's supposed to be referencing MemoryDB


avital:
Thanks, you're right!

PR #94, opened 2016-07-22
remove op.options implementation; maintain change to options in doc methods

nateps:
Rolls back changes that added options to ops, because this is non trivial to implement and we probably need a different API for it. It won't work as expected if ops are composed. There are perhaps good ways to do this, but punting for now as this is not high priority currently.

Keeping the options format change to create, del, and submitOp doc methods, because having source as an option rather than an argument is more clear and allows for further extension.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -556,29 +556,26 @@ Agent.prototype._submit = function(collection, id, op, callback) {
   });
 };
 
-function CreateOp(src, seq, v, create, options) {
+function CreateOp(src, seq, v, create) {
   this.src = src;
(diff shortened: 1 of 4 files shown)


curran:
It's very exciting to see this great work on ShareDB. Do you have any pointers on how one might recover document content from a given point in time? I'm looking for a way to implement a function that takes a Date as input, and computes as output the content of a given ShareDB document at that point in time. Thank you.

nateps:
Hey curran, can you please ask your question on the mailing list or open a new issue about your question?

PR #69, opened 2016-03-10
Chaining backend.use calls

droganov:
To allow:
backend
    .use( "submit", function( shareRequest, next ){
        console.log( "submit" )
        next()
      }
    )
    .use( "query", function( shareRequest, next ){
        console.log( "query" )
        next()
      }
    )

PR #101, opened 2016-07-26
Chaining with `use` method

matto1990:
A quick PR which adds the missing unit test from #69.

Allows code such as:

```
backend.use("submit", function(shareRequest, next) {
    console.log("submit");
    next();
  })
  .use("query", function(shareRequest, next) {
    console.log("query");
    next();
  });
```


nateps:
Awesome. Thanks!

PR #102, opened 2016-07-26
Add JSHint configuration

matto1990:
I have added the `.jshintrc` file from [DerbyJS](https://github.com/derbyjs/derby) to help catch any obvious small errors and aid with contributions. I have also configured it to run on Travis to catch any errors. It can be run locally using `npm run jshint`. Ideally it would be run automatically with a pre-commit hook or similar.

After adding in JSHint, I fixed some obvious small errors. There are two however which I have left as found, but with ignores and FIXME's next to them. These are:

- Some code in `lib/ot.js` which has "non-obvious" use of `!`
- Usage of a variable out of scope in `lib/agent.js`

Let me know if you would like me to make any alterations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,27 @@
+# OS X
 *.swp
 *.DS_Store
-node_modules
+
+# Logs
+logs
+*.log
+npm-debug.log*
+
+# Runtime data
+pids
+*.pid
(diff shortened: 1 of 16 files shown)


nateps:
Very cool. Thanks!

PR #108, opened 2016-08-04
Feature/custom serialization

DetweilerRyan:
This pull request adds support for custom serialization of ot types as documented [here]( https://github.com/ottypes/docs).

Sharedb is a key component of [Retrium.com](http://retrium.com/demo)’s (a real time meeting collaberation web app) tech stack.  Retrium has been dependent on an out of date fork of an early version of sharedb for a year and that’s not ideal :).

The bulk of the changes required to support custom serialization are in the client/doc.js file as outlined below. 

- **Doc.data**
  - Is always deserialized if the type supports serialization

- **Doc.create**
  - Will accept initial data in its deserialized form if the type supports serialization and will serialize the initial data before submitting.

- **Doc.ingestSnapshot**
  - If the snapshot specifies a type and type.serialize is defined then the snapshot.data is deserialized.

- **Doc._otApply**
  - If the operation is a create operation and type.serialize is defined then op.create.data is deserialized.
 
- **Doc._tryCompose**
  - If an operation is to be composed into a create operation and the type is serializable then deserialize last.create.data before applying the operation to the initial data of create.
  - Another option is to reject apply the operation to create.data if the type supports serialization.

Two changes were made to the lib/ot.js file to support just in time serialization when handling op.op and op.create operations on the server.  I’m not sure if this is the best approach for serialization support on the server.  

Please let me know if I’ve missed anything crucial.  I’ve taken care to add a number of new tests and to follow the style in the surrounding code.  It is very important to Retrium that custom serialization is supported in sharedb and we are prepared to help out in any way.

Thanks for such a great library, looking forward to 1.0!


lib/client/doc.js
```diff
@@ -186,9 +186,9 @@ Doc.prototype.ingestSnapshot = function(snapshot, callback) {
   if (this.version > snapshot.v) return callback && callback();
 
   this.version = snapshot.v;
-  this.data = snapshot.data;
   var type = (snapshot.type === undefined) ? types.defaultType : snapshot.type;
   this._setType(type);
+  this.data = (this.type && this.type.serialize) ? this.type.deserialize(snapshot.data) : snapshot.data
```
DetweilerRyan: i'm concerned that setting this.data after the call to this._setType() is problematic since _setType() sets this.data to null if the type is null, but we need access to the type in order to deserialize the snapshot.data.  Thoughts?

nateps:
Hey Ryan!

I've spent some time thinking about this, and it is not something where there is an obviously elegant and performant solution, but I think it can be done.

## Approach 1

Fully serialize and deserialize every time data goes in or out of the OT type, so it is always deserialized outside of the type. This is the same whether ShareDB calls the methods or the type encapsulates it, so the type might as well implement this internally instead of requiring additional support in ShareDB. For example, the type could just call its own deserialize / serialize methods at the start and end of type.apply() and serialize at the end of type.create(). This is simple, easy to make sure it is correct, and easy to understand.

The downside is a loss of efficiency with the apply method on the client: you might be doing a lot of deep copies of the same data repeatedly on the client to implement your serialization and deserialization. On the server, we typically load a new copy of the snapshot from the database for every time we apply an op anyway, so any type that requires deep deserialization is going to be equally inefficient.

## Approach 2

For the ability to directly read from doc.data (writes should still go through doc.submitOp only), it might be nice to have doc.data be a custom object type (ex: MyObject or Immutable.Map). However, on the server, we don't really keep a doc around in memory, and we grab the snapshot from the DB every time an op is submitted. So when ingesting a snapshot into the client, we could deserialize the snapshot before setting it as doc.data and the type could return deserialized data from `type.apply()` and `type.create()`. The client should not call serialize, because it never sends snapshots to the server, only ops. On the server, we could always call deserialize and serialize before and after `type.apply()` and `type.create()`.

At the end of the day, you should probably be more concerned about write performance on the server than the client, because it will likely be the bottleneck. The server has to apply the ops for all clients connected to it, and clients are generally dealing with a small subset. I am concerned this approach has a foot-gun lurking, in that it seems like it will encourage implementation of deep serializing and deserializing the entire document on every single op on the server. For small documents this might be fine, but for large documents this could be too much overhead.

## Approach 3

If we were to optimize for the server first, we likely want to keep the snapshot in the serialized format as much as possible. If some intermediate datastructure needs to be created for us to apply an op, we probably only want to generate it for the part of the snapshot being modified and not the entire snapshot. We are already naturally getting a deep copy of the doc for every single op by fetching the document from the DB. Thus, the more I think about it, it seems like full deserialization and serialization on the server should ideally be avoided. Once again, this would mean that ShareDB would not call deserialize / serialize on the server and would pass in serialized snapshot data to `type.apply()`.

The wins in option #2 are due to keeping data deserialized in the client, where documents are maintained in memory and are not refetched from the database on each op. A solution that would be more complex but more optimized for performance in these different situations could do deserialization only when ingesting data into a client doc. Then, we could do type checking of the input to the apply method and do the apply and return deserialized data if the input was deserialized data (which would be the case on the client but not the server). This increases the complexity of the apply method in order to make server writes efficient even on larger docs and client reads straight from the doc.data efficient.

## Proposal

Ultimately, I think the additional complexity of #3 will be needed for certain use cases and is where a good general implementation would end up, but it would be nice to have the option to start with something simpler like #2 if you don't make frequent updates to large documents. For this to work, the server should not call serialize and deserialize, and this should be handled internally in the type if the input data is deserialized. The only method that ShareDB would call would be to deserialize on ingestion in the client. Client-side create would also need to deserialize, though it would probably be most efficient to add a function to create deserialized data directly. A simple type implementation would start by deserializing the entire input and then serializing the output to `type.apply()` unless the input is already deserialized. This allows the type to gradually optimize based on what the op is modifying.

The type API changes would be:

1. Apply is overloaded to take either serialized data or deserialized data: `apply(serializedData, op) -> serializedData'` and `apply(deserializedData, op) -> deserializedData'`. Internally, the version that deals with serializedData can call deserialize on the input and serialize on the output to start with.
2. Add: `deserialize(serializedData) -> deserializedData`
3. Add: `createDeserialized([initialData]) -> deserializedData`

And the changes to ShareDB would be:

1. Doc::ingestSnapshot should call `type.deserialize` before setting doc.data if defined
2. Doc::_otApply should call `type.createDeserialized` instead of create if defined

DetweilerRyan:
Hi Nate,

Thanks for the thoughtful response.  Your proposal for the custom serialization is well thought out and I'd be happy to implement it.  I'll create a fresh pull request.  Hopefully, we can get this into the 1.0.0 release!

Again, thanks for such a great project.  

Ryan

DetweilerRyan:
closed in favor of #127

PR #115, opened 2016-08-25
Update rich-text example README

### LINKED ISSUE #114: rich-text example README
The README of the new rich text example found at https://github.com/share/sharedb/tree/master/examples/rich-text looks to be copied from the textarea example and not updated to be about the new example.

curran:
Closes #114


nateps:
Good catch. Thanks for the fix!

curran:
My pleasure, thanks for the great work!

PR #119, opened 2016-08-31
Leaderboard - Eliminate connection global var

### LINKED ISSUE #118: Global var Antipattern in Leaderboard
In the Leaderboard example `index.jsx`, the variable `connection` is defined globally:

```js
connection = new sharedb.Connection(new WebSocket('ws://' + window.location.host));
```

Then later it is referenced from within `Leaderboard.jsx`:

```js
...
  componentDidMount: function() {
    var comp = this;
    var query = connection.createSubscribeQuery('players', {$sort: {score: -1}});
...
```

I would suggest to encapsulate `connection` as a module that makes the connection once (effectively a singleton connection instance) and exposes it to any code that `require`s it, so we can avoid having this global variable.

curran:
Closes #118

Verified that everything builds correctly and still works.


curran:
Any thoughts on this refactoring?

Issue #109, opened 2016-08-06
Textarea Example : Multiples docs


Derpolino:
Hi ! Sorry for my bad english

I want to do this with the textarea example : 
http://localhost:8080?id=1
http://localhost:8080?id=2
http://localhost:8080?id=hello

I want every id has its own document, how to do this ?

Thanks you !

sriramveeraghanta:
Even i got the same problem, can anyone help me.

serip88:
+ Me too, how to setup Multiples docs ?
+ And I don't know how to save data from sharedb to my database @.@

Plz help!

PR #123, opened 2016-09-26
Routing example

curran:
This new example uses React to encapsulate [sharedb-string-binding](https://github.com/share/sharedb-string-binding) as a component. It's bootstrapped from [create-react-app](https://github.com/facebookincubator/create-react-app). When complete, this will provide one example that addresses #109 

Remaining things to do:
- [x] Use react-router to implement navigation between documents
- [x] Styles
- [x] Figure out how to leverage the create-react-app devtools AND connect to ShareDB

I'm submitting this incomplete PR now to solicit feedback on the approach. One thing I'm not totally certain about is memory/connection leakages - I'm assuming that invoking `doc.destroy()` is enough, but please let me know if I'm missing something.


curran:
@avital @nateps This PR is ready for review. Any feedback would be welcome. Thank you.

curran:
image

curran:
Closing, as I doubt this will ever be merged.

Would appreciate any feedback as to why there's no interest here. I'd guess because this example brings in too many dependencies, and the "right way" to integrate ShareDB and React is not totally clear.

PR #129, opened 2016-10-08
[Proposal] Op Event Changes 

roark:
Standardizes events so that they always work the same way, no exceptions. Changed `op` event to `after op`. Added `before component` and `after component` events.
## 

Currently there are 2 events surrounding the application of an operation. They are:
- `before op` - Emitted before op is applied
- `op` - Emitted after op is applied

These behave in a standard way with one exception, when applying remote ops of json0 type. In this case the op is manually shattered, applied incrementally and the events are emitted before and after each "component op" has been applied.

_the comments explain:_

> Ideally, we would not need this extra complexity. However, it is helpful for implementing bindings…

Indeed it can be very helpful to have events triggered on the application of each component. It may also be helpful on a different OT types that support it. Or even on local ops as well as remote ops, depending on the application. 

That said, it also remains useful to have events that treat the op more atomically… before any portion of an op has been applied and after an op has been fully applied.
## 

Unfortunately in its current implementation it’s either one of the other (atomic or incremental) and it’s not up to the user to decide. It’s only based on a hardcoded data type’s remote ops.

This creates several interesting issues, an important one in particular is that there is no way to run code only after an entire json0 remote op has been applied. Which is relevant, if your app operations are meant to be treated as ‘atomic’. Such that they change the doc in a way that only makes sense after the op has been applied in its entirety. 

_As an example:_ One might want to render after an entire op has been applied. However, when applying json0 remote ops they are split up and the `op` event would be run multiple times. The doc.data may only make sense to render once the entire op has been applied, but you aren’t sure which `op` event is the last one… Another cases could be needing to transform a complex cursor by an full operation.

**Generally the issues are:**
1. No way to process json0 remote ops atomically.
2. No way to process other OT types incrementally.
3. No way to process local ops incrementally.
4. No way to decide if ops should be applied atomically or incrementally.
5. Events change meaning depending on type and source.

What would be nice is if we could listen to both op level and component level events. Have operations applied atomically or shattered if the type supports it. For both remote and local ops, if desired.
## 

I'd like to **propose** that rather than changing the meaning of the `before op` and `op` events based on OT type and source. The two events should more appropriately be four separate events:
- `before op` - emit before any portion of the op is applied.
- `before component` - emit before each component op is applied.
- `after component` - emit after each component op has been applied.
- `after op` - emit after the entire op has been applied.

And whether ops are applied atomically or incrementally should be based upon… 

``` javascript
  // if OT type implements .shatter() 
  doc.type.shatter !== undefined;

  // and upon two properties of the doc object
  doc.applyLocalOpsIncremental = false;
  doc.applyRemoteOpsIncremental = true;
```

**Benefits of this design:**
- Allows any OT type to define .shatter() and emit granular events.
- Enables `after op` to always emit entire ops, including remote json0 ops.
- Enables `before op` to always emit entire ops, including remote json0 ops.
- Enables ability to decide whether ops should be applied atomically or incrementally.
- Events always work the same way, no confusion.

**Design decisions**
- `before component` and `after component` will always fire. A single component op or an OT type that doesn’t support `shatter`, will emit the entire op. I did this to make transitioning `op` events to `after component` always work. However, I think the case could be made for not emitting these events if apply incrementally is disabled or shatter undefined.
- `applyLocalOpsIncremental` defaults to `false` and `applyRemoteOpsIncremental` defaults to `true`. This was only chosen to remain consistent with ShareDBs current implementation. I think ultimately it would make more sense for both to either be `true`… or possibly `false`, but most importantly consistent.
## 

I created this pull request to show the implementation of the ideas above. Please check it out and let me know your thoughts. 

I realized that by recommending an update to the event system, it means developers implementing ShareDB would need to update their code. So I’d like to hear fellow developer’s thoughts and input. These changes would require developers to implement the following updates:

``` javascript
on(‘begin op’)    // to… either on(‘begin op’) or on(‘begin component’)
on(‘op’)          // to… either on(‘after op’) or on(‘after component`)
```

In most cases, `begin op` would stay the same and `op` would change to `after op`. Only when granular events are specifically needed would a developer change events to use `before component` or `after component`.
## 

Thanks for everyone’s work on this awesome library, especially @nateps for continually improving it. It really is a hidden gem in the code universe.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -87,6 +87,10 @@ function Doc(connection, collection, id) {
   // The OT type of this document. An uncreated document has type `null`
   this.type = null;
 
+  // Enable ops be incrementally applied. OT Type must support type.shatter()
+  this.applyLocalOpsIncremental = false;
+  this.applyRemoteOpsIncremental = true;
+
   // The applyStack enables us to track any ops submitted while we are
   // applying an op incrementally. This value is an array when we are
   // performing an incremental apply and null otherwise. When it is an array,
@@ -507,12 +511,11 @@ Doc.prototype._otApply = function(op, source) {
       return this.emit('error', err);
     }
 
-    // Iteratively apply multi-component remote operations and rollback ops
-    // (source === false) for the default JSON0 OT type. It could use
-    // type.shatter(), but since this code is so specific to use cases for the
-    // JSON0 type and ShareDB explicitly bundles the default type, we might as
-    // well write it this way and save needing to iterate through the op
-    // components twice.
+    // The 'before op' event enables clients to pull any necessary data out of
+    // the snapshot before it gets changed
+    this.emit('before op', op.op, source);
+
+    // Iteratively apply multi-component for the OT types that support type.shatter().
     //
     // Ideally, we would not need this extra complexity. However, it is
     // helpful for implementing bindings that update DOM nodes and other
@@ -522,12 +525,15 @@ Doc.prototype._otApply = function(op, source) {
     // that the snapshot only include updates from the particular op component
     // at the time of emission. Eliminating this would require rethinking how
     // such external bindings are implemented.
-    if (!source && this.type === types.defaultType && op.op.length > 1) {
+    if ( (this.type.shatter && op.op.length > 1) && 
+         ( (this.applyLocalOpsIncremental && source) || 
+           (this.applyRemoteOpsIncremental && !source) ) ) {
+
       if (!this.applyStack) this.applyStack = [];
       var stackLength = this.applyStack.length;
-      for (var i = 0; i < op.op.length; i++) {
-        var component = op.op[i];
-        var componentOp = {op: [component]};
+      var shatteredOps = this.type.shatter(op.op);
+      for (var i = 0; i < shatteredOps.length; i++) {
+        var componentOp = {op: shatteredOps[i]};
         // Transform componentOp against any ops that have been submitted
         // sychronously inside of an op event handler since we began apply of
         // our operation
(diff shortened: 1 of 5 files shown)


lib/client/doc.js
```diff
@@ -522,12 +525,15 @@ Doc.prototype._otApply = function(op, source) {
     // that the snapshot only include updates from the particular op component
     // at the time of emission. Eliminating this would require rethinking how
     // such external bindings are implemented.
-    if (!source && this.type === types.defaultType && op.op.length > 1) {
+    if ( (this.type.shatter && op.op.length > 1) &&
```
roark: Considering removing `op.op.length > 1` check...

My reasons would be:

1. Can operations for custom OT types be other objects than arrays? If so this length check would cause an error.
2. Maybe an OT type operation has a length of one but due to its structure it can actually be shattered into smaller ops.

Doing so would call op.shatter on ops of length 1 or 0. Don't believe this should be an issue since the shatter method would be extremely fast in these cases and OT types should support these cases. Thoughts?

nateps:
Some good suggestions here

roark:
Thanks @nateps. Looking forward to your input, once you've been able to really mull everything over.

I had a couple of relevant ideas come to mind and wanted to put them into the ether.

1. In order to make any transition smoother, we could continue to emit the `op` event as an alias to `after op`, along with a depreciation message. This would cover the majority of cases, except for those special cases which require updates of `op` to `after component` or `before op` to `before component`.

2. Another thought... Instead of having the `applyLocalOpsIncremental` and `applyLocalOpsIncremental` flags, we could apply the op incrementally iff there is a listener on the doc for `before component` or `after component`.

 This would mean that if a user was listening to a `component` event, then the doc would apply both local and remote ops incrementally. Otherwise it would apply them atomically.

 This idea could keep things simpler, however it would mean the apply behavior changes based on subscription to an event. And it would also do away with the ability to apply remote and local ops differently.

roark:
Hi @nateps,

I know you keep quite busy, when you get a chance can you review this proposal and related pull request. 

I'd like to know your thoughts on what changes may be required to merge these ideas.

**Proposal TLDR Recap**
- Standardized application of ops for different ot-types (shatter / atomic) 
- Add additional of `before component` and `after component` events 
- Create `after op` -> `op` alias   (consider `op` depreciation in favor of `after op`) 

I think these additions, especially since some involve event names being integrated by end users, would be good to have in the official `1.0.0` release.

Thanks, keep up the great work.

dlio:
Any updates on this?  Looks good to me.

nateps:
Hey @roark!

Thanks for the well-thought out explanation and proposal! You've clarified many ways in which these event interfaces could be improved.

I do like the general direction of the proposal, and I'd like to implement this in a non-breaking way (all new events and options) in the current version of ShareDB along with deprecation of the prior inconsistent event names. In a future major version, we'd remove the deprecated events.

@ericyhwang and I have recently begun dedicating more time to ShareDB, and we're first prioritizing getting existing PRs merged in.

What are your thoughts around implementing a deprecation strategy for these changes?

Another question, should we have a way of specifying defaults? Perhaps you could configure a default per type and/or global default on the Connection?

gkubisa:
I really like this proposal. :+1: 

As we're going to change the events, perhaps it's a good opportunity to make some more small improvements too:

1. Pass in an Event object to the event handler as the only parameter, instead of multiple positional parameters (eg `op` and `source`). We could then pass in more information to the handlers in the future easily without affecting backward compatibility.
2. Make the events always async. Related to https://github.com/share/sharedb/issues/206.

curran:
It looks like the changes are generally good, but possibly breaking, so should be behind a deprecation strategy / flag. If that were added to this PR, retaining exactly the old behavior if the flag is not enabled, perhaps it could be merged.

Is anyone really interested in these changes? The win does not seem that big to be honest. If not, I suggest to close without merge due to lack of interest.

curran:

alecgibson:
Closing this, as we've merged https://github.com/share/sharedb/pull/401

The above Pull Request adds support for `before op batch` and `op batch` events as a non-breaking change (where this PR renamed the existing `before op` and `op` events.

Features that the above PR **does not** add, which have been proposed here (and could be proposed in future PRs):

 - Flags for toggling incremental op updates (we discussed this, and saw no harm in leaving them on all the time, and you just don't subscribe to the events if you don't care)
 - Support for any shatterable type, as opposed to just `json0`

PR #134, opened 2016-11-27
Add missing done variable, fix test

### LINKED ISSUE #133: Undefined "done" variable in test
[test/client/submit.js#L1062](https://github.com/share/sharedb/tree/bb11214136584b8e67e8373efedd5216a572def8/test/client/submit.js#L1062) references an undefined `done` variable. This seems to be the root cause of the first error in the failing `sharedb-mongo` tests ( tracked as https://github.com/share/sharedb-mongo/issues/48 ).

curran:
Closes #133. Inching towards https://github.com/share/sharedb-mongo/issues/48

Verified that this removes the first encountered error when running sharedb-mongo tests.

### DIFF
--- a/test/client/submit.js
+++ b/test/client/submit.js
@@ -1060,7 +1060,7 @@ describe('client submit', function() {
       var doc = this.backend.connect().get('dogs', 'fido');
       doc.create([3], deserializedType.type.uri, function(err) {
         if (err) return done(err);
-        db.getSnapshot('dogs', fido, null, null, function(err, snapshot) {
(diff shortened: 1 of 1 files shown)


curran:
Thanks @rkstedman for reviewing!

rkstedman:
Nate wants to verify a couple of related things before merging this--even though these small changes in themselves look good. So that's why I'm holding off merging. I'll try and get an update back to you soon as I can! Thanks again for taking the time to track down and fix these tests :)

Issue #121, opened 2016-09-14
`Connection` does not reconnect automatically


jahewson:
When I restart my web socket server, the `Connection` on the client emits a `state` event with the `newState` set to `"disconnected"`, which according to the documentation means:

> 'disconnected' Connection is closed, but it will reconnect automatically

However, the connection does not reconnect. Given that the socket which was passed to the  `Connection`'s constructor is now closed, I don't understand how reconnection would be possible. Don't we need a new WebSocket? What's going on here?

DetweilerRyan:
The inline documentation you are referring to is incorrect.  I recommend using one of the following modules for re-connectable websockets:

[websocket.js](https://www.npmjs.com/package/websocket.js)
[reconnectingwebsocket](https://www.npmjs.com/package/reconnectingwebsocket)

jahewson:
Ok, that's good to hear. I had thought that maybe it was just me. Might want to update the README which says:

- Reconnection of document and query subscriptions
- Offline change syncing upon reconnection

atsepkov:
I too encountered reconnection problems, everything works fine locally, but Heroku seems to kill the socket if no event occurs within a minute. I have now added a ping-pong to help, and replaced my WebSocket with a ReconnectingWebSocket, but running into a weird issue. The socket does indeed reconnect, but the events no longer sync between clients after the reconnection.

Oddly, it seems like they still make it to the server, just not the other client - almost as if the two clients started thinking they're in different rooms until one of the pages is reloaded, or as if the subscription to remote changes broke after the disconnection. Do I need to rerun `subscribe` events after the reconnection? Do I need to refetch the docs? Do I need to recreate the ShareDB handle on every reconnection? On the main page in the list of features it says that ShareDB handles reconnection, what exactly does that mean and what should we handle ourselves?

Thanks

curran:
This is addressed in PR #138 .

I've had quite good luck using https://github.com/pladaria/reconnecting-websocket , dropping it in in place of raw WebSocket.

FWIW, initially I tried https://github.com/joewalnes/reconnecting-websocket , but decided not to use it because it depends on `window` (breaks in a Node environment), and is no longer maintained.

pedrosanta:
Yep, I've also come across this issue of `Connection` not reconnecting automatically.

[...]

PR #138, opened 2017-01-26
Reconnecting Websocket Example and Updated Readme

mbalex99:
I updated the **textarea** example and updated the readme about how to handle reconnections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,9 +27,25 @@ tracker](https://github.com/share/sharedb/issues).
 - Projections to select desired fields from documents and operations
 - Middleware for implementing access control and custom extensions
 - Ideal for use in browsers or on the server
-- Reconnection of document and query subscriptions
 - Offline change syncing upon reconnection
 - In-memory implementations of database and pub/sub for unit testing
+
+### Reconnection
+
+**TLDR**
+```javascript
+const WebSocket = require('reconnecting-websocket');
+var socket = new WebSocket('ws://' + window.location.host);
+var connection = new sharedb.Connection(socket);
+```
+
+The native Websocket object that you feed to ShareDB's `Connection` constructor **does not** handle reconnections. 
+
+The easiest way is to give it a WebSocket object that does reconnect. There are plenty of example on the web. The most important thing is that the custom reconnecting websocket, must have the same API as the native rfc6455 version.
+
+In the "textarea" example we show this off using a Reconnecting Websocket implementation from [https://github.com/pladaria/reconnecting-websocket](reconnecting-websocket).
+
+
 
 ## Example apps
 
--- a/examples/textarea/client.js
+++ b/examples/textarea/client.js
@@ -2,14 +2,35 @@ var sharedb = require('sharedb/lib/client');
 var StringBinding = require('sharedb-string-binding');
 
 // Open WebSocket connection to ShareDB server
+const WebSocket = require('reconnecting-websocket');
 var socket = new WebSocket('ws://' + window.location.host);
 var connection = new sharedb.Connection(socket);
+
+var element = document.querySelector('textarea');
+var statusSpan = document.getElementById('status-span');
+status.innerHTML = "Not Connected"
+
+element.style.backgroundColor = "gray";
+socket.onopen = function(){
+  status.innerHTML = "Connected"
+  element.style.backgroundColor = "white";
+};
+
+socket.onclose = function(){
+  status.innerHTML = "Closed"
+  element.style.backgroundColor = "gray";
+};
+
+socket.onerror = function() {
+  status.innerHTML = "Error"
+  element.style.backgroundColor = "red";
+}
 
 // Create local Doc instance mapped to 'examples' collection document with id 'textarea'
 var doc = connection.get('examples', 'textarea');
 doc.subscribe(function(err) {
   if (err) throw err;
-  var element = document.querySelector('textarea');
+  
   var binding = new StringBinding(element, doc);
   binding.setup();
 });
--- a/examples/textarea/package.json
+++ b/examples/textarea/package.json
@@ -15,6 +15,7 @@
   "license": "MIT",
   "dependencies": {
     "express": "^4.14.0",
+    "reconnecting-websocket": "^3.0.3",
     "sharedb": "^1.0.0-beta",
     "sharedb-string-binding": "^1.0.0",
     "websocket-json-stream": "^0.0.1",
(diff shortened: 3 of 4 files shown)


README.md
```diff
@@ -27,10 +27,26 @@ tracker](https://github.com/share/sharedb/issues).
 - Projections to select desired fields from documents and operations
 - Middleware for implementing access control and custom extensions
 - Ideal for use in browsers or on the server
-- Reconnection of document and query subscriptions
```
curran: Nice! It's high time this documentation be updated.

curran:
Anyone reviewing these PRs? This change looks good to me.

examples/textarea/client.js
```diff
@@ -2,14 +2,35 @@ var sharedb = require('sharedb/lib/client');
...
 var socket = new WebSocket('ws://' + window.location.host);
 var connection = new sharedb.Connection(socket);
 
+var element = document.querySelector('textarea');
+var statusSpan = document.getElementById('status-span');
+status.innerHTML = "Not Connected"
+
+element.style.backgroundColor = "gray";
+socket.onopen = function(){
+  status.innerHTML = "Connected"
```
curran: This is quite useful stuff!

curran:
@nateps @rkstedman This PR looks like a nice improvement. Would merge it if I could. Any chance one of you could review this one? Thank you.

README.md
```diff
@@ -27,10 +27,26 @@ tracker](https://github.com/share/sharedb/issues).
...
 - Offline change syncing upon reconnection
```
pedrosanta: @mbalex99 I mean, if reconnection doesn't come out of the box, I think this line should also be removed, as it's misleading.

README.md
pedrosanta: Actually, as per #121, this is correct.

When the WebSocket reconnects, the [Connection middleware](https://github.com/share/sharedb/blob/master/lib/client/connection.js) reconnects both [documents](https://github.com/share/sharedb/blob/master/lib/client/connection.js#L302) and [query](https://github.com/share/sharedb/blob/master/lib/client/connection.js#L296) subscriptions.

We might need more clear documentation, but this line is correct.

README.md
pedrosanta: Actually I retract my comment, this actually happens too.

[Ops are added to pending 'queue'/array](https://github.com/share/sharedb/blob/master/lib/client/doc.js#L678) and [flushed/synced on socket reconnection](https://github.com/share/sharedb/blob/master/lib/client/doc.js#L329).

This line is also correct.

README.md
pedrosanta: Perhaps a more clear wording would be `- Reconnection of document and query subscriptions on socket reconnection (Check 'Reconnection' section below)`. Thoughts?

README.md
pedrosanta: Also, a better wording could be `- Offline change syncing upon socket reconnection`.

README.md
curran: @pedrosanta I think that would be more clear. I'd welcome that addition here.

@nateps This PR seems to address a common source of frustration/confusion on ShareDB reconnection. Any chance it could be reviewed by the team?

curran:
@nateps @rkstedman Any chance of reviewing this PR? I find it to be a great improvement.

pedrosanta:
@curran @nateps @rkstedman I think this PR still needs some updates though, as I've commented.

It was great if @mbalex99 could address those. 🙂

curran:
@pedrosanta Indeed, thanks for your suggestions. The discussion is quite long, and it will take some work to go through and see specifically which updates/improvements still need to be made.

Would you mind summarizing the specific changes you'd like to see?

(or, better yet, maybe you could branch off this and create a new PR with your proposed changes :wink: )

curran:
@pedrosanta To summarize your proposed changes from the discussion:

 * Instead of deleting `Reconnection of document and query subscriptions`, replace it with `Reconnection of document and query subscriptions on socket reconnection (Check 'Reconnection' section below)`
 * Change `Offline change syncing upon reconnection` to `Offline change syncing upon socket reconnection.`
 * in https://github.com/share/sharedb/blob/master/lib/client/connection.js#L90, change

```
  // - 'disconnected' Connection is closed, but it will reconnect automatically
  // - 'closed'       The connection was closed by the client, and will not reconnect
  // - 'stopped'      The connection was closed by the server, and will not reconnect
```

to

```
  // - 'disconnected' Connection is closed, and reconnect can be attempted
  // - 'closed'       The connection was closed by the client, a reconnect should not be attempted
  // - 'stopped'      The connection was closed by the server, a reconnect should not be attempted
```

Are these the complete changes to this PR you'd like to see? It almost feels like these changes could be made in a separate documentation-focused PR, after this PR gets merged. Thoughts?

curran:
@nateps @rkstedman [PR tending note] I think this PR can be merged. It would close the long-standing issue https://github.com/share/sharedb/issues/121 , which would be a big win. Changes are in documentation and examples only, so no risk of ShareDB breakage.

README.md
```diff
@@ -27,10 +27,26 @@ tracker](https://github.com/share/sharedb/issues).
...
+const WebSocket = require('reconnecting-websocket');
+var socket = new WebSocket('ws://' + window.location.host);
+var connection = new sharedb.Connection(socket);
+```
+
+The native Websocket object that you feed to ShareDB's `Connection` constructor **does not** handle reconnections. 
+
+The easiest way is to give it a WebSocket object that does reconnect. There are plenty of example on the web. The most important thing is that the custom reconnecting websocket, must have the same API as the native rfc6455 version.
+
+In the "textarea" example we show this off using a Reconnecting Websocket implementation from [https://github.com/pladaria/reconnecting-websocket](reconnecting-websocket).
```
nateps: This is a great README addition, and it clarifies one way that ShareDB Connections support reconnection.

I'd also like to highlight that it is possible and intended that reconnection is implemented by calling `Connection::bindToSocket()` with a new WebSocket instance, as @DetweilerRyan pointed out. We've chosen not to implement reconnection logic in ShareDB, itself. There are other great libraries for that purpose, and application developers may need to leverage different libraries or custom logic depending on their requirements.

Reading the code of Reconnecting WebSocket, a significant amount of its complexity is necessitated by the constraint of exposing a single instance externally and  internally creating new instances of WebSocket objects. If the reconnection logic of Reconnecting WebSocket implementation works for you, this is totally fine!

However, if you want to implement custom reconnection logic for your production environment (say you want to do backoff in a different way or fallback to other protocols or libraries), calling the ShareDB `bindToSocket()` with a new WebSocket or WebSocket-like instance might be simpler.

PR #140, opened 2017-02-24
catch not opened error

krichprollsch:
I'm doing some tests with https://artillery.io/, and I encounter a crash of the server with the error:

```
 Websocket : connected and listening to JSON Stream
/home/pierre/wisembly/notes-poc/node_modules/ws/lib/WebSocket.js:356
      else throw new Error('not opened');
           ^
Error: not opened
    at WebSocket.send (/home/pierre/wisembly/notes-poc/node_modules/ws/lib/WebSocket.js:356:18)
    at WebSocketJSONStream._write (/home/pierre/wisembly/notes-poc/node_modules/websocket-json-stream/index.js:27:11)
    at doWrite (_stream_writable.js:333:12)
    at writeOrBuffer (_stream_writable.js:319:5)
    at WebSocketJSONStream.Writable.write (_stream_writable.js:246:11)
    at Agent.send (/home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/agent.js:164:15)
    at Agent._reply (/home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/agent.js:219:8)
    at callback (/home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/agent.js:240:15)
    at /home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/agent.js:490:7
    at /home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/backend.js:340:9
    at /home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/backend.js:289:7
    at Backend.trigger (/home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/backend.js:137:20)
    at Backend._sanitizeSnapshot (/home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/backend.js:205:8)
    at /home/pierre/wisembly/notes-poc/node_modules/sharedb/lib/backend.js:286:13
    at Query.callback (/home/pierre/wisembly/notes-poc/node_modules/sharedb-postgres/index.js:150:11)
    at Query.handleReadyForQuery (/home/pierre/wisembly/notes-poc/node_modules/pg/lib/query.js:89:10)
npm ERR! Linux 4.2.0-1-amd64
npm ERR! argv "/usr/local/lib/node-v6.9.1-linux-x64/bin/node" "/usr/local/bin/npm" "start"
npm ERR! node v6.9.1
npm ERR! npm  v3.10.8
```

I don't know why the error happen, But I suppose the crash of the server is not the good behavior in this case. Catching and logging it should be better.

But not sure if it's the good way to go. WDYT ?

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -161,7 +161,12 @@ Agent.prototype.send = function(message) {
   if (this.closed) return;
 
   this.backend.emit('send', this, message);
-  this.stream.write(message);
+  try {
+      this.stream.write(message);
+  } catch(err) {
+      // Quietly drop replies if the stream was closed
+      console.error('Send message stream error', message, err);
+  }
 };
 
 Agent.prototype._sendOp = function(collection, id, op) {


krichprollsch:
The error happen when I send an operation with a bad version.

sferoze:
I have noticed this same error.

curran:
Would either of you be able to post a small example that can reproduce the error?

zag2art:
It is a common error. The reason is that ws-connection is already closed (because a user reloaded the page) but server tries to send a message anyway. Here is a right way of handling this https://github.com/websockets/ws/issues/322

I think the problem is here - https://github.com/avital/websocket-json-stream/blob/master/index.js#L27

There should be something like this
```js
this.ws.send(JSON.stringify(msg), function fn(err) {
  if (err) handleError(err); // or just ignore
});
```

sferoze:
For me in agent.js line 164 `this.stream.write(message)` was causing throwing the uncaught error.

To fix I wrapped that line in a try/catch statement. This should be updated in sharedb master branch.

DetweilerRyan:
Errors on a stream instance can be caught by attaching a handler for the error event on the stream instance.  A call to stream.write( msg, callback ) that results in an error will only throw if you do not provide a callback to the write method or there is no handler for the error event.

The call `this.stream.write(message)` may fail for a whole host of reasons other than the common `not opened` error and therefore it would not be appropriate the sharedb implementation to make assumptions around how those errors should be handled. 

Bellow is how we handle stream errors:
```
	stream.on('error', (err) => {
		switch(err && err.message && err.message.toLowerCase()) {
			case 'timed out':
			case 'write after end':
			case 'not opened':
			case 'reconnected':
			case 'disconnected':
				return;
			default:
				airbrake.notify(err);
				return;
		}
	});
```
As you can see there are a few other common errors that we have encountered in production other than the `not opened` error that we ignore, but all other errors are reported to airbrake. The other errors we've encountered are a result of our use of both browserchannel and websockets.

Hope this helps.

curran:
I'm seeing a similar error during development. Interestingly, `mongodb/lib/utils.js` is referenced as the origin of it.

```
/home/curran/repos/notes-poc/node_modules/mongodb/lib/utils.js:123
    process.nextTick(function() { throw err; });
                                  ^

Error: not opened
    at WebSocket.send (/home/curran/repos/notes-poc/node_modules/ws/lib/WebSocket.js:358:18)
    at WebSocketJSONStream._write (/home/curran/repos/notes-poc/node_modules/websocket-json-stream/index.js:27:11)
    at doWrite (_stream_writable.js:371:12)
    at writeOrBuffer (_stream_writable.js:357:5)
    at WebSocketJSONStream.Writable.write (_stream_writable.js:274:11)
    at Agent.send (/home/curran/repos/notes-poc/node_modules/sharedb/lib/agent.js:164:15)
    at Agent._reply (/home/curran/repos/notes-poc/node_modules/sharedb/lib/agent.js:219:8)
    at callback (/home/curran/repos/notes-poc/node_modules/sharedb/lib/agent.js:240:15)
    at /home/curran/repos/notes-poc/node_modules/sharedb/lib/agent.js:490:7
    at /home/curran/repos/notes-poc/node_modules/sharedb/lib/backend.js:340:9
```

gkubisa:
As @DetweilerRyan said [above](https://github.com/share/sharedb/pull/140#issuecomment-315221774), the solution is to attach an `error` handler to the stream. However, the problem is that [websocket-json-stream] just throws errors on write and does not emit `error` events. Since [websocket-json-stream] is not maintained, I implemented my own version, [@teamwork/websocket-json-stream], which emits the `error` events correctly and fixes some other issues, too. [@teamwork/websocket-json-stream] is a drop-in replacement for [websocket-json-stream].

Since the problem is not with ShareDB but with a library it recommends, I'd suggest updating the docs and examples instead of making the code change.

[websocket-json-stream]: https://www.npmjs.com/package/websocket-json-stream
[@teamwork/websocket-json-stream]: https://www.npmjs.com/package/@teamwork/websocket-json-stream

curran:

PR #174, opened 2017-10-17
Emit error on failed resubscribe

gkubisa:
Normally when a subscribe callback is missing, an error event is generated on error, except for the following scenario addressed by this PR:

1. state === 'connecting'
2. subscribe() - can't subscribe in this state
3. state === 'connected'
4. _resubscribe()
5. Backend returns an error

Expected: doc.emit('error', error)
Actual: no error event

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+end_of_line = LF
+charset = utf-8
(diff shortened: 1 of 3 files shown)


gkubisa:

ericyhwang:
Nate's comments, from the PR review meeting:

* The state machine on reconnecting is interesting.
* I'll have to look at the rest of the code to check for consistency.

curran:

PR #179, opened 2017-11-02
fixed wrong property name in example

EragonJ:
it looks like `selectedPlayer` is a Doc class and the `name` property is within `data` property. With this fix, the UI can be shown correctly when you select any person from the leaderboard.


ericyhwang:
Looks good, thanks for noticing the issue and fixing!

And sorry for the delay - a couple new people (me included) have just been added to help with processing smaller PRs, which should improve response times.

curran:
Thanks @ericyhwang !

PR #201, opened 2018-04-03
Fix server crashing on failed publish

gkubisa:
This PR fixes the issue where the entire server crashes when ShareDB fails to publish a message to its PubSub system.

The easiest way to reproduce the problem is to use Redis for PubSub (with node-redis configured to try to restore the connection indefinitely), stop Redis and then send some operations to ShareDB server. ShareDB will then try to publish a message to Redis and crash the server.

The fix is to simply add an empty callback to the `publish` function. This works fine as ShareDB can gracefully recover from losing those messages by reading the missing document revisions from the database (eg MongoDB). For the same reason I don't think it is necessary to propagate the publishing error up, although it certainly is a possibility.

### DIFF
--- a/lib/pubsub/index.js
+++ b/lib/pubsub/index.js
@@ -16,6 +16,11 @@ function PubSub(options) {
   // isn't complete until the callback returns from Redis
   // Maps channel -> true
   this.subscribed = {};
+
+  var self = this;
+  this._defaultCallback = function(err) {
+    if (err) return self.emit('error', err);
+  };
 }
 module.exports = PubSub;
 emitter.mixin(PubSub);
@@ -27,22 +32,29 @@ PubSub.prototype.close = function(callback) {
       map[id].destroy();
     }
   }
-  if (callback) callback();
+  if (callback) process.nextTick(callback);
 };
 
 PubSub.prototype._subscribe = function(channel, callback) {
-  callback(new ShareDBError(5015, '_subscribe PubSub method unimplemented'));
+  process.nextTick(function() {
+    callback(new ShareDBError(5015, '_subscribe PubSub method unimplemented'));
+  });
 };
 
 PubSub.prototype._unsubscribe = function(channel, callback) {
-  callback(new ShareDBError(5016, '_unsubscribe PubSub method unimplemented'));
+  process.nextTick(function() {
+    callback(new ShareDBError(5016, '_unsubscribe PubSub method unimplemented'));
+  });
 };
 
 PubSub.prototype._publish = function(channels, data, callback) {
-  callback(new ShareDBError(5017, '_publish PubSub method unimplemented'));
+  process.nextTick(function() {
+    callback(new ShareDBError(5017, '_publish PubSub method unimplemented'));
+  });
 };
 
 PubSub.prototype.subscribe = function(channel, callback) {
+  if (!callback) callback = this._defaultCallback;
   if (this.prefix) {
     channel = this.prefix + ' ' + channel;
   }
@@ -64,6 +76,7 @@ PubSub.prototype.subscribe = function(channel, callback) {
 };
 
 PubSub.prototype.publish = function(channels, data, callback) {
+  if (!callback) callback = this._defaultCallback;
   if (this.prefix) {
     for (var i = 0; i < channels.length; i++) {
       channels[i] = this.prefix + ' ' + channels[i];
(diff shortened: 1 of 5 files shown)


curran:
@gkubisa The Travis build is failing. Did you run the CI check locally first? Perhaps some tests need to be updated as well.

gkubisa:
When I check the [build details in Travis](https://travis-ci.org/share/sharedb/builds/364851601?utm_source=github_status&utm_medium=notification), it says the build failed only in "Node.js 0.1" - I don't believe anyone cares about that ancient version anymore. The build succeeds in all the more recent node versions.

curran:
@gkubisa Aha great point. Perhaps remove that old version from the Travis config then? Not sure if @nateps would approve of that, but seems like a reasonable thing to do. Also it's odd that CI passes for the old version in previous commits, but breaks after this change. I wonder what could be the issue...

gkubisa:
I first fixed the `.travis.yml` file - it meant to run tests on node `0.10` not `0.1`. However, the tests fail on `0.10` too. It looks like the reason is an update to mingo, which now uses `Symbol`, which is not supported in earlier versions of nodejs. See https://travis-ci.org/share/sharedb/jobs/365189513#L545-L548.

So, I then updated .travis.yml to run tests on all [supported nodejs versions](https://github.com/nodejs/Release) only. The tests pass now.

curran:
@ericyhwang Excellent proposal! Looks good to me.

gkubisa:
Thanks for the review @ericyhwang. I mostly agree with this proposal and already made `PubSub` an `EventEmitter` and removed `util.doNothing`.

Regarding emitting the `error` event, I think a better place to do that would be in [sharedb-redis-pubsub](https://github.com/share/sharedb-redis-pubsub/blob/master/index.js). All the methods (`_subscribe`, `_unsubscribe`, `_publish`, `close`) would emit an error internally but only if no callback is provided. This behavior would be consistent with most node libs out there and it would make it impossible to forget to emit the error in the client code (ShareDB in this case).

To emit errors as above, this PR will need to be merged first and a new version of ShareDB published, so that `sharedb-redis-pubsub` could depend on the latest version.

curran:
Looking forward to see this merged :)

Tangentially related to https://github.com/share/sharedb-redis-pubsub/issues/4#issuecomment-384194599

gkubisa:
@ericyhwang Great point about emitting the errors in the base class - I've just implemented this approach. Since only `_subscribe`, `_publish` and `_unsubscribe` are supposed to be overridden, I think we should be good. I completely agree about emitting only unhandled errors.

I believe this PR is ready for merging now.

nateps:
In response to @curran about Travis config ("Perhaps remove that old version from the Travis config then? Not sure if @nateps would approve of that"):

1. My guidance on node versions to test in CLI is that we should include the current major version of Node.js and all versions under LTS. Happy to accept updates to this as time goes on. Ideally we'd do it as a separate PR in the future, but let's just get it merged this time, since it is already done. Thanks for the help! Glad we were already on the same page about what versions would be useful to test.

2. @curran would you mind creating a `contributing.md` file a la https://help.github.com/articles/setting-guidelines-for-repository-contributors/? We can start with the stuff we've discussed before, and let's make it easier on ourselves by starting with something as simple as possible; we can always add more. I'm happy for you to suggest some guidelines based on other open source project guidelines you respect.

lib/pubsub/index.js
```diff
@@ -13,8 +16,14 @@ function PubSub(options) {
   // isn't complete until the callback returns from Redis
   // Maps channel -> true
   this.subscribed = {};
+
+  var self = this;
```
nateps: As a style convention, the use of `self` as a variable totally works and that is what I use when there is not a better name to use. (You'll see this in sharedb tests even)

However, in sharedb and other repos, we do prefer to use the instance variable name (`pubSub` would be the instance name of the class `PubSub`) rather than `self`. I believe this makes the code and stack traces more specific, easier to understand, and easier to grep / search.

lib/pubsub/index.js
```diff
@@ -13,8 +16,14 @@ function PubSub(options) {
...
+  this._defaultCallback = function(err) {
```
nateps: I'm impressed you repeated this pattern from QueryEmitter! Great job. 🎉 

For future people reading this PR, the pattern means that we only have to create this callback function and the closure of the `pubSub` reference one time. The alternative would have been to create an anonymous function every time we publish. This is an ideal application of the same optimization, and it is a good use of consistency in naming and code patterns.

lib/pubsub/index.js
```diff
@@ -109,9 +126,7 @@ PubSub.prototype._removeStream = function(channel, stream) {
   // complete before we can count on being subscribed again
   delete this.subscribed[channel];
 
-  this._unsubscribe(channel, function(err) {
-    if (err) throw err;
-  });
+  this._unsubscribe(channel, this._defaultCallback);
```
nateps: Thanks for spotting this inconsistency and cleaning it up!

Issue #161, opened 2017-06-21
Looks like server ops to a doc are queued up and not applied after a client doc is destroyed


YilinGao:
Hello, I'm working on a code collaborative editing project (like Google Doc) and I'm using ShareDB to manage updates by multiple users. Because there are more than one code file for users to update, users need to leave one doc at some time and go back to it later. But I have encountered a problem (or noticed a phenomenon) during this process:

Suppose there are 2 users, A and B, and they are both editing doc X. Then B leaves doc X, while A is still editing X (submitting a series of ops to ShareDB server). Now B comes back to doc X, and here I find the data that B's doc X has received are: **the initial `doc.data` which is what user B left before, and then the series of ops that A submitted after B left**.

With exact code:
```javascript
handleUserEnterLeaveModule(leaveModule, enterModule) {
    if (leaveModule) {   // user B leaves doc X
      if (this.doc) {
        this.doc.destroy();
        this.doc.removeListener('op', this.handleSharedbUpdateRunMethod);
        this.doc = null;
      }
    }
    if (enterModule) { // user B enters doc X
      this.initSharedbDoc(enterModule);
    }
  }

initSharedbDoc(enterModule) {
    if (this.doc && this.doc.id === enterModule) return;
    this.doc = this.connection.get('module-editor-run-method', enterModule);
    this.doc.subscribe((err) => {
      if (err) throw err;
      if (!this.doc.data) {
        this.doc.create(this.codeMirror.getDoc().getValue());
      } else {
        // when user B enters doc X again, this.doc.data is not the latest version in server,
        // but where user B left before, without updates from user A
        this.codeMirror.getDoc().setValue(this.doc.data);
      }
      this.doc.on('op', this.handleSharedbUpdateRunMethod);
    });
  }

// after doc initialization, the op handler is called for every op from user A
handleSharedbUpdateRunMethod = (op, source) => {
    if (!source) {
      // update CodeMirror value according to op
    }
  };
```

**Although in the current way my program has correct outcome, I feel it's kind of not reasonable that data from server are not up-to-date. My question is, is this the correct way that ShareDB is designed to work? If not, can anybody give any idea about how I should get it work in an expected way?**

Thanks!

gkubisa:
I think the problem is that the document is still cached in the connection after calling destroy.

[...]

PR #204, opened 2018-04-18
Fix Doc.prototype.destroy

gkubisa:
It fixes the issue where a document is re-added to a collection after calling `destroy`, causing a memory leak.

It should also fix https://github.com/share/sharedb/issues/161.

Re not waiting for the unsubscribe callback in `destroy`, I think it is not necessary because:

- unsubscribe cannot fail on the server side (as far as I see),
- unsubscribe is performed automatically on disconnect,
- the doc is marked as unsubscribed immediately on the client side.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -104,7 +104,9 @@ emitter.mixin(Doc);
 Doc.prototype.destroy = function(callback) {
   var doc = this;
   doc.whenNothingPending(function() {
-    if (doc.wantSubscribe) doc.unsubscribe();
+    if (doc.wantSubscribe) {
+      doc.unsubscribe();
+    }
     doc.connection._destroyDoc(doc);
     if (callback) callback();
   });
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -105,10 +105,15 @@ Doc.prototype.destroy = function(callback) {
   var doc = this;
   doc.whenNothingPending(function() {
     if (doc.wantSubscribe) {
-      doc.unsubscribe();
+      doc.unsubscribe(function(err) {
+        if (!err) doc.connection._destroyDoc(doc);
+        if (callback) return callback(err);
+        if (err) this.emit('error', err); 
+      });
+    } else {
+      doc.connection._destroyDoc(doc);
+      if (callback) callback();
     }
-    doc.connection._destroyDoc(doc);
-    if (callback) callback();
   });
 };
 
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -104,10 +104,8 @@ emitter.mixin(Doc);
 Doc.prototype.destroy = function(callback) {
   var doc = this;
   doc.whenNothingPending(function() {
+    if (doc.wantSubscribe) doc.unsubscribe();
     doc.connection._destroyDoc(doc);
-    if (doc.wantSubscribe) {
-      return doc.unsubscribe(callback);
-    }
     if (callback) callback();
   });
 };
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -106,9 +106,13 @@ Doc.prototype.destroy = function(callback) {
   doc.whenNothingPending(function() {
     if (doc.wantSubscribe) {
       doc.unsubscribe(function(err) {
-        if (!err) doc.connection._destroyDoc(doc);
-        if (callback) return callback(err);
-        if (err) this.emit('error', err); 
+        if (err) {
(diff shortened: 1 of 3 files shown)


gkubisa:
Thanks for the review @curran.

I added the braces as you suggested - I also prefer this style but tried to be consistent with the rest of the code base, which very often omits braces.

I also added an assertion to an existing test which failed before my changes and now it passes.

gkubisa:

ericyhwang:
We've got our ~monthly Share PR review meeting tomorrow, so we'll take a look at this and the other couple PRs during the meeting!

(I'll also ask about perhaps switching to shorter, more frequent review sessions, since that'll be easier all around if Nate's schedule can now accommodate it.)

nateps:
Great catch!

I agree that it isn't 100% clear whether we need to wait for the unsubscribe to happen, and we might get away with calling the callback before the unsubscribe is fully effective. As you mentioned, the client calls back when it is disconnected.

Here is the reasoning behind that: Calling back on disconnect is needed to ensure that the callback to unsubscribe is always called. (The unsubscribe callback may be called after acknowledgement from the sever if we are connected, in a nextTick if we are disconnected, or at the time of disconnection.) If the client is disconnected, the server won't be able to send response messages and it will clean up the server-side agent responsible for the client. The client will get a new agent if it reconnects. So from the perspective of the client, the unsubscribe is effective immediately after a disconnection.

It is different if the client is connected, since if we call `destroy()` on a subscribed document, we could continue to get ops from the server until it acknowledges the unsubscribe. Now, if the document has been destroyed, the connection will ignore these ops. But if the document is then created or fetched again, it could get complicated. I'm not 100% sure I know where this would go wrong, but it seems like we should just be conservative and delay the callback until after the unsubscribe is complete to avoid any possibility of complexity arising from not waiting long enough to call back. In every case where I've assumed something should be safe to do and couldn't prove it, that later bit me. ;-)

Roughly, I could imagine something like the following being an issue:
1. Connection A is subscribed to a doc
2. Connection B performs an op on a doc. The server commits it
3. Connection A deletes the doc, then calls destroy immediately
4. As soon as the delete op is committed, the Connection A sends an unsubscribe command
5. In the callback to destroy, Connection A creates the doc
6. Connection B's op is sent to Connection A *before* the unsubscribe or create messages are received by the server
7. Connection A receives the op from Connection B thinking that it shouldn't be getting any more ops, because it called destroy
8. Connection A might not be able to OT the op from Connection B against its local data, because the document was created anew. If we had waited until the unsubscribe were complete, then we wouldn't have this race condition.

The above is really complicated, but I think it might be an issue, and there is an easy way to avoid testing fate in this case. Knowing that unsubscribe will always call back and destroy waits until pending operations are complete, I think it is best if we just wait until unsubscribe calls back in all cases before calling the destroy callback. I'm a lot more confident we won't run into any race conditions if we do it that way.

Thus, I recommend the following:
```
Doc.prototype.destroy = function(callback) {
  var doc = this;
  doc.whenNothingPending(function() {
    if (doc.wantSubscribe) {
      doc.unsubscribe(function() {
        doc.connection._destroyDoc(doc);
        callback();
      });
    } else {
      doc.connection._destroyDoc(doc);
      if (callback) callback();
    }
  });
};
```

ericyhwang:
Nate's comments from the PR review meeting:

> * Philosophy: In Share, server shouldn't be responsible for maintaining client state, that should be all handled by the client.
> * Not going to do promises any time soon. Code should always be async (e.g. `process.nextTick`) or always be sync, not both.
> * Waiting on the callback longer likely won't be an issue, since you generally won't be blocking on `destroy`. I'm leaning towards doing what is safe.
> * For this PR: Remove the `sync` workaround, make a separate PR for making `whenNothingPending` async.

gkubisa:
I made the requested changes and created a new [PR to fix whenNothingPending](https://github.com/share/sharedb/pull/222).

lib/client/doc.js
```diff
@@ -104,11 +104,20 @@ emitter.mixin(Doc);
 Doc.prototype.destroy = function(callback) {
   var doc = this;
   doc.whenNothingPending(function() {
-    doc.connection._destroyDoc(doc);
     if (doc.wantSubscribe) {
-      return doc.unsubscribe(callback);
+      doc.unsubscribe(function(err) {
+        if (err) {
+          if (callback) callback(err);
+          else this.emit('error', err);
```
nateps: `this` should be doc. I'll just go ahead and merge this change and make the fix, since we're close.

PR #205, opened 2018-04-24
Fix hasWritePending in op's callback

gkubisa:
`doc.hasWritePending()` should be false in the op's callback after the last operation is complete.

Additionally, checking for `this.inflightOp` first in `this.inflightOp && this.inflightOp.callbacks` avoids a TypeError which would happen, if [this _otApply](https://github.com/share/sharedb/blob/ea5053b8064876538c61a2284c4f931fa8b52c79/lib/client/doc.js#L861) failed and caused a hard rollback. Unfortunately, I don't have a test for this situation, as it's a rather obscure edge case.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -888,9 +888,10 @@ Doc.prototype._hardRollback = function(err) {
 };
 
 Doc.prototype._clearInflightOp = function(err) {
-  var called = callEach(this.inflightOp.callbacks, err);
-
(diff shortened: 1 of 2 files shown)


gkubisa:

ericyhwang:
Nate from the PR review meeting:

> Makes sense. I'll need to read through the rest of the code offline to check a few things [before merging].

curran:
Suggest to close as stale.

Issue #128, opened 2016-10-05
Add cursor synchronization to an editor


MrTin:
Hi ShareDB Team,

It's so fun to play around with this – thanks for sharing ShareDB to the open source community!

I was wondering how I could go about and add cursor sync to ShareDB. I'm trying out the rich-text example (that uses Quill) and was curious how you could make it look like Google Docs, where it points out another clients cursor, color and name. I was looking at https://github.com/hivejs/hive-plugin-cursor-broadcast-codemirror but I realized that you guys proabably have a better answer to my question.

Is this a good example to use middleware for? Am I thinking right here? ShareDB listens to the WS stream so won't I make it "crash" by sending my own commands (cursor position) in the stream?

My idea is the following:
- Create a middleware for authentication using JWT.
- During socket connection, store the color and name of the client.
- Create a middleware for synchronizing cursor position(?)

Thanks for any help!

Regards,
Martin

zag2art:
Have you seen the quill issue? - https://github.com/quilljs/quill/issues/918
Quill before 1.0 had multicursor module, the @benbro guy updated the module for 1.0.

MrTin:
Thanks! Will take a look :) If I understand the solution correct he is not keeping track of where I have my cursor but rather indicate where the latest `op` occured. Simple enough to me!

nateps:
This is high on my list of features to get in at some point, though it does require proper integration with OT and ideally support for ephemeral data sync.

## OT Integration

When you make edits locally, they need to offset the position of other users' cursors in your local client synchronously. For example, if your cursor is after mine and I start typing, each character I type should increase the position of your cursor by one character. Thus, the cursor UI does need to understand the OT type's operations and know how to adjust all the cursors being displayed by the edit ops that are being created locally and remotely. You can see this in the quill multi-cursor module here: https://github.com/benbro/quill/blob/multi-cursor/modules/multi-cursor.js#L76. As @zag2art pointed out, looks like this is going to get re-implemented in Quill for richtext and we should be able to use that out of the box.

Other OT types (such as strings in JSON0 or the text type) would need their own implementations of the same kind of logic for their own op formats.

## Ephemeral data

[...]

PR #207, opened 2018-04-26
Sync ephemeral presence data

gkubisa:
This PR adds OT-aware ephemeral data sync as described by @nateps [here](https://github.com/share/sharedb/issues/128#issuecomment-252152894).

# How It Works

## Server

The server-side changes are trivial because I reuse all the existing infrastructure for subscribing to docs and publishing ops. So, when you're subscribed to a document, you can publish and receive presence data. The server is essentially a PubSub system for presence data. Apart from that, its other important function is to broadcast _null_ presence data on behalf of the client when it unsubscribes or disconnects, so that its peers know the client is gone.

The only issue I see with this solution is that if a server crashes or PubSub (eg Redis) fails, then the _null_ presence would not be delivered and the peers would not know the client is gone. Considering that this is going to be rare in practice and user impact is likely minimal, I'm not sure, if it is worth the effort. Anyway, I think the solution is to simply expire peer presence data after some time, if they don't send regular heartbeat messages.

### Handling Current Presence Data

The current local (this client) and remote (peers) presence data is in `doc.presence`, which is also a part of the public API. Whenever any operation is applied to the local document, `doc.presence` is automatically transformed, so that for example a cursor would move forward by one character, if a user typed one character. In case of a hard rollback this data is wiped and syncing must start from scratch.

### Publishing Local Presence Data

Publishing is delayed until the document is subscribed to and all pending writes have completed. When we're finally ready to publish, we can simply send the local data stored in `doc.presence`.

When publishing for the first time, or after subscribe, or reconnect, or hard rollback, a special flag (`r` - `requestReplyPresence`) is set on the message to ask all peers to publish their own presence data as soon as possible. This allows new clients to initialize. This system could be optimized a bit by sending the "reply" presence only to the client who requested it, instead of broadcasting. I left it out for now because I'm not sure if it is worth the effort.

### Receiving Peer Presence Data

There are a few scenario to handle here...

1. If we get an out-of-sequence message, we ignore it.
2. If we get a _null_ presence data from the server (`version == null` - peer disconnected or unsubscribed), we clear the relevant data in `doc.presence`.
3. If we get presence data for a future version of the document, we try to fetch the missing ops first.
4. If we get presence data for the current version of the document, we transform it against any inflight and pending ops and then store in `doc.presence`.
5. If we get presence data for an older version of the document, we transform it against the old ops, inflight op and pending ops, and then store the result in `doc.presence`.

A few things to note:

- transforming presence against create, del or no-op makes presence data null
- full presence data is sent every time. It greatly simplifies the implementation, and should have minimal impact on performance, considering that presence data is expected to be small in terms of bytes.
- I cache past ops for 1 minute to enable transformation of presence data, if necessary. Considering that presence data is supposed to be real-time, I think we can safely discard that which arrives for versions more than 1 minute old.
- I cache received presence data for 1 minute, so that I can compare the sequence numbers of newly received presence data and reject out-of sequence messages. In practice this should eliminate the risk of race conditions without introducing a risk of a memory leak (caused by peers constantly sending presence data and disconnecting).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ node_js:
   - "10"
   - "8"
   - "6"
-script: "npm run jshint && npm run test-cover"
+script: "ln -s .. node_modules/sharedb; npm run jshint && npm run test-cover"
 # Send coverage data to Coveralls
 after_script: "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js"


lib/agent.js
```diff
@@ -582,3 +603,30 @@ Agent.prototype._createOp = function(request) {
...
+    if (err) return callback(err);
+    callback(null, { seq: presence.seq });
+  });
+};
+
+Agent.prototype._createPresence = function(collection, id, data, version, requestReply, seq) {
+  return {
+    a: 'p',
+    src: this.clientId,
+    seq: seq != null ? seq : this.maxPresenceSeq,
```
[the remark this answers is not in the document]
gkubisa: `!= null` checks for `null` AND `undefined`. `!== null` would check only for `null`.

test/mocha.opts
```diff
@@ -1,3 +1,4 @@
 --reporter spec
 --check-leaks
 --recursive
+--exit
```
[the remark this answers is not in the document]
gkubisa: Yes, I completely agree. The existing tests did not close everything properly, so I just added this flag after upgrading mocha to get the old behaviour and keep the tests passing. In this PR I focused on adding presence data sync. Fixing tests to close everything properly is a separate issue to me.

lib/agent.js
```diff
@@ -582,3 +603,30 @@ Agent.prototype._createOp = function(request) {
     return new DeleteOp(src, request.seq, request.v, request.del);
   }
 };
+
+Agent.prototype._presence = function(presence, callback) {
+  if (presence.seq <= this.maxPresenceSeq) {
+    return callback(new ShareDBError(4026, 'Presence data superseded'));
```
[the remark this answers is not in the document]
gkubisa: Fixed, thanks :+1:

lib/agent.js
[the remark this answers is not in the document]
dcharbonnier: right sorry I never use equality operator

rawkode:
I'd love to see this merged to the project. Has there been any discussion regarding this functionality?

curran:
This looks good to me!

The only remaining issue IMO is the merge conflicts. @gkubisa Would you be able to resolve these merge conflicts? Thanks!

rawkode:
@gkubisa If you don't have the time to rebase this, I'd be happy to pick it up

gkubisa:
I resolved those merge conflicts now and the branch is ready to merge IMHO.

curran:
@nateps I'm curious what your initial thoughts are on this PR? Thank you for your time.

gkubisa:
I've just open-sourced [Teamwork/ot-rich-text](https://github.com/Teamwork/ot-rich-text) to provide a concrete example of an OT type supporting presence data. We use it at [Teamwork.com](https://www.teamwork.com/) for real-time rich-text collaboration.

@nateps , would you be able to review/merge this PR some time soon, please?

curran:
@gkubisa It's awesome to see the great work!

I'm considering switching our dependency to `@teamwork/sharedb`, as this project is clearly suffering. This PR introduces such a useful feature, I'm surprised not a single maintainer has commented on it in 8 weeks.

gkubisa:
> I'm surprised not a single maintainer has commented on it in 8 weeks.

Yes, this is really disappointing. :-(

> I'm considering switching our dependency to @teamwork/sharedb

If you do switch, then you might also want to use the other related forks: https://www.npmjs.com/search?q=%40teamwork%2Fsharedb and https://www.npmjs.com/package/@teamwork/websocket-json-stream. The updated docs for OT types are here: https://github.com/teamwork/ot-docs.

Re the future of my fork, I'm going to add local undo/redo very soon. Afterwards I'll most likely just fix bugs and merge PRs.

dcharbonnier:
I'm in @gkubisa

brainkim:
😭 why can’t we all just get along 😤

dcharbonnier:
I agree @brainkim just for now we all get "alone" xD just looking for a way to get along

brainkim:
@gkubisa re: local undo/redo, if you do decide to publish something I‘d appreciate if it was it’s own module and not part of the sharedb codebase, b/c I think it would be relatively easy to keep separate and b/c I’d rather not have that functionality tied to a fork... Just a thought!

gkubisa:
@brainkim

> I think it would be relatively easy to keep (ShareDB and undo/redo) separate

It is certainly possible (I've done it) to implement undo/redo without modifying ShareDB, however, it requires undesirable compromises in the design of the public API. So, the only way I see myself publishing undo/redo is as an integral part of ShareDB.

> I’d rather not have that functionality tied to a fork

We're on the same page here. I'm not a fan of maintaining a fork, however, I feel like I have no other choice. I open PRs for all the fixes and features I add to ShareDB but some of them just sit on the list for months without even a comment from the maintainers. So, I'll have to continue with my fork until the PRs start getting merged.

rawkode:
Thanks for your work, @gkubisa; we've successfully migrated over to Teamwork's ShareDB. We hope maintenance of this project gets sorted in the coming conversations :+1:

curran:
@nateps @ericyhwang Is the idea behind this PR "approved" for ShareDB core?

curran:
@gkubisa I'm digging into this a bit deeper, and I'm wondering do you have any examples that use the presence feature?

I'm considering the cursor as well as selections. Looking through the [tests](https://github.com/Teamwork/sharedb/blob/master/test/client/presence.js), it's not clear to me how one might support selections (regions of selected text).

Thanks!

gkubisa:
Hi @curran ,

first of all a disclaimer, this feature has not gone through any serious review, so it's design might change in the future. One thing I might change one day is to implement presence on the connection level, which would contain things like user ID, etc. The per-document presence would then contain only what's relevant to that document, eg selection positions. Currently both of the above are managed on a document level. Ok, with this out of the way, let's go back to your questions...

Using `ot-rich-text` might be the path of least resistance indeed, if you don't already have a lot of code depending on another text type.

As you've already noticed, adding presence to existing types requires some new functions to be defined on the type `createPresence`, `transformPresence` and `comparePresence`. When and if presence makes it into ShareDB, I think the presence functions should be defined in the main repos of the existing types but for now I think a fork would be the most appropriate.

Adding the presence functions to [text](https://github.com/ottypes/text) and [rich-text](https://github.com/ottypes/rich-text) should be simple, as those types already define functions for transforming positions, so it's just a matter of using them appropriately in the implementation of the presence functions.

Adding presence to [json0](https://github.com/ottypes/json0) would definitely be harder, as you'd need to take into the account all json0 operations and map the presence to the correct text property in the json object. It's certainly doable but might be difficult.

> would it be possible to use the presence feature with JSON0 strings?

Yes, it would be possible, if you're willing to implement the presence functions for json0, which might be rather difficult.

> Or, is it required to create a new OT type to use presence?

The bottom line is that the presence functions mentioned above are defined. It's up to you how you want to achieve that.

> So, for use with plain text, it appears I'd need to fork either the text0 or text types and add the required methods. Is that correct?

Yes.

> The text I'm dealing with is a property of a JSON object using json0, so I wonder would I need to fork json0 itself? Not sure...

Yes.

Alternatively, you might be able to restructure your application so that the text is not embedded in json0. The benefit would be that you would not need to implement presence on json0 and your text type but only on the text type.

I hope it helps. Let me know, if you have any other questions.

curran:
@gkubisa Thank you for your thoughtful response! Much appreciated.

To summarize, this is the "tree of alternatives" I see here for directions forward:

 * Add presence to [json0](https://github.com/ottypes/json0)
 * Restructure my application so that the text is not embedded in `json0`,
   * then use [ot-rich-text](https://github.com/Teamwork/ot-rich-text#presence) (which has presence),
   * or add presence to [text](https://github.com/ottypes/text) and use that.

Food for thought!

ericyhwang:
Hey all,

Nate and I actually took a glance at this a few weeks ago, and I thought I posted our preliminary comments... but it looks like it didn't actually go through. :(

I've tried to reconstruct the notes below. I was going to start going through this in detail, but then I got shifted over to help on some important app performance work that bubbled up. That's still ongoing, and it'll likely be another couple weeks at the very least before I can get back to this.

****

The overall approach looked good! While we were chatting, we came up with a couple requests on the current PR and a couple additional ideas for the future.

Requests for the current PR:

* Add a ShareDB constructor option/flag for enabling presence. It looks like currently, some things like op caching could affect performance, even if you never call `doc.submitPresence`. It'd be good to not run that type of code if you're not using presence, like if none of your doc types don't support it.
* Nate would like all the presence-related info to be nested underneath a top-level property, to make it easier to inspect what's related to presence. The current presence map might be `doc.presence.currentDataByClient = {'src1': presenceData1, ...}`.

Ideas for future work:

* Add middleware hooks to allow doing things on the server like validating / modifying incoming presence - adding validation means someone who knows the wire protocol can't fake who they are - or doing a projection prior to pushing out to clients.
* Support custom presence metadata (non-transformed) that's handled by Share itself, so that you can have a common metadata for your app that's not dependent on doc type
* Presence opt-in based on collection
* Have a configurable cap on how many clients to track presence data for. Google Docs and such will punt you over to a read-only view if there are too many people active in a doc.

curran:
I'm leaning towards prototyping presence for JSON0 using the existing ShareDB API with a separate "presence" collection (parallel to an existing "documents" collection). The idea is that clients would submit ops to "presence" documents that include transforms for the cursors and selections of other users. Are there any downsides to this approach? Would it work?

Forking and modifying JSON0 to implement the presence methods defined in this PR feels like too heavy a lift.

gkubisa:
> I'm leaning towards prototyping presence for text0 (within json0) using the existing ShareDB API by using a separate "presence" collection (parallel to an existing "documents" collection). The idea is that clients would submit ops to "presence" documents that include transforms for the cursors and selections of other users. Are there any downsides to this approach? Would it work?

Storing presence (ephemeral data) in a database would incur unnecessary costs: latency related to communication with the database as well as extra storage and load on the database. Additionally, you'd still have to somehow detect when users go offline and remove their presence from the database.

Storing presence in a parallel collection would probably lead to issues with synchronizing presence with the main data. I'm also not sure how you'd like to handle the transformations of the presence data. Personally I wouldn't recommend the parallel collection approach due to its complexity.

If you really want to go down the route of storing presence in the database, I think storing the main data and the related presence in the same json0 document would be much easier than using a parallel collection. I'd imagine you'd still need to develop an OT type for the presence data and embed it in json0.

> Forking and modifying JSON0 to implement the presence methods defined in this PR feels like too heavy a lift.

Yes, it's not a trivial task but I think it would be the right thing to do. It would also need to be done only once and then it could be reused easily in any project.

curran:
Thanks @gkubisa for your thoughts here!

alecgibson:
NB: We have a repo that ties presence into Quill here: https://github.com/reedsy/quill-cursors

Could be interesting to compare implementations.

ericyhwang:
Re presence for json0, it'd be nice to have full support eventually. That is a non-trivial chunk of work, as @curran points out.

I wonder if it's feasible to do a partial implementation initially, like only support presence for top-level text0 fields first and error on anything else. I haven't dove too deeply yet into this PR to know if that's advisable.

gkubisa:
> I wonder if it's feasible to do a partial implementation initially, like only support presence for top-level text0 fields first and error on anything else.

Given that json0 is so generic, perhaps presence on content embedded in json0 is all we could reasonably do? In that case json0 would just delegate presence handling to the appropriate embedded types. Probably the only presence transformations in json0 itself would be related to list operations, so that the correct embedded type would receive the presence later. Anyway, it seems very doable with some effort.

curran:
That makes sense. I'm interested in adding presence for `text0` fields within nested `json0` objects.

curran:
Hey check this out - JSON0 presence has been implemented by @houshuang ! Here's the PR https://github.com/ottypes/json0/pull/25

houshuang:
In fact, that PR was submitted in error - I was trying to submit against a different upstream :) I thought I closed it, but I guess I didn't. Anyway we are absolutely happy to share this work, just not quite done yet (but feedback welcome!). Here's a quick demo video with Quill and presence/shared cursors. https://www.youtube.com/watch?v=CVjH56Q18cU

curran:
I started working on json0 presence here https://github.com/datavis-tech/json0/pull/2

curran:
I asked @nateps et. al. "What's your take on presence?" during the last video call PR review meeting, and the ensuing conversation was completely new information to me. In summary, Nate expressed concern for scalability, and explained the problem as follows:

 * The server does not cache presence data.
 * Every time a new user joins and requests up-to-date presence, the server asks all other clients for their presence, resulting in a client -> server -> remote client -> server -> client round trip, for all remote clients.
 * This is fine for a few users, but if thousands of users join, performance would be problematic.
 * Because of this, namespacing the implementation internally would leave room for future implementations that have better performance characteristics.

I haven't dug into this PR enough to verify that all of this is true, but I just wanted to bump the conversation here with the latest thoughts from the Lever side.

My gut feeling is that the server could/should be made to cache presence data for all documents, probably in memory for a bounded period of time.

gkubisa:
Nate's explanation is 100% valid. I implemented it the way it is because:

- I assumed that the number of users editing the same document concurrently is not going to be huge in practice - max a few dozen. It is a valid assumption for the documents we manage in my company.
- Caching and transforming presence on the server and reconciling it with the clients seemed very complex.
- I had concerns about the server-side scalability of caching and transforming presence on the server.
- I wanted to avoid stale presence data persisting on servers in case one server crashed and did not clean up the presence it managed.
- I did not want to store presence as part of operations due to performance reasons.

houshuang:
I started a write-up to better understand how this works here: https://docs.google.com/document/d/1lxBa74_sfkhaLg0s7S_8yIcboaNZvE6TNHtUYnhAlNc/edit# - note that this is intermingling a bit the concerns between the ShareDB presence mechanisms, how we are proposing to implement ot-json0 with subtypes etc. It already benefited from some clarifications from @gkubisa (thanks!)

lib/client/doc.js
```diff
@@ -104,11 +139,24 @@ emitter.mixin(Doc);
 Doc.prototype.destroy = function(callback) {
   var doc = this;
   doc.whenNothingPending(function() {
-    doc.connection._destroyDoc(doc);
     if (doc.wantSubscribe) {
-      return doc.unsubscribe(callback);
+      doc.unsubscribe(function(err) {
```
curran: This would be a good place to use the ShareDB constructor presence flag to conditionally use the more complex `unsubscribe` handler. Here's the original commend from @ericyhwang & @nateps :

> Add a ShareDB constructor option/flag for enabling presence. It looks like currently, some things like op caching could affect performance, even if you never call doc.submitPresence. It'd be good to not run that type of code if you're not using presence, like if none of your doc types don't support it.

lib/client/doc.js
```diff
@@ -186,12 +234,14 @@ Doc.prototype.ingestSnapshot = function(snapshot, callback) {
   if (this.version > snapshot.v) return callback && callback();
 
   this.version = snapshot.v;
+  this.cachedOps.length = 0;
```
curran: `if(ShareDB constructor presence flag)`

lib/client/doc.js
```diff
@@ -186,12 +234,14 @@ Doc.prototype.ingestSnapshot = function(snapshot, callback) {
...
   var type = (snapshot.type === undefined) ? types.defaultType : snapshot.type;
   this._setType(type);
   this.data = (this.type && this.type.deserialize) ?
     this.type.deserialize(snapshot.data) :
     snapshot.data;
   this.emit('load');
+  this._processAllReceivedPresence();
```
curran: `if(ShareDB constructor presence flag)`

lib/client/doc.js
```diff
@@ -257,6 +309,7 @@ Doc.prototype._handleSubscribe = function(err, snapshot) {
   if (this.wantSubscribe) this.subscribed = true;
   this.ingestSnapshot(snapshot, callback);
   this._emitNothingPending();
+  this.flush();
```
curran: It's not clear why `this.flush()` had to be added here.

lib/client/doc.js
```diff
@@ -318,7 +379,9 @@ Doc.prototype._handleOp = function(err, message) {
   }
 
   this.version++;
+  this._cacheOp(serverOp);
```
curran: This scattered logic could be better grouped.

```js
if(ShareDB constructor presence flag){
  var serverOp = {
    src: message.src,
    time: Date.now(),
    create: !!message.create,
    op: message.op,
    del: !!message.del
  };
  this._cacheOp(serverOp);
  this._processAllReceivedPresence();
}
```

lib/client/doc.js
```diff
@@ -426,14 +494,22 @@ function pushActionCallback(inflight, isDuplicate, callback) {
...
+//
+// If there are no pending ops, this method sends the pending presence data, if possible.
 Doc.prototype.flush = function() {
-  // Ignore if we can't send or we are already sending an op
-  if (!this.connection.canSend || this.inflightOp) return;
+  if (this.paused) return;
 
-  // Send first pending op unless paused
-  if (!this.paused && this.pendingOps.length) {
+  if (this.connection.canSend && !this.inflightOp && this.pendingOps.length) {
```
curran: The changes in logic changes here seem like minor cleanup unrelated to the presence feature. If so, these could be done in a separate PR, to more cleanly isolate the presence changes.

lib/client/doc.js
```diff
@@ -426,14 +494,22 @@ function pushActionCallback(inflight, isDuplicate, callback) {
...
     this._sendOp();
   }
+
+  if (this.subscribed && !this.inflightPresence && this.pendingPresence && !this.hasWritePending()) {
```
curran: `if(ShareDB constructor presence flag)`

lib/client/doc.js
```diff
@@ -868,29 +957,54 @@ Doc.prototype._rollback = function(err) {
 };
 
 Doc.prototype._hardRollback = function(err) {
-  // Cancel all pending ops and reset if we can't invert
-  var op = this.inflightOp;
-  var pending = this.pendingOps;
+  var callbacks = [];
+  if (this.inflightPresence) {
```
curran: Could be better grouped under `if(ShareDB constructor presence flag)`

lib/client/doc.js
```diff
@@ -868,29 +957,54 @@ Doc.prototype._rollback = function(err) {
...
+  }
+  for (var i = 0; i < this.pendingOps.length; i++) {
+    callbacks.push.apply(callbacks, this.pendingOps[i].callbacks);
+  }
+
   this._setType(null);
   this.version = null;
   this.inflightOp = null;
   this.pendingOps = [];
+  this.cachedOps.length = 0;
```
curran: `if(ShareDB constructor presence flag)`

lib/client/doc.js
```diff
@@ -908,3 +1022,281 @@ function callEach(callbacks, err) {
   }
   return called;
 }
+
+// *** Presence
+
+Doc.prototype.submitPresence = function (data, callback) {
```
curran: All of these methods could be definde only `if(ShareDB constructor presence flag)`

package.json
```diff
@@ -16,7 +16,7 @@
     "expect.js": "^0.3.1",
     "istanbul": "^0.4.2",
     "jshint": "^2.9.2",
-    "mocha": "^3.2.0",
+    "mocha": "^5.1.1",
```
curran: Mocha upgrade should be separate PR.

gkubisa:
Thanks for the review @curran . Unfortunately, I currently have no time to work on it. Feel free to continue the work on this feature in your own fork. :-)

curran:
Thanks for letting me know @gkubisa . I may be able to continue this work.

I did get presence working in a [fork of json0](https://github.com/datavis-tech/json0), continuing the work started by @houshuang .

Here's a demo app: https://github.com/datavis-tech/json0-presence-demo

presence

selections

curran:
@gkubisa Are there any presence-related changes in the Teamwork fork of ShareDB that are not in this PR? I'm starting in on the work of a new PR from this one, and wanted to make sure it's not missing anything. Thanks!

gkubisa:
Great that you're taking it on. This branch should contain all presence-related changes, @curran .

curran:
Excellent. Thanks for confirming! Work has commenced in #286 .

curran:
Suggest to close as stale. Other presence PR has landed https://github.com/share/sharedb/pull/322

PR #209, opened 2018-05-12
Add new middleware events for document reads

l8on:
* Trigger `queryDocs` or `docBulk` middleware action when many documents are fetched at once. * Only trigger `doc` action when retrieving a single doc.

*WIP*
There are no tests, this is not ready for merging. I would like feedback on the general approach before I dive into testing and making this more production ready.

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -195,55 +195,45 @@ Backend.prototype._sanitizeOpsBulk = function(agent, projection, collection, ops
 };
 
 Backend.prototype._sanitizeSnapshot = function(agent, projection, collection, id, snapshot, callback) {
-  try {
-    this._projectSnapshot(projection, snapshot);
-  } catch (err) {
-    return callback(err);
-  }
-
-  var backend = this;
-  this.trigger('docs', agent, {collection: collection, id: null, snapshots: [snapshot]}, function(err) {
-    if (err) { return callback(err); }
-    backend.trigger('doc', agent, {collection: collection, id: id, snapshot: snapshot}, callback);
-  });
-};
-
-Backend.prototype._projectSnapshot = function(projection, snapshot) {
   if (projection) {
-    projections.projectSnapshot(projection.fields, snapshot);
+    try {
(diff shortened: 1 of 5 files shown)


l8on:
@nateps

lib/backend.js
```diff
@@ -199,22 +249,40 @@ Backend.prototype._sanitizeSnapshot = function(agent, projection, collection, id
     try {
       projections.projectSnapshot(projection.fields, snapshot);
     } catch (err) {
-      return callback(err);
+      return process.nextTick(function() {
```
nateps: If we chance this to be async, I think it is important that the callback below is async, too. Trigger itself is sync, so the call could be sync or async depending on the middleware that people write. Probably safer to make it all async, but let's make sure that we cover all cases if we are going to make a change.

Also, it might be better to make the sync callback change as a separate sharedb release just in case it has an impact on others' code. I'd like to do this, but I think it may be best to do separately.

test/middleware.js
```diff
@@ -47,12 +48,246 @@ describe('middleware', function() {
...
+
+  });
+
   describe('submit', function() {
 
     it('gets options passed to backend.submit', function(done) {
       this.backend.use('submit', function(request, next) {
         expect(request.options).eql({testOption: true});
-        done();
+        next();
```
nateps: Is this trying to fix a problem? It seems like an atypical pattern, but i'm trying to understand it.

nateps: This test could have been more cleanly written as:

```
it('gets options passed to backend.submit', function(done) {
      this.backend.use('submit', function(request, next) {
        expect(request.options).eql({testOption: true});
        next();
      });
      var op = {create: {type: types.defaultType.uri}};
      var options = {testOption: true};
      this.backend.submit(null, 'dogs', 'fido', op, options, done);
    });
```

Perhaps you can use that pattern in the other tests, too

l8on:
@nateps I've made the requested changes. LMK if there is anything else amiss.

PR #217, opened 2018-06-25
Allow op to be a non-object

gkubisa:
On the client side, non-object ops are processed successfully.

When ops are received by a server, they must not be `null`.

https://github.com/share/sharedb/blob/68bde0047eccd4709bd41e64f380ea7841f9864f/lib/ot.js#L30-L32

However, when ops are applied by the server, they must be `object`.

https://github.com/share/sharedb/blob/68bde0047eccd4709bd41e64f380ea7841f9864f/lib/ot.js#L105

This PR fixes the inconsistency in validation and allows ops to be non-objects.


ericyhwang:
@nateps from the PR review meeting says this is good to merge.

Though, I noticed there's a package-lock.json file committed, and Share's not using package lock files yet. @gkubisa mind removing that real quick before we merge?

gkubisa:
I've removed `package-lock.json` for now but I think we should start using it. I don't think it causes any real issues but it keeps the repo more stable for CI and ShareDB devs.

nateps:
Awesome, thanks for the contribution! Looks great. ⚡️

Issue #218, opened 2018-06-26
Document version fetching


alecgibson:
# Problem statement

As a ShareDb client, I want to fetch a read-only snapshot of my document at a given version number or timestamp.

# Motivation

By the time we're going to all the effort of storing a set of deltas between document versions, it's only natural that a client would wish to leverage this power to view a document at any point in its history.

The problem statement mentions fetching a document at a given timestamp, because it is far more natural to request a document at a particular time, than at a given (arbitrary) version number.

# API

The proposed API is to add two methods to the [`Doc` class][1]:

 - `Doc.prototype.fetchVersion(version, callback)` takes `version`, which is a `number`, and recreates the snapshot up to that version number. The result is stored in `doc.data`
 - `Doc.prototype.fetchAtTime(time, callback)` takes `time`, which is a `Date`, and recreates the snapshot using ops whose timestamps are up-to-and-including that `Date`. The result is stored in `doc.data`

# Implementation details

## Data flow

The request for the document version will be submitted like the existing `fetch` function - by submitting an event to the server, and attaching a callback.

The message will be picked up by `Agent._handleMessage`, which will then leverage `Backend.getOps` to fetch the requested ops.

We may need to make a small change to `Backend.getOps` to let us request metadata from the backend using the `options` object. As discussed in [this Pull Request][5], this will be done in such a way that keeps the `option` object out of the public API (probably by creating an internal `Backend._getOps` method that can take an `options` object, and calling it with `null` from `Backend.getOps`).

# Discussion

## Read-only snapshots

Using the `Doc` class is potentially a leaky concern, given that it will also have `Doc.prototype.submitOp`, which doesn't really make sense when fetching an historical document.

A possible alternative could be to expose these functions instead on the [`Connection` class][2]? That way it should be very clear that the consumer is receiving a snapshot, and not a full-blown `Doc`?

# Out of scope

The following possibilities are deemed out-of-scope for the initial solution.

## Optimising for reversible types

Making a type reversible [is optional][3]. As such, any solution *must* at least be able to construct a document from its initial version, and build up. However, given the nature of documents, it is highly likely that users will wish to return to more recent versions, where it will probably be faster to start from the current version and work backwards.

This is deemed out-of-scope.

## Caching ops

Fetching ops [can be expensive][4]. Ideally we would cache the ops for a given document, and - so long as the requested version/timestamp is lower than the latest op - then we could simply read ops back from the cache.

This is deemed out-of-scope.

## Other starting snapshot optimisations

[...]

PR #220, opened 2018-06-29
Add support for fetching a particular version of a snapshot

alecgibson:
This change follows on from discussion in [this issue][1]. Its primary
aim is to allow clients to fetch an historical version of a document
(a snapshot), by providing either:

  - a version `number`, or
  - a desired `Date`

The entry-point for this feature is added to the `Connection` class,
deliberately separate from the `Doc`, because `Doc` is concerned with
"live" document actions such as subscribing and submitting ops,
whereas fetching an historical version of a document should not be
associated with these ideas.

The feature is called with:

```javascript
connection.getSnapshot(collection, id, version, callback): void;
```

The details of the interface are detailed in the README, and in the
code documentation.

This change includes support for projections, and use of the
`readSnapshots` middleware. It also hooks into `Connection`'s
`hasPending` method.

Performance optimisations are deemed out-of-scope for this change (see
the [issue][1] for more details).

Note that this change also adds a development dependency on [`lolex`][2]
which is used for mocking the time.

[1]: https://github.com/share/sharedb/issues/218
[2]: https://github.com/sinonjs/lolex


alecgibson:
Things that may need discussion:

  1. Function signature
  2. Treatment of deleted documents
  3. Anything I've missed

## Function signature

In the original [issue](https://github.com/share/sharedb/issues/218), we had discussed the following signature:

```javascript
Connection.prototype.getSnapshot(collection: string, id: string, { version: number, timestamp: Date }, (Error, Snapshot) => void): void;
```

However, in this change I have implemented the slightly different:

```javascript
Connection.prototype.getSnapshot(collection: string, id: string, version: (number|Date), (Error, Snapshot) => void): void;
```

This is because `version` should really be either/or - you should never be able to have both a version number defined and a `Date`. Also, if the idea is to ease further types of argument, then dealing with the possible combinations of arguments grows prohibitive and ugly.

## Treatment of deleted documents

For now, I have added a `deleted` flag to the returned snapshot, which indicates if the requested snapshot was deleted. However, I was unsure of how this should act. Other possibilities include:

  - if the last op of a doc is a deletion (ie the doc is currently deleted), then return a deleted document error on any requested version
  - if the requested op of a doc is a deletion, then return a deleted document error

## Anything I've missed

I'm very new to this codebase, so I've doubtlessly missed something. I've tried to implement everything I could see (including projections, middleware and hooking into `Connection.hasPending`), but please let me know if I've missed anything.

README.md
```diff
@@ -212,6 +212,29 @@ changes. Returns a [`ShareDB.Query`](#class-sharedbquery) instance.
...
   All other options are passed through to the database adapter.
 
+`connection.getSnapshot(collection, id, version, callback): void;`
+Get a read-only snapshot of a document at the requested version.
+
+* `collection` _(String)_
+  Collection name of the snapshot
+* `id` _(String)_
+  ID of the snapshot
+* `version` _(number | Date)_
```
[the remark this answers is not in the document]
alecgibson: By checking `typeof` for `number` and `instanceof` for `Date`: https://github.com/share/sharedb/pull/220/files#diff-adc2354b96b029a19a73805a6f3f7fb4R19

test/client/snapshot-request.js
```diff
@@ -0,0 +1,333 @@
...
+    });
+
+    describe('with a registered projection', function () {
+      beforeEach(function () {
+        backend.addProjection('bookTitles', 'books', { title: true });
+      });
+
+      it('applies the projection to a snapshot', function (done) {
+        backend.connect().getSnapshot('bookTitles', 'don-quixote', 2, function (error, snapshot) {
+          if (error) done(error);
```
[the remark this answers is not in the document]
alecgibson: Good catch

README.md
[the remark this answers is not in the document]
dcharbonnier: So you don't support a date passed has a number. Ok, from my understanding of the ticket this was supposed to be implemented but it's not an issue  for me :+1:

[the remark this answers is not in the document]
alecgibson: Yeah I don't think there's any way we could do that sensibly without having two separate methods (which was my original proposal).

There's no real way to tell the difference between a really, really, really high version number and a really, really, really old diff.

lib/backend.js
```diff
@@ -568,6 +573,76 @@ Backend.prototype.getChannels = function(collection, id) {
...
+
+      if (typeof timestamp === 'number' && op.m.ts > timestamp) {
+        op = ops[index - 1];
+        break;
+      }
+
+      if (op.create) {
+        type = types.map[op.create.type];
+        if (!type) return callback({ code: 4008, message: 'Unknown type' });
+        snapshot = op.create.data;
```
gkubisa: I think it should be ```snapshot = type.create(op.create.data);```

lib/backend.js
```diff
@@ -568,6 +573,76 @@ Backend.prototype.getChannels = function(collection, id) {
...
+        deleted = false;
+      } else if (op.del) {
+        snapshot = undefined;
+        deleted = true;
+      } else {
+        snapshot = type.apply(snapshot, op.op);
+      }
+    }
+
+    if (!snapshot && !deleted) return callback({ code: 4015, message: 'Document does not exist' });
```
gkubisa: I don't think snapshot needs to be verified... the `create` and `apply` functions should guarantee that it's always correct for the given type. Additionally, in this context we're applying operations which were applied successfully before, which gives us even more confidence in the correctness of the result. Furthermore, `null`, `0` and `false` are all valid snapshots for ot-json.

lib/backend.js
```diff
@@ -568,6 +573,76 @@ Backend.prototype.getChannels = function(collection, id) {
...
+    var type;
+    var snapshot;
+    var deleted;
+    var op;
+
+    for (var index = 0; index < ops.length; index++) {
+      op = ops[index];
+
+      if (typeof timestamp === 'number' && op.m.ts > timestamp) {
+        op = ops[index - 1];
```
gkubisa: If `index === 0`, `op` will become `undefined`, which will cause a `TypeError` further down in this function. I'd actually prefer tracking `version` and `timestamp` as local variables (just like `type`, `snapshot`, etc) and limit the usage of `op` to within the loop.

lib/backend.js
```diff
@@ -568,6 +573,76 @@ Backend.prototype.getChannels = function(collection, id) {
...
+
+    var type;
+    var snapshot;
+    var deleted;
+    var op;
+
+    for (var index = 0; index < ops.length; index++) {
+      op = ops[index];
+
+      if (typeof timestamp === 'number' && op.m.ts > timestamp) {
```
gkubisa: I think it would be good to store the result of `typeof timestamp === 'number'` in a local variable, to ensure that the expression is not evaluated every time and hurting performance.

lib/client/connection.js
```diff
@@ -584,3 +597,44 @@ Connection.prototype._firstQuery = function(fn) {
...
+  }
+};
+
+/**
+ * Get a read-only snapshot at a given version or time
+ *
+ * @param collection - the collection name of the snapshot
+ * @param id - the ID of the snapshot
+ * @param version - the version number, or Date of the snapshot to fetch. If an exact version or Date match is not made,
+ *   then the next lowest version is returned. ie if a document has 6 versions, asking for v7 will return v6. If ops
```
gkubisa: "then the next lowest version is returned" - should it be "then the next lower version is returned"?

lib/client/connection.js
```diff
@@ -584,3 +597,44 @@ Connection.prototype._firstQuery = function(fn) {
...
+ *   were submitted at 02:00 and 03:00, then asking for a Date at 02:30 will return the 02:00 version.
+ * @param callback - (error, snapshot) => void, where snapshot takes the following schema:
+ *
+ * {
+ *   collection: string; // collection name of the snapshot
+ *   id: string;         // ID of the snapshot
+ *   version: number;    // version number of the snapshot
+ *   timestamp: number;  // the UNIX timestamp of the snapshot
+ *   deleted: boolean;   // true if the returned version is a deleted snapshot
```
gkubisa: Replace `deleted` with `type`, as in README.md.

lib/client/snapshot-request.js
```diff
@@ -0,0 +1,78 @@
...
+  this.ready = false;
+  this.sent = false;
+}
+
+SnapshotRequest.prototype.parseVersion = function (version) {
+  if (typeof version === 'number') {
+    this.version = version;
+  } else if (version instanceof Date) {
+    this.timestamp = version.getTime();
+  } else if (!version) {
```
gkubisa: Would `version == null` be more appropriate?

lib/client/snapshot-request.js
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  this.connection.send(message);
+  this.sent = true;
+};
+
+SnapshotRequest.prototype._onConnectionStateChanged = function () {
+  if (this.connection.canSend && !this.sent) {
+    this.send();
+  }
```
gkubisa: If `!this.connection.canSend`, then probably set `this.sent` to `false`, so that the request would be resent after re-connection.

lib/client/connection.js
```diff
@@ -584,3 +597,44 @@ Connection.prototype._firstQuery = function(fn) {
...
+ *   data: any;          // the snapshot
+ * }
+ *
+ */
+Connection.prototype.getSnapshot = function(collection, id, version, callback) {
```
gkubisa: Could `version` be an optional param?

lib/client/snapshot-request.js
alecgibson: It starts as `false`, and is only changed to `true` at the end of `send()`.

alecgibson: Aha cool. Still trying to piece together all the data flows around this library! Some of it's pretty confusing.

alecgibson:
@dcharbonnier & @gkubisa I've addressed all of your comments. Could you please re-review?

README.md
```diff
@@ -212,6 +212,29 @@ changes. Returns a [`ShareDB.Query`](#class-sharedbquery) instance.
 * `options.*`
   All other options are passed through to the database adapter.
 
+`connection.getSnapshot(collection, id, version, callback): void;`
+Get a read-only snapshot of a document at the requested version.
```
curran: Nice documentation.

lib/backend.js
```diff
@@ -568,6 +573,80 @@ Backend.prototype.getChannels = function(collection, id) {
...
+    });
+  });
+};
+
+Backend.prototype._getSnapshot = function (agent, index, id, version, timestamp, callback) {
+  version = (typeof version === 'number' && isFinite(version)) ? version + 1 : null;
+  var options = { metadata: true };
+  var timestampIsNumber = typeof timestamp === 'number';
+
+  this._getOps(agent, index, id, 0, version, options, function (error, ops) {
```
curran: It looks like if you specify a Date rather than a version number, _all_ the ops are loaded into memory, but if you specify a version number, _only_ the ops until that version are loaded into memory, is that right?

If this is the case, it may be worth mentioning the overhead of passing in a Date in the docs.

lib/client/connection.js
curran: It looks like it is an optional param, with test coverage for the `undefined` case: https://github.com/share/sharedb/pull/220/files#diff-9ca1c4bd7a12ccd08af041e04970dfe7R146

lib/backend.js
alecgibson: Yes, that's exactly right. As far as I'm aware, with the current definition for `getOps`, this is unavoidable, right? I'll update the docs.

curran: Yes, with the current definition for `getOps`, it appears to be unavoidable.

Although, conceivably, one could optimize that case by passing the date into the database driver, so the database query itself could specify "less than this date" (as the timestamps are there on the ops and queryable). That feels like a rabbit hole though, and would require changing the database driver API, and then implementing the change in the actual drivers. Maybe something to think about for the future, once this first iteration lands.

Thanks for updating the docs!

gkubisa: In fact, it looks like all ops are always loaded into memory, see https://github.com/share/sharedb-mongo/blob/master/index.js#L362, so I'd say we don't need a specific warning about getting a snapshot for a `Date`. https://github.com/share/sharedb-mongo will need to be optimized but that would be a separate piece of work.

alecgibson: Fine - will reset

curran: @gkubisa Oh wow, that's surprising! I would have thought sharedb-mongo would query against the version. Good catch.

curran: @alecgibson This fork is active https://github.com/teamwork/sharedb-mongo/ . I'm using that fork currently as my dependency, instead of the original https://github.com/share/sharedb-mongo.

I opened a new issue just now **Optimization opportunity: use "to" in getOps query** https://github.com/share/sharedb-mongo/issues/61

test/client/snapshot-request.js
```diff
@@ -0,0 +1,386 @@
...
+        expect(snapshot).to.eql(v2);
+        done();
+      });
+    });
+
+    it('can call without a callback', function () {
+      backend.connect().getSnapshot('books', 'don-quixote');
+    });
+
+    it('errors if the version is -1', function (done) {
```
gkubisa: > returns an empty snapshot when trying to fetch a snapshot before the document existed

I think this is great. :+1: 

For consistency, shouldn't we also return an empty snapshot, if asked for a version before the document existed (negative version)?

lib/backend.js
```diff
@@ -568,6 +573,80 @@ Backend.prototype.getChannels = function(collection, id) {
...
+    if (error) return callback(error);
+
+    var type;
+    var snapshot;
+    var fetchedTimestamp;
+    var fetchedVersion;
```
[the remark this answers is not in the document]
alecgibson: I'm not sure we should initialise `fetchedTimestamp` and `fetchedVersion` to these values. If the version can't be found, then we'll erroneously tell the consumer that we fetched things with these values, when we didn't.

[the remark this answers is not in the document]
alecgibson: (Also the linter calls me out for initialising to `undefined`, which I think is a fair thing for it to complain about)

alecgibson:
@gkubisa I've addressed your comments (apart from the variable declaration - discussed above)

lib/backend.js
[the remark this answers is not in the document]
gkubisa: - `var type = null;` - I suggested it only to simplify the implicit type of the variable `null | |OtType`, rather than `undefined | null | OtType`. Not a big deal either way.
- `var snapshot = undefined;` - just to be explicit. I don't mind - it can be left unassigned.
- `var fetchedVersion = 0;` - I think the version should be 0 for nonexistent snapshots for consistency with `Doc` and https://github.com/share/sharedb#data-model, especially this part of the docs: "ShareDB implicitly has a record for every document you can access.". I think it's also a good practice to stick to a single data type for each variable whenever possible. In the case of `version` I think it is possible and quite natural to have `version === 0` for nonexistent snapshots - the same as `Doc.version` for nonexistent docs.
- `var fetchedTimestamp = timestamp;` - similar to `fetchedVersion` - keep the data type the same (number). The user would now that the snapshot does not exist by checking that `version === 0`. Actually, it might be better to do `var fetchedTimestamp = 0;`, in case `timestamp` is not specified. Additionally, `timestamp` would be meaningless for nonexistent snapshots anyway.

gkubisa:
I've been thinking about requesting snapshots for future timestamps and versions past the latest version, and I'm not sure, if the current approach is right... Currently we'd just return the latest available snapshot, however, that snapshot could be different on subsequent requests as more operations are recorded. In my opinion this is at odds with the idea of immutable snapshots. Could we return an error instead?

I would still keep the current behaviour (return the latest snapshot), if the `version` param of `getSnapshot` is omitted or `null`. In terms of implementation, this case could be trivially optimized be just loading the latest snapshot from the database, instead of applying all operations in order.

dcharbonnier:
I would even remove the date possibility, this create confusion, there should be an other api to request 
{snapshotId, date}[] but there is no need to implement it on this PR, the snapshoot id is enough for this one IMO

alecgibson:
> I would even remove the date possibility

This change is completely useless to me if we don't have timestamps. I'm happy to break this apart into two separate methods, like `getSnapshotByVersion` and `getSnapshotByTimestamp`. I personally prefer this sort of approach, because it helps to alleviate aforementioned confusion.

> requesting snapshots for future timestamps and versions past the latest version

I realise this is odd. It's certainly not idempotent. However, I would fully expect to be able to request a snapshot for the current time, and get a valid snapshot back (even if an op hasn't been recorded in some time).

How about this behaviour?

 - if the version is higher than the current document version, return an `Invalid version` error
 - if the timestamp is before `Date.now()`, existing behaviour (return a snapshot if a relevant timestamp is found, or else return current version)
 - if the timestamp is after `Date.now()`, return an `Invalid version` error

gkubisa:
> This change is completely useless to me if we don't have timestamps.

That's what I thought. :-)

> I'm happy to break this apart into two separate methods, like getSnapshotByVersion and getSnapshotByTimestamp.

I'd be ok with two separate methods too but could the one which takes in `version` be called just `getSnapshot`? It's because getting a snapshot by version is the most natural way to get a snapshot and it's unlikely to change, whereas we might come up with a nicer API in the future, which would allow retrieving snapshots by timestamp and possibly render `getSnapshotByTimestamp` obsolete.

> I would fully expect to be able to request a snapshot for the current time, and get a valid snapshot back (even if an op hasn't been recorded in some time).

Omitting the version/timestamp or setting it to `null` would always give you the latest/current snapshot.

> How about this behaviour? (...)

Yes, that makes sense. :+1:

The only thing to keep in mind though is that the client and server time might differ, so that if the client's clock is ahead of the server's clock, the server might return an error, if the client requests the snapshot for *its* current time.

alecgibson:
@gkubisa I've split it into two methods, but kept the internals basically the same. It also errors when asking for a snapshot after the current time or version. "Current time" is determined by the server.

gkubisa:
Great, thanks @alecgibson !

From my side, the only remaining issue is https://github.com/share/sharedb/pull/220/files#r201677449.

Hopefully this PR will be merged soon. :+1:

lib/backend.js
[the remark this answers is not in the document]
alecgibson: > I think it is possible and quite natural to have version === 0 for nonexistent snapshots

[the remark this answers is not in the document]
alecgibson: For what it's worth, the other reason I don't like this is that there's no easy way to tell if the snapshot you got back is valid. For example, if you asked for a `timestamp`, you could check if `timestamp === 0`, but (I suppose) you could have asked for `timestamp: 0`. You could also check `type === null`, but this is `true` if the snapshot is deleted, but still valid. But if this is what you want, I'll get a change pushed.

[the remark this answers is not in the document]
gkubisa: >  It's possible to call this method and get two snapshots claiming to be version === 0, but with different data.

I don't see how it would be possible - `version` is `0` only if we don't process any ops, and in that case `data` is always `undefined`.

> For what it's worth, the other reason I don't like this is that there's no easy way to tell if the snapshot you got back is valid. For example, if you asked for a timestamp, you could check if timestamp === 0, but (I suppose) you could have asked for timestamp: 0. You could also check type === null, but this is true if the snapshot is deleted, but still valid. But if this is what you want, I'll get a change pushed.

*I assume that by "valid snapshot" you mean that a document existed at a specific point in time.*

Your reasoning above is correct and the only reliable way to know that a document did not exist at a particular time is to check for `version === 0`. So, you still get the same information as before the change I suggested but in a slightly different way.

lib/client/connection.js
```diff
@@ -584,3 +597,87 @@ Connection.prototype._firstQuery = function(fn) {
...
+  for (var id in this.snapshotRequests) {
+    var snapshotRequest = this.snapshotRequests[id];
+    if (fn(snapshotRequest)) {
+      return snapshotRequest;
+    }
+  }
+};
+
+/**
+ * Get a read-only snapshot at a given version or time
```
gkubisa: Please remove timestamp references from the `getSnapshot` docs.

lib/client/snapshot-request.js
```diff
@@ -0,0 +1,67 @@
...
+  this.connection = connection;
+  this.id = id;
+  this.collection = collection;
+  this.callback = callback;
+
+  this.ready = false;
+  this.sent = false;
+}
+
+SnapshotRequest.prototype.hasPending = function () {
```
gkubisa: `ready` and `hasPending` are not needed because all requests in `Connection.snapshotRequests` are pending.

lib/backend.js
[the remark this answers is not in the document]
alecgibson: > version is 0 only if we don't process any ops, and in that case data is always undefined

The creation op is v0.

alecgibson:
> thanks for bearing with me and making all the changes!

No worries. Sorry if I seem impatient at times! It's challenging juggling this with my job, and I'm just eager to get this feature in.

> there might be a better way to achieve the same result - get a snapshot given a timestamp

I'm all ears. Especially if there's a more efficient way of doing it!

> An alternative would be to return a Snapshot synchronously

In principle I'm very okay with anything that removes callbacks, but I have these reservations:

  - producing a snapshot on a very large set of ops can take a very long time (10s of seconds, as I mentioned in my [PR on `sharedb-mongo`][1]), and I don't think we should block the client for anywhere near that length of time
  - it runs against the general approach taken elsewhere in the library, and also against the general JavaScript asynchronous spirit, which feels weird to me. Using a callback is an important signal to the consumer that this is an operation that could take a while, and they should be ready to deal with that (with whatever shiny spinners/loaders/apologetic messages/gifs/etc. they feel they need)

[1]: https://github.com/share/sharedb-mongo/pull/58

gkubisa:
Sorry, I meant returning the `Snapshot` object synchronously in a pending state. The data would still be loaded asynchronously, populated into the Snapshot and then either `load` or `error` event would be emitted.

alecgibson:
Aha I see. Yes, that would be more in keeping with `Doc` and `Query`. The main reason I didn't do that was to reinforce the idea that the snapshot in the callback is a dumb, read-only blob of data that has no functions or way of interacting with it.

gkubisa:
I would slightly prefer returning a `Snapshot`, mostly for consistency and extensibility. Getting a snapshot by time could then be exposed on `Snapshot`, rather than on the `Connection`. I'd love to hear @nateps and @ericyhwang 's opinions on this too.

lib/backend.js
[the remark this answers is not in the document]
gkubisa: Ok, I think I get it now... op version `X` in the database means that the op can be applied to a doc at version `X`. After the op is applied, the doc is at version `X + 1`. I think `Backend.prototype._getSnapshot` will need to be updated to reflect that, then we should both be happy. :-)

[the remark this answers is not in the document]
alecgibson: Riiiight. So creation is v1?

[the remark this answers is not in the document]
gkubisa: The initial `create` is applied to a doc at version 0, so after it's applied, the doc is v1.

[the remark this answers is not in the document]
alecgibson: Great. Pushed a change that matches this logic. I'm now much happier about v0 being a non-existent doc.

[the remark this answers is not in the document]
gkubisa: Nice, thanks @alecgibson :+1:

test/client/snapshot-request.js
```diff
@@ -0,0 +1,430 @@
...
+  });
+
+  afterEach(function (done) {
+    clock.uninstall();
+
+    backend.close(done);
+  });
+
+  describe('a document with some simple versions a day apart', function () {
+    var emptySnapshot = {
```
gkubisa: It would be nice to update the variable names for clarity:

- emptySnapshot -> v0
- v0 -> v1
- v1 -> v2
- v2 -> v3

lib/backend.js
```diff
@@ -568,6 +573,86 @@ Backend.prototype.getChannels = function(collection, id) {
...
+        type: request.type
+      });
+    });
+  });
+};
+
+Backend.prototype._getSnapshot = function (agent, index, id, version, timestamp, callback) {
+  version = (typeof version === 'number' && isFinite(version)) ? Math.max(0, version) : null;
+  var options = { metadata: true };
+  var timestampIsNumber = typeof timestamp === 'number';
```
dcharbonnier: are we covered when we send NaN ?

alecgibson: I think we would have been covered by the `if` block, because `1 > NaN === false`, but I've added an explicit check, too.

curran:
How hard would it be to pull this out into a separate library? That would pave the way for optimization. Related discussion: https://github.com/share/sharedb/issues/223#issuecomment-404730966

gkubisa:
IMHO getting snapshots is a core feature, so I'd prefer to keep it in ShareDB, however, I realize that some optimizations should not be in the core, eg cache. Perhaps we could support a suitable middleware, which would allow us to plug in optimizations implemented as separate modules?

gkubisa:
On the surface it would make sense to have "get snapshot at version" at the driver level, however, I don't think it would be appropriate for a few reasons.

First of all, all DB drivers would have to implement a new method with similar optimizations. It would create a maintenance overhead and discourage creation of new drivers. I'd be very reluctant to introduce any new concepts into the drivers beyond what we currently have.

Secondly, DB drivers don't have access to OT types, so even though they have all data, they can't manipulate it in any meaningful way, severely limiting optimization opportunities. For example, drivers cannot generate meta ops nor snapshot milestones. Introducing OT types to drivers is not an option for me, as it would break the separation of concerns between the code and the drivers.

Lastly, if we implement optimizations/caching as separate modules, then they could be reused with any DB drivers. I think that ideally those modules would operate as middleware actions.

alecgibson:
Okay @gkubisa all compelling arguments. I think I'm back round to thinking it _is_ a `sharedb` concern ha. Not 100% sure what optimizing middleware would look like, but let's maybe leave that to [the optimization issue][1].

[1]: https://github.com/share/sharedb/issues/223

curran:
I agree with @gkubisa 's recent arguments that this would be great to merge as-is into core ShareDB.

My thinking is that there are lots of "pros" and not many "cons":

 * It's working
 * It's a clean and well reviewed implementation
 * It would be a great selling point for ShareDB
 * A more complex optimized implementation could exist externally

alecgibson:
@nateps and anyone else - we discussed in the PR meeting adding a `Snapshot` class (like `Doc`) and returning this from `connection.getSnapshot()`.

Do you see this working like this:

```javascript
var snapshot = connection.getSnapshot("collection", "id", 123);
snapshot.fetch(callback);
```

...or like this:

```javascript
var snapshot = connection.getSnapshot("collection", "id");
snapshot.fetch(123, callback);
```

That is, does a `Snapshot` inherently have a version (or timestamp) from the outset, or is it just an empty vessel into which whatever the latest `fetch` operation dumps its contents?

My initial thinking is that the second option is possibly more consistent with `Doc`, in that `Doc` starts as an empty vessel, whose metadata (`type`, `v`, etc.) are populated after `fetch` (and subsequent `fetch` calls just overwrite the existing info...right...?).

What do you think?

gkubisa:
I would actually be in favour of the first option because a static version is what differentiates a Snapshot from a Doc.

ericyhwang:
My takeaway from the discussion was that we should keep the method signatures as you currently have them:
- `connection.getSnapshot(collection, id, version, callback): void;`
- `connection.getSnapshotAtTime(collection, id, timestamp, callback): void;`

With the one change that the callback should get called with a fully-populated instance of a to-be-written Snapshot class, instead of getting called with a generic Object. The Snapshot class can be moved out from MemoryDb.

I don't think the Snapshot class needs any new methods or anything in that case, as it's just acting as a type definition.

gkubisa:
I knew you'd bring it up. :-) What about supporting snapshot retrieval by version only and flexible ops queries? We could have the following methods:

- `connection.getSnapshot(collection, id, version): Snapshot`
- `connection.getOps(options): Ops`

With those 2 methods you could support your original use case of getting a snapshot by time by first finding an op by time and then getting a snapshot at version of the op. However, with ops queries you could provide a much better user experience, for example by displaying a list of interesting ops for users to choose from.

I'm not sure what the best format for `options` is. Maybe it could be DB driver specific? That would be consistent with `query` and provide a lot of flexibility.

gkubisa:
@ericyhwang I think it might actually make sense to have some methods etc in the `Snapshot` class and return `Snapshots` synchronously from `getSnapshot`. For example, when you ask for a snapshot, the document might not exist yet but might be created later. You could call `fetch` on the `Snapshot` to load the data once it becomes available. Perhaps you could even `subscribe` to automatically load the data when it becomes available.

It could even make sense to have `submitOp` on `Snapshot`. One use case is for adding inline comment markers in text documents. For example, a user views a past (published) version of a document and adds a comment to a particular part of the text. The same comment should appear in the latest (draft) version. There are still a few things I have to figure out here but the point I'm trying to make is that `Snapshot.submitOp` might actually have valid use cases.

gkubisa:
@ericyhwang 

> From what I understand, each snapshot is considered immutable after it's created and committed by the server, so I'm a bit wary of adding the same client statefulness that's present in Doc until a practical use-case comes up for snapshots that requires similar statefulness.

Yes, snapshots are immutable *once created*. So, if you ask for a snapshot at version > 0, the answer might be "does not exist" now and a second later it might be a valid snapshot, if it was created in the meantime. One solution is a stateful Snapshot, which gets populated once the snapshot is created. Another is to have stateless Snapshots and pass an Error to the callback, if the Snapshot does not currently exist for version > 0 - that's what we currently have and I'm happy with this solution.

> For the specific case of commenting on past revisions, I think it'd be simpler for the application to store the comments as a first-class collection, with references to the text's collection, id, version, character range. That way, the comments themselves would get the full benefit of OT, their own ops, etc.

Yes, I would certainly keep the comments separate as first class entities, and maybe get those comments reference the text's collection, id, version and character range. However, what I also would like to do is to display those comments in the latest version of the document. Here's a practical use case... I write a book. When I have the first draft ready, I send it to some people for review at version X. They add comments referencing version X and specific text ranges. As I work through the comments and make amendments, my document gets to version X + Y. I still want to see the comments in the correct places at version X + Y, even though they were added to version X. I don't want to switch between the versions to see the comments (X) and make changes (X + Y).

The solution I had in mind was: when an inline comment is added at version X, I create an operation which adds a text attribute containing a comment thread ID (see [rich-text#retain-operation](https://github.com/ottypes/rich-text#retain-operation)) and I submit it. ShareDB transforms it all the way up to the latest version and commits it. This way I can see the comment in the latest version. In order to see the comments in version X, I just need to apply the operation. An API for that could be:

```javascript
Snapshot.prototype.submitOp(op, callback): Snapshot;
Snapshot.prototype.submitSnapshot(snapshot, callback): Snapshot;
```

The new snapshots are returned synchronously because `apply` is synchronous. I still have some open questions regarding this solution, eg storing the ops/snapshots which add/contain the inline comment markers, but I think it's going in the right direction.

> Couple specific ideas on API design:

I think we could just mimic `createFetchQuery` and `createSubscribeQuery`, or maybe even extend those methods to allow querying ops. Either way, I think we could reuse most of the query-related code for querying ops.

alecgibson:
@gkubisa / @ericyhwang I've added a `Snapshot` class in line with what I've said above in [What next][1]. Please let me know what you think.

[1]: https://github.com/share/sharedb/pull/220#issuecomment-406168151

gkubisa:
@alecgibson 

> I don't think I want to have to fetch all ops to determine what version I want from a timestamp. This couldn't even be sped up with an optimisation. Practically, I'd probably instead have to fiddle around in the ops collection myself to determine a version from a timestamp, which doesn't really feel like a very friendly client interface. We could alternatively add a getVersionFromTimestamp method, but again it all feels quite manual when all I want is a snapshot from a timestamp.

With the right query you could load only a single op, eg `time<=X ORDER BY time DESC LIMIT 1`. You could do that efficiently in MongoDB too.

> As I said in the meeting, I actually personally think that timestamps are a more natural way of thinking about snapshots than in versions (from the consumer point-of-view. Obviously from a storage point-of-view, versions are a more obvious discrete unit).

That's right. Considering that ShareDB is a low level library for which "version" is a central concept and timestamps are not used for anything, I'd argue that we should not add new APIs designed specifically for timestamps.

> what do you think?

The plan looks good to me. I would also like to know @ericyhwang's and @nateps's opinions about whether or not to keep `getSnapshotByTime`.

lib/snapshot.js
```diff
@@ -0,0 +1,9 @@
+module.exports = Snapshot;
+function Snapshot(collection, id, version, type, data, meta) {
+  if (collection) this.collection = collection;
```
gkubisa: I would always assign `collection` and `m` to keep the object shape always the same, which is friendly to the JS optimizer.

alecgibson: So I tried that, and then a bunch of tests broke. I can do it, but I personally don't understand the knock-on effects of eg storing `collection` in `MemoryDB`...?

alecgibson:
> With the right query you could load only a single op

@gkubisa yes I'm aware of that, but I'm just trying to highlight that going _around_ the library to achieve (what I think) is a very normal thing to do feels off to me. It also assumes that we always use the same driver, which we may want to swap (and indeed we do for eg tests).

But yes, I'd also be interested to hear @ericyhwang and @nateps's opinions on it. I think there's a very valid case to be made for keeping and supporting `getSnapshotAtTime`.

lib/snapshot.js
gkubisa: Re storing the `collection` field, I agree that it's not necessary and would expect no impact on tests and apps. Maybe for simplicity the tests are just written to check if there are no extra fields beyond those expected (using `expect(...).eql(...)`)?

PS. Here's the first article I found about keeping object shape the same https://ariya.io/2012/02/javascript-object-structure-speed-matters. You can certainly find more detailed explanations.

alecgibson: As in you think I _should_ store `collection` in `MemoryDB`?

gkubisa: I wouldn't *mind* if it was stored in MemoryDB. It certainly *should not* be stored by *production* DB drivers, as it would be just a waste of space.

If mere presence of the `collection` property is a problem, then I would question whether we should reuse the same class for the UI and MemoryDB.

alecgibson: Yeah, I mean I'm happy to revert this change if people don't want it. I'm only doing it because people seemed keen in the PR meeting.

alecgibson:
@ericyhwang / @gkubisa / @nateps as discussed in the meeting yesterday, I have:

  - Removed the ability to fetch by time. We agreed that this would be a useful API to implement, but we need to consider the ramifications on the database adapters before exposing it. I suggest moving forward with the version-only approach, and optimising this approach, and then once we've done the whole process, then we should have a better idea of what needs to be done to fetch by time. For now it is left to the consumer to query their own database to figure out which version they need.
  - Updated the `Snapshot` class to remove the `collections` property to be consistent with the database snapshots; and always include the `m` property
  - Renamed `getSnapshot` to `fetchSnapshot` to imply that we touch the database layer

lib/client/connection.js
```diff
@@ -584,3 +600,47 @@ Connection.prototype._firstQuery = function(fn) {
...
+ *
+ */
+Connection.prototype.fetchSnapshot = function(collection, id, version, callback) {
+  if (typeof version === 'function') {
+    callback = version;
+    version = null;
+  }
+
+  var snapshotRequest = new SnapshotRequest(this, collection, id, version, callback);
+  snapshotRequest.version = version;
```
gkubisa: This line is probably redundant, considering that `version` is passed into the constructor just above.

alecgibson: Good catch.

alecgibson:
@nateps @ericyhwang I've made the changes we discussed in the meeting yesterday (see commit message for details).

adityachugh:
Great work guys, this looks extremely useful! 😊Any idea on when this may be merged?

lib/backend.js
```diff
@@ -580,6 +582,68 @@ Backend.prototype.getChannels = function(collection, id) {
...
+        data: request.snapshots[0],
+        v: request.v,
+        type: request.type
+      });
+    });
+  });
+};
+
+Backend.prototype._fetchSnapshot = function (agent, index, id, version, callback) {
+  this.getOps(agent, index, id, 0, version, function (error, ops) {
```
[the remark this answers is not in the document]
alecgibson: NB: I'm directly using `backend.db.getOps`, because `backend._getOps` would have basically had the same signature and just been a wrapper.

alecgibson:
@ericyhwang / @nateps I've addressed the issues we spoke about in yesterday's meeting. Could you please re-review?

ericyhwang:
This has been published as [sharedb@1.0.0-beta.11](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.11)

PR #222, opened 2018-07-12
Make whenNothingPending always async

gkubisa:
See https://github.com/share/sharedb/pull/204#issuecomment-404236295


lib/client/doc.js
```diff
@@ -200,7 +200,14 @@ Doc.prototype.whenNothingPending = function(callback) {
     this.once('nothing pending', callback);
     return;
   }
-  callback();
+  var doc = this;
+  process.nextTick(function() {
+    if (doc.hasPending()) {
```
curran: There is duplicated logic here, which it seems is not required.

Your current formulation is this:

```js
Doc.prototype.whenNothingPending = function(callback) {
  if (this.hasPending()) {
    this.once('nothing pending', callback);
    return;
  }
  var doc = this;
  process.nextTick(function() {
    if (doc.hasPending()) {
      doc.once('nothing pending', callback);
      return;
    }
    callback();
  });
};
```

but could it be this instead?

```js
Doc.prototype.whenNothingPending = function(callback) {
  var doc = this;
  process.nextTick(function() {
    if (doc.hasPending()) {
      doc.once('nothing pending', callback);
      return;
    }
    callback();
  });
};
```

The end result should be the same in terms of behavior.

gkubisa: Thanks @curran , I've just made this change.

gkubisa:
@ericyhwang , could this PR be merged now?

test/client/doc.js
```diff
@@ -14,28 +14,39 @@ describe('client query subscribe', function() {
...
+      if (err) return done(err);
+      var doc2 = connection.get('dogs', 'fido');
+      expect(doc).not.equal(doc2);
+      expect(doc).eql(doc2);
+      done();
+    });
   });
 
-  it('doc.destroy() calls back', function(done) {
+  it('doc.destroy() works without a callback', function() {
```
curran: Maybe keep both of theses cases?

```js
it('doc.destroy() calls back', ...);
it('doc.destroy() works without a callback', ...);
```

The case where it does invoke the callback is a good thing to have covered by tests IMO.

gkubisa: @curran the previous 2 tests use `destroy` with a callback, so this use case is covered sufficiently IMHO.

curran: Aha yes you're right. Apologies for missing that.

curran: Aha yes you're right. Apologies for missing that.

In that case I'd say this looks good to merge.

curran:
@gkubisa This change looks good, but unfortunately now has merge conflicts.

@ericyhwang I think this would be a nice improvement. In fact I just ran into issues around this async behavior today.

gkubisa:
This PR has been already open for over 9 months, so I have very little motivation to keep it up to date with the master branch. It's already merged into the Teamwork fork of ShareDB.

curran:
Understood. It's unfortunate that this good work has not gotten the attention it deserves.

ericyhwang:
Ah, I lost track of this, sorry - my email notification routing for the Share org wasn't properly set up at that point.

I'll do some integration testing on my end, and if that all checks out, we can get it merged tomorrow.

curran:
Potentially this nextTick can be removed

https://github.com/share/sharedb/blob/5394fa4494458fdff551ef71dbd3a397e66b05d8/lib/client/connection.js#L583

ericyhwang:
This is published as [sharedb@1.0.0-beta.22](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.22)

PR #224, opened 2018-07-16
Store custom client ID on Agent/Connection ID

alecgibson:
It can be useful to identify which connection submitted what operation.
This is already possible to some extent by tracking the ID randomly
generated when instantiating `Agent`.

However, this ID obviously changes on every connection, and bears no
relation to the actual client that is connecting.

This change allows the client to specify some ID when connecting, which
will be concatenated on to the agent's ID like this:

`<randomId>:<clientId>`

Keeping the `randomId` ensures that the IDs remain unique, and
concatenating the `clientId` allows the consumer to later determine
which operations were submitted by the client that matches that ID, by
checking the `src` field stored on an operation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ var socket = new WebSocket('ws://' + window.location.host);
 var connection = new sharedb.Connection(socket);
 ```
 
-The native Websocket object that you feed to ShareDB's `Connection` constructor **does not** handle reconnections. 
+The native Websocket object that you feed to ShareDB's `Connection` constructor **does not** handle reconnections.
 
 The easiest way is to give it a WebSocket object that does reconnect. There are plenty of example on the web. The most important thing is that the custom reconnecting websocket, must have the same API as the native rfc6455 version.
 
@@ -123,6 +123,12 @@ share.listen(stream);
 For transports other than WebSockets, expose a duplex
 stream that writes and reads JavaScript objects. Then
 pass that stream directly into `share.listen`.
+
+The `listen` method accepts the following arguments:
+
+  - `stream` - the stream to listen to for messages from the client
+  - `req` (optional) - an initial request which is passed through to any `connect` middleware. This is useful for inspecting cookies; Express session; etc. on the request object in the middleware
+  - `clientId` (optional) - an identifier for the connecting client. This will be prepended to the connection ID, and can be used to identify the source client of an operation. The `src` field will look like `<clientId>:<randomId>`
 
 ### Middlewares
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -11,11 +11,12 @@ var types = require('./types');
  * @param {Backend} backend
  * @param {Duplex} stream connection to a client
  */
-function Agent(backend, stream) {
+function Agent(backend, stream, clientIdSuffix) {
   this.backend = backend;
   this.stream = stream;
 
-  this.clientId = hat();
+  clientIdSuffix = typeof clientIdSuffix === 'string' ? (':' + clientIdSuffix) : '';
+  this.clientId = hat() + clientIdSuffix;
   this.connectTime = Date.now();
 
   // We need to track which documents are subscribed by the client. This is a
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -11,12 +11,12 @@ var types = require('./types');
  * @param {Backend} backend
  * @param {Duplex} stream connection to a client
  */
-function Agent(backend, stream, clientIdSuffix) {
+function Agent(backend, stream, clientIdPrefix) {
   this.backend = backend;
   this.stream = stream;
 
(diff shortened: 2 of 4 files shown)


alecgibson:
@gkubisa I've added a bit more documentation for `Backend.listen`. I've also inverted the order of the ID. The custom ID is now prepended instead of appended, so that if the underlying datastore has an index on the `src` field, you can easily group and fetch ops by a single client.

gkubisa:
Nice, thanks @alecgibson :+1: It's merged and published as `@teamwork/sharedb@v1.3.0`.

gkubisa:
It would be great, if the remote `src` was reported in `create`, `op` and `del` events... currently we always get `false`. How do you access the `clientId` on the client side, @alecgibson ?

In my opinion the above events should be emitted with an Event object, rather than individual params for `op`, `src`, etc. This would be much more flexible, however, it would require a major version bump.

alecgibson:
@gkubisa this isn't to be confused with the (other) [`source`][1] on `Doc` events. I've never been very sure of what the purpose of that one is apart from figuring out which events have been fired from which clients.

The intended use case for _this_ `src` modification is to get some client-identifying tag into the `src` field stored in the database (which is currently just a random Agent/Connection ID).

[1]: https://github.com/share/sharedb/blob/master/lib/client/doc.js#L737

gkubisa:
Grand, I might look into exposing it in the client API at some point.

alecgibson:
Yes, that would make sense. My use-case is really only about dumping them into the database for retrieval via Mongo (outside of ShareDb), but they should be pretty trivial to extract again, too at the other end. I thought I wouldn't implement that end, because I don't really know what would be required.

alecgibson:
As we discussed in the PR meeting, I'll close this in favour of just storing this on metadata in the middleware.

If anyone does want this sort of functionality in the future, we discussed that the best option is just to let consumers pass in a `function` that generates an ID, to give more control to them.

curran:
How would the equivalent middleware code look?

Might it be possible to convert this PR into a reusable middleware package?

It would be a shame to see this work go to waste, especially if it address a common problem.

alecgibson:
@curran I need to implement this, so once I've done it I'll try to share a gist or something.

alecgibson:
@curran for reference, this is roughly what the middleware looks like:

```javascript
share.use('connect', (request, callback) => {
  // Clone request to avoid mutation after connection
  const requestJson = JSON.stringify(request.req || {});
  const requestData = JSON.parse(requestJson);

  Object.assign(request.agent.custom, requestData);
  callback();
});

share.use('submit', (request, callback) => {
  request.op.m.userId = request.agent.custom.userId;
  callback();
});
```

curran:
Awesome. Thanks!

houshuang:
@alecgibson Hi, this looks like exactly what I need, but I need a bit more context to use it. My use case is that I want to identify a userid when logging in from the client, however for the life of me I cannot figure out how to pass the custom userId when connecting from a client? Your help would be much appreciated! I tried 

```js
  const ReconnectingWebSocket = require('reconnectingwebsocket');
  const sharedbClient = require('sharedb/lib/client');

  socket = new ReconnectingWebSocket(shareDbUrl);
  _connection = new sharedbClient.Connection(socket, { userId: 'stian' });
```

but when console.logging the request in the middleware, it was undefined. And anyway I don't see any code in the client library which would pass any information along to the server?

alecgibson:
@houshuang you need to pass the information in from the [server side][1], so wherever it is in your backend code that you call `share.connect` or `share.listen`, you need to pass the user Id in there (eg by retrieving it through an authentication cookie or similar).

[1]: https://github.com/share/sharedb/blob/master/lib/backend.js#L137-L168

alecgibson:
@houshuang on your server you should some code that looks something like this (from the [`README`][1]):

```javascript
var WebSocketJSONStream = require('websocket-json-stream');

// 'ws' is a websocket server connection, as passed into
// new (require('ws').Server).on('connection', ...)
var stream = new WebSocketJSONStream(ws);
share.listen(stream);
```

Each websocket will be unique to the connecting client, which you are hopefully somehow authenticating when they connect. With that authentication information, you should have a handle on their user ID on the server, so you should be able to adapt the above to something like:

```javascript
var wss = new WebSocket.Server();
wss.on('connection', function(ws, req) {
  // If you're using Express you'll need to use cookie-parser.
  const authenticationToken = req.cookies.token;
  // Then you'll presumably need to do something with your token
  // to check what the user's ID is (eg db lookup, or parse JWT, etc.)
  const userId = getUserIdFrom(token);

  const stream = new WebSocketJSONStream(ws);   
  backend.listen(stream, { userId: userId });
});
```

[1]: https://github.com/share/sharedb#listening-to-websocket-connections

Issue #202, opened 2018-04-13
modernize code


dcharbonnier:
It would be nice to modernize this code, this include :
- using es6 syntax (classes, exports, let, const)
- using promises and only promises
- using typescript or at least write a type definition
- dropping support for callback, browserchannel, etc.
- expect.js is a dead project
- add linter / formater
- callback with real Error objects, remove things like callback({ code: 4008, message: 'Unknown type' })

curran:
Seems like a heavy lift without much gain to me. Also it would break working code and require a major version release (and the project is not even at 1.0 yet).

As a user of ShareDB not using TypeScript, what do I stand to gain from the proposed changes?

dcharbonnier:
The only point that would require a change for the users of the library is 
"- dropping support for callback, browserchannel, etc."
I'm not personally favorable on hacking a code to support both callback and promises but it's something possible.
The fact that the library will never reach 1.0 is not so important IMO, better have a 2.0 with a code that look like 2018 than an already outdated 1.0.
The benefits of TS is to be able to refactor the code with less pain, to generate a proper documentation, to avoid some of the tests, to provide interfaces for the development of plugins or database drivers (Today this is a reverse engineering work which make me quite uncomfortable). 
You will find plenty of blog posts of projects moving from JS to TS and the benefits, if you find one against it please share.

curran:
Indeed, the benefit for users of ShareDB seems to be the switch from callbacks to Promises.

The other benefits are for maintainers and contributors.

dcharbonnier:
Example, it you want to create a new PubSub you just need to extend the abstract PubSub :

```typescript
export interface IPubSubOptions {
    prefix?: string;
}
export interface IStream {
    id: string;
}
export declare abstract class PubSub {
    private static shallowCopy(object);
    protected prefix?: string;
    protected nextStreamId: number;
    protected streamsCount: number;
    protected streams: {
        [channel: string]: IStream;
    };
    protected subscribed: {
        [channel: string]: boolean;
    };
    protected constructor(options?: IPubSubOptions);
    close(callback: () => void): void;
    publish(channels: string[], data: object, callback: (err: Error | null) => void): void;
    subscribe(channel: string, callback: (err: Error | null, stream?: IStream) => void): void;
    protected abstract _subscribe(channel: string, callback: (err: Error | null) => void): void;
    protected abstract _unsubscribe(channel: string, callback: (err: Error | null) => void): void;
[...]

PR #225, opened 2018-07-19
[WIP] Promisify client-facing methods

alecgibson:
This change makes client-facing methods return a `Promise` when no `callback` is provided.

## Motivation

 - As a consumer of ShareDB using modern JavaScript with `Promise`s and `async`/`await`, I'm bored of using callbacks, which are inconsistent with the rest of my codebase
 - As a contributor to ShareDB, I'm bored of tests with deeply nested callbacks (and other maintenance burdens imposed by callbacks)

## Work so far

This is a work-in-progress pull request in order to get opinions on how people feel about this as a feature; and how it should be implemented.

So far, I have added a `promisify` method to `util`, which has the following signature:

`util.promisify = function(callback): { promise, callback }`

If a `callback` is provided, then `promise` is `undefined`, and the returned `callback` is the same as the provided `callback`. This is to avoid clients inadvertently creating logic forks, where code follows both a callback and a `Promise`.

If `callback` is not a `function`, then the returned `callback` is a `function` that will `resolve` or `reject` the returned `promise`.

Also, for exiting functions early, `callback()` will return the `promise`, so you can write things like: `if (condition) return callback();`.

## Discussion

Through this work I've discovered at least two points for discussion:

 - Implementing this in `Doc` as it currently is means that `emit` is basically never called, because `callback` will now _always_ be defined. I don't know enough about how people use these events, but we could potentially have consumers configure ShareDB to enable promises? If promises are not enabled, then `util.promisify` will _always_ return the same `callback` as provided (even if it's falsy).
 - The signature for a `Query` callback is `callback(error, results, extra)`, but a `Promise` can only `resolve` a single argument. Should we combine this into `{ results, extra }` for the purposes of a `Promise`?
 - Do we bother including a polyfill for `Promise`, or just not bother promisifying if `Promise` is undefined? (And if we avoid a polyfill, how do I get the linter to calm down without targeting ES6?)

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -1,6 +1,7 @@
 var emitter = require('../emitter');
 var ShareDBError = require('../error');
 var types = require('../types');
+var util = require('../util');
 
 /**
  * A Doc is a client's view on a sharejs document.
@@ -102,6 +103,7 @@ function Doc(connection, collection, id) {
 emitter.mixin(Doc);
 
 Doc.prototype.destroy = function(callback) {
+  var { promise, callback } = util.promisify(callback);
   var doc = this;
   doc.whenNothingPending(function() {
     doc.connection._destroyDoc(doc);
@@ -110,6 +112,8 @@ Doc.prototype.destroy = function(callback) {
     }
     if (callback) callback();
   });
+
+  return promise;
 };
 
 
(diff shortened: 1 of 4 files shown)


alecgibson:
Closing this, as apparently @nateps is [not keen on the idea of `Promise`s][1] in this library.

[1]: https://github.com/share/sharedb/issues/202#issuecomment-406245765

PR #227, opened 2018-07-19
Fix flaky maxSubmitRetries test

alecgibson:
This is a slightly speculative fix for a test that fails intermittently
on `sharedb-mongo`. I believe these intermittent failures are due to a
race condition in a concurrency test.

The test works by attempting to fire two commits off at the same time,
and hoping that one of them is committed just before the other, so that
a `SubmitRequest.retry` is triggered whilst the `maxSubmitRetries` is
set to `0`, resulting in an error that is expected.

However, I believe it's possible for these commits to (in some cases)
happen sequentially rather than concurrently, and fail to error.

This change attempts to force them into this retry condition by:

  - Catching both ops in the `commit` middleware, _just_ before they're
    about to be committed (and hit a `retry` if applicable)
  - Waiting until both ops have reached this state
  - Triggering the first op's `commit`
  - Then in the callback of that op, triggering the second op's `commit`
  - The second op should now find that the first op has beaten it to
    committing, and trigger a `retry`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,6 @@ node_js:
   - "8"
   - "6"
   - "4"
-script: "npm run jshint && npm run test-cover"
+script: "ln -s .. node_modules/sharedb; npm run jshint && npm run test-cover"
 # Send coverage data to Coveralls
 after_script: "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js"
--- a/test/client/submit.js
+++ b/test/client/submit.js
@@ -509,6 +509,7 @@ describe('client submit', function() {
   });
 
   it('submits fail above the backend.maxSubmitRetries threshold', function(done) {
+    var backend = this.backend;
     this.backend.maxSubmitRetries = 0;
     var doc = this.backend.connect().get('dogs', 'fido');
     var doc2 = this.backend.connect().get('dogs', 'fido');
@@ -516,18 +517,24 @@ describe('client submit', function() {
       if (err) return done(err);
       doc2.fetch(function(err) {
         if (err) return done(err);
-        var count = 0;
-        var cb = function(err) {
-          count++;
-          if (count === 1) {
-            if (err) return done(err);
-          } else {
-            expect(err).ok();
-            done();
+        var docCallback;
+        var doc2Callback;
+        backend.use('commit', function (request, callback) {
+          if (request.op.op[0].na === 2) docCallback = callback;
+          if (request.op.op[0].na === 7) doc2Callback = callback;
+
+          if (docCallback && doc2Callback) {
+            docCallback();
           }
-        };
-        doc.submitOp({p: ['age'], na: 2}, cb);
-        doc2.submitOp({p: ['age'], na: 7}, cb);
+        });
+        doc.submitOp({p: ['age'], na: 2}, function (error) {
+          if (error) return done(error);
+          doc2Callback();
+        });
+        doc2.submitOp({p: ['age'], na: 7}, function (error) {
+          expect(error).ok();
+          done();
+        });
       });
     });
   });
--- a/test/client/submit.js
+++ b/test/client/submit.js
@@ -519,16 +519,28 @@ describe('client submit', function() {
         if (err) return done(err);
         var docCallback;
         var doc2Callback;
+        // The submit retry happens just after an op is committed. This hook into the middleware
+        // catches both ops just before they're about to be committed. This ensures that both ops
+        // are certainly working on the same snapshot (ie one op hasn't been committed before the
+        // other fetches the snapshot to apply to). By storing the callbacks, we can then
+        // manually trigger the callbacks, first calling doc, and when we know that's been committed,
+        // we then commit doc2.
         backend.use('commit', function (request, callback) {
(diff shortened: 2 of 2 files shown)


test/client/submit.js
```diff
@@ -509,25 +509,32 @@ describe('client submit', function() {
...
+          if (docCallback && doc2Callback) {
+            docCallback();
           }
-        };
-        doc.submitOp({p: ['age'], na: 2}, cb);
-        doc2.submitOp({p: ['age'], na: 7}, cb);
+        });
+        doc.submitOp({p: ['age'], na: 2}, function (error) {
+          if (error) return done(error);
+          doc2Callback();
```
ericyhwang: I think you can move this up to just below `docCallback()`, as by that point both clients will have already fetched the snapshot in `submit` and gone through `apply`.

Having the `doc2Callback()` call up there makes it a bit more obvious that the middleware is deferring the first callback until the second one is ready, and calling both in sequence.

test/client/submit.js
```diff
@@ -509,25 +509,32 @@ describe('client submit', function() {
...
-        var cb = function(err) {
-          count++;
-          if (count === 1) {
-            if (err) return done(err);
-          } else {
-            expect(err).ok();
-            done();
+        var docCallback;
+        var doc2Callback;
+        backend.use('commit', function (request, callback) {
```
ericyhwang: I and future contributors (like future you!) would appreciate it if you added some comments here about why this is needed.

ericyhwang:
Forgot to drop this note for the future - [the failing test](https://github.com/share/sharedb/blob/v1.0.0-beta.8/test/client/submit.js#L511-L533) and its output:
https://travis-ci.org/share/sharedb-mongo/jobs/405812496#L1163

```
  1) db client submit submits fail above the backend.maxSubmitRetries threshold:
     Uncaught Error: expected undefined to be truthy
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.ok (node_modules/expect.js/index.js:115:10)
      at cb (node_modules/sharedb/test/client/submit.js:525:25)
      at callEach (node_modules/sharedb/lib/client/doc.js:905:7)
      at Doc._clearInflightOp (node_modules/sharedb/lib/client/doc.js:891:16)
      at Doc._opAcknowledged (node_modules/sharedb/lib/client/doc.js:836:8)
      at Doc._handleOp (node_modules/sharedb/lib/client/doc.js:286:10)
      at Connection.handleMessage (node_modules/sharedb/lib/client/connection.js:243:20)
      at StreamSocket.socket.onmessage (node_modules/sharedb/lib/client/connection.js:125:18)
      at node_modules/sharedb/lib/stream-socket.js:59:12
      at _combinedTickCallback (internal/process/next_tick.js:131:7)
      at process._tickCallback (internal/process/next_tick.js:180:9)
```

The failure can be reproduced artificially by wrapping the `doc2.submitOp` call in a very short `setTimeout`, to simulate the doc1 commit going through very quickly or the doc2 commit's snapshot fetch being a bit slow.

test/client/submit.js
alecgibson: Hmm. But isn't the `commit` hook just before they've been committed to the db? ie there's still a potential race condition if one of them gets committed before the other. That is, we don't know which callback will finish first, and get forced into `retry`...? This way we guarantee that `doc` will _always_ be committed before `doc2`.

alecgibson: (That could be mitigated by putting back the old shared callback with a `count`, but I find that sort of thing a bit sketchy; it's like admitting that you're not really in control of your own test.)

test/client/submit.js
alecgibson: Done

test/client/submit.js
ericyhwang: > That is, we don't know which callback will finish first, and get forced into retry...? This way we guarantee that doc will always be committed before doc2.

> (That could be mitigated by putting back the old shared callback with a count, but I find that sort of thing a bit sketchy; it's like admitting that you're not really in control of your own test.)

Got it, you want to sequence the actual commits so you know 1 goes before 2. Makes sense, thanks for explaining!

.travis.yml
```diff
@@ -5,6 +5,6 @@ node_js:
   - "8"
   - "6"
   - "4"
-script: "npm run jshint && npm run test-cover"
+script: "ln -s .. node_modules/sharedb; npm run jshint && npm run test-cover"
```
ericyhwang: Dropping another note for the future, this is to work around the circular dependency issue between sharedb and sharedb-mingo-memory:
https://github.com/share/sharedb/pull/226#issuecomment-406297957

I'm OK with this workaround for now to unblock the tests as the workaround is isolated to Travis, until we figure out a better solution in the linked issue above.

ericyhwang: Actually, maybe it's better to get the #226 merged first, as it also updates the Travis config's Node versions to reflect latest + LTE versions.

alecgibson: Sure. Happy to wait and rebase. Would just be nice to get this through to unblock other PRs.

nateps:
Thanks for the contribution! Fixing flakey tests always brings me great joy. 💥

PR #228, opened 2018-07-20
Warn about using deprecated features

gkubisa:
I've also removed `doc` from README.md because the other deprecated action, "after submit" was not there either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,8 @@
 language: node_js
 node_js:
   - "10"
-  - "9"
   - "8"
   - "6"
-  - "4"
-script: "npm run jshint && npm run test-cover"
+script: "ln -s .. node_modules/sharedb; npm run jshint && npm run test-cover"
 # Send coverage data to Coveralls
 after_script: "cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js"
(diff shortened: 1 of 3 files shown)


nateps:
Thanks, @gkubisa! Great idea to notify users about feature deprecation. I think we should start following this example with other breaking changes in ShareDB. 👍

stanographer:
I'm following this thread but I'm still not exactly sure how to get rid of these warnings. As far as I can tell these warnings are originating from the actual source code itself, and not the way I'm implementing ShareDB? Could someone please enlighten me? Thanks!!

alecgibson:
@stanographer as the warning says, pass the `disableSpaceDelimitedActions: true` when constructing ShareDB to disable both the warning, and the middleware that uses spaced names.

curran:
@stanographer It was not obvious to me either, exatly where/how these options should be pased in.

It turns out they need to be passed as properties of the object passed as the first argument to the `ShareDB` constructor, like this:

```js
  const backend = new ShareDB({
    // The following options serve only to squelch errors.
    // It should be possible to remove these when ShareDB 1.0 is released.
    disableDocAction: true,
    disableSpaceDelimitedActions: true
  });
  const connection = backend.connect();
```

PR #230, opened 2018-07-25
Remove sharedb-mingo-memory circular dependency

ericyhwang:
sharedb had depended on sharedb-mingo-memory for 4 tests, mostly relating to query subscriptions. This removes that dependency by faking out enough of MemoryDB in the tests to handle simple queries. (MemoryDB itself returns all documents in a collection given a query.)

I didn't want to put the simple query handling in MemoryDB, as that might lead people to think it may do more than it can. We don't want to re-implement all of Mingo, either.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ node_js:
   - "10"
   - "8"
   - "6"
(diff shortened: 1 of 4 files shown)


curran:
Woohoo! This is majorly awesome!

PR #236, opened 2018-08-14
Add Milestone Snapshots

alecgibson:
This non-breaking change introduces the concept of "Milestone
Snapshots". A milestone snapshot is a snapshot of a document for a given
version, which is persisted in the database. The purpose of this is to
speed up the `Backend.fetchSnapshot` method, which currently has to
fetch all the ops required to build a snapshot from v0. Instead,
`fetchSnapshot` can now fetch the most recent, relevant milestone
snapshot and build on top of that with fewer ops.

In order to do this, the database adapter API has been updated to
include two new methods:

- `saveMilestoneSnapshot(collection, snapshot, callback): void;` stores
  the provided snapshot against the collection
- `getMilestoneSnapshot(collection, id, version, callback): void`
  fetches the most recent snapshot whose version is equal to or less
  than the provided `version` (or the most recent version if version is
  `null`, in keeping with the `to` argument in `getOps`).

The adapter also has the responsibility of saving the appropriate
milestone snapshots when a new op is committed.


alecgibson:
# Discussion

## API

I've added these new methods to the driver API. They're non-breaking changes. Everything will continue to work even if a driver doesn't implement the new methods. Milestone snapshots are also off by default.

In theory, we don't need to expose `saveMilestoneSnapshot`, but I think this will be useful for back-filling milestone snapshots (see below).

## Backfilling

In its current state, this change will add milestone snapshot functionality, but only for new snapshots. That is, let's assume I already have a database with a document with 100 versions. If I now turn on milestone snapshots with an interval of 10, I will only generate milestones at v110, v120, v130, etc., and still won't have them for v10, v20, v30 ... v100.

This isn't ideal, but it's also no worse than the current performance. It may also be excusable in the (common?) cases where users will want to access the most recent snapshots anyway.

Note that this case also applies if we ever want to change the interval (ie if I have milestones every 10 versions, but I'd like them every 5 instead, then I'd be missing v5, v15, etc. for old documents).

These are the approaches I can think of for this:

  - Ignore it (for this pull request, at least). This is still a performance improvement, but we could make it better in future.
  - Leave backfilling as an exercise to the reader. By exposing the `saveMilestoneSnapshot` method on the driver, we enable the consumer to regenerate milestones however they see fit, which is probably the most flexible approach, but also the ickiest for the consumer.
  - Expose a method on ShareDB core's `Connection` class that will backfill milestones for a given document. This is friendly to the user, allowing them to pick when they want to do this (performance-heavy) task, but not having them worry about the details.
  - Automagically backfill on `Backend.fetchSnapshot`. When we perform `Backend.fetchSnapshot`, we already have access to all the ingredients we need to create arbitrary milestones. As the document is built up from ops, we could store these milestones. The performance of the first fetch would therefore be a bit worse than current, but after that it would be much faster. It would also handle changing intervals quite nicely. This is maybe the most fool-proof approach from the consumer's point of view?

## Default interval

At the moment, I've set the default milestone interval to 1,000. This is mostly based on a finger-in-the-wind gut feel (10 is definitely too frequent, 10,000 definitely too infrequent. We probably want something between 100-1,000, and 1,000 is a bit more conservative with regards to disk space).

Obviously what a consumer will want will vary depending on server performance stats, typical load, how they want to optimise disk vs speed, etc.

Do people agree with 1,000 as a "sensible" default? Or would people even prefer no default at all, and force consumers to pick a value? (I personally think everything should have sensible defaults out of the box, but I know opinion can be divided on this).

gkubisa:
Looks great to me @alecgibson :+1:

Re saving milestone snapshots on commit, I think they should be saved asynchronously after commit, and any errors when saving the milestone snapshots should not cause the commit to fail. This way the commit latency and the current implementation would not be affected - after all, milestone snapshots are just an optimization. 

Re backfilling, my preference would be to "Automagically backfill on `Backend.fetchSnapshot`" because, as you mentioned, it's user friendly and we would already have all the necessary data, so we'd just need to save the missing snapshots - we could do it asynchronously to avoid slowing down the fetchSnapshot call which backfills the snapshots. I also think we'll need automatic backfilling anyway, in case saving the milestone fails on commit.

Re default interval, I think we should *not* save any milestone snapshots by default. This way we would not accidentally use excessive storage for users who do use `Connection.fetchSnapshot`.

curran:
Great work!

alecgibson:
@ericyhwang / @nateps as discussed, I have:

  - moved the milestone logic into its own new `MilestoneDB` class
  - implemented a very simple `MemoryMilestoneDB` instance, which is provided to `Backend` by default (but doesn't save anything by default)
  - made the `MilestoneDB` an emitter, so that we can save with an optional, non-blocking callback
  - made the milestone save non-blocking, and moved the logic for saving into `SubmitRequest`
  - added a `SubmitRequest.saveMilestoneSnapshot` flag for overriding interval logic in the middleware (with an example in a test case)

@gkubisa agreed with non-blocking. That should have been addressed in these recent changes. However, regarding backfilling, we decided to leave this as an exercise to the reader, because:

  - in production you probably don't want to perform this work on a fetch (you'd probably want to run a background task)
  - you may want to add milestone generation to a queue, rather than processing (potentially unlimited) new milestones in parallel
  - it's impossible to anticipate the needs and context in which people want to backfill

lib/milestone-db/memory.js
```diff
@@ -0,0 +1,61 @@
...
+  // Map form collection name -> doc id -> array of milestone snapshots
+  this._milestoneSnapshots = {};
+}
+
+MemoryMilestoneDB.prototype = Object.create(MilestoneDB.prototype);
+
+MemoryMilestoneDB.prototype.getMilestoneSnapshot = function (collection, id, version, callback) {
+  var milestoneSnapshots = this._getMilestoneSnapshotsSync(collection, id);
+
+  let milestoneSnapshot;
```
ericyhwang: `let` -> `var`, as ShareDB code doesn't use ES6 syntax yet - I don't think there's a specifically declared policy on browser compatibility.

Server-only code could certainly start using ES6 features, as the oldest supported version of Node is 6 LTS. It'd be nice to use Map for higher performance than object properties. However, it isn't always obvious what's server-only vs client-only vs shared.

lib/milestone-db/index.js
```diff
@@ -0,0 +1,39 @@
...
+function MilestoneDB(options) {
+  emitter.EventEmitter.call(this);
+
+  // The interval at which milestone snapshots should be saved
+  this.interval = options && options.interval;
+}
+emitter.mixin(MilestoneDB);
+
+MilestoneDB.prototype.close = function(callback) {
+  if (callback) callback();
```
ericyhwang: The base and memory DB implementations aren't super-consistent about ensuring their callbacks are async, but for these two new milestone classes, I think it'd be good to consistently `process.nextTick` before calling callbacks.

lib/backend.js
```diff
@@ -24,6 +25,7 @@ function Backend(options) {
   this.pubsub = options.pubsub || new MemoryPubSub();
   // This contains any extra databases that can be queried
   this.extraDbs = options.extraDbs || {};
+  this.milestoneDb = options.milestoneDb || new MemoryMilestoneDB();
```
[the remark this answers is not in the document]
alecgibson: Yeah I was in two minds about this. I used `MemoryMilestoneDB` for consistency with `MemoryDB` and `MemoryPubSub`, but actually `MemoryMilestoneDB` just acts like the dumb `MilestoneDB` when initialised like this, so it's probably better to just use the base implementation. If you wanted to set an interval, you'd need to re-initialise it and explicitly pass it in anyway.

I'll switch for the base implementation (which was written to be completely harmless anyway), and then it's super clear that nothing's happening by default.

lib/milestone-db/memory.js
alecgibson: Ha whoops, yeah that was just a force of habit from my day-to-day work. I thought this would get picked up by the linter? (It certainly had a go at me for using the spread operator)

lib/milestone-db/index.js
alecgibson: Fine by me. I was unsure why I'd seen so many `nextTick`s around (I work in TypeScript and avoid callbacks wherever possible!). I had a read of [this blog post][1]. Is that pretty much the gist of what we're trying to achieve?

[1]: https://jsblog.insiderattack.net/event-loop-best-practices-nodejs-event-loop-part-5-e29b2b50bfe2

alecgibson:
@ericyhwang I've made the changes you requested. Please have another look and let me know if there's anything else.

alecgibson:
NB: I've thrown together an initial implementation of this for Mongo here: https://github.com/alecgibson/sharedb-milestone-mongo

lib/milestone-db/index.js
ericyhwang: > I had a read of this blog post. Is that pretty much the gist of what we're trying to achieve?

Yup, that's exactly it!

ericyhwang:
> Also, I've started work on a Mongo implementation of this. Should it live in the share organisation? What's the best way to go about that?

We probably do want it in the share org eventually. I'm not sure if there's a way open-source projects typically handle this? @nateps - you have any thoughts?

* One option is to do a [repository transfer](https://help.github.com/articles/about-repository-transfers/) after review, which will preserve all your existing commits.
* The other option is for us to set up a new empty repo in the share org, and then you make a PR against that.

I don't know enough about transfers to have a strong opinion one way or another, and either way you'll get commit authorship credit. With the second option, we'll probably want to add you as a writer to the new repo - as the original author, you'd know the code the best.

> It's (controversially?) written in ES7, but I guess we can discuss that when we get to it.

Yeah, worth a discussion. There's an argument for keeping syntax consistency within an org's projects so contributors don't have to jump back and forth between coding styles, and if we want to keep supporting, say, IE11, we can't use async functions in any client or shared code.

nateps:
I think transferring to the Share org sounds great. I added you as a contributor, so you should be able to make repos. We should keep doing code reviews, and I think we're in a good cadence on that now.

As far as ES7 goes, I'd like to avoid it for anything that has browser code and stick to good old ES5. I'd prefer to avoid the need for adding additional compilation complexity. These libraries are low level enough that they should be pretty easy to pull into any other framework or project, so I'd like to keep to very tried and true JavaScript.

If it runs on the server, it should work in all versions of Node.js that are still in LTS, so 6+ at this point. We should definitely be consistent on code style within a repo, but I'm OK with having new repos be in more modern code style if it is the direction that we'd like to move in with all repos.

If we were going to introduce a compiler requirement, we should discuss which direction we'd like to go: Regular ES7, TypeScript, or something else. But let's start that discussion on a more appropriate thread instead of this PR.

alecgibson:
@ericyhwang / @nateps I've opened a pull request for the Mongo Implementation here: https://github.com/share/sharedb-milestone-mongo/pull/1

It's been downgraded to vanilla ES6 to support Node LTS (down to v6). We can continue discussion there.

PR #237, opened 2018-08-20
Expose `fetchSnapshot` requested version in `readSnapshots` middleware

alecgibson:
In the `readSnapshots` middleware, it's currently impossible to
determine whether a snapshot is being returned as the result of a
`fetch` or a `fetchSnaphot`.

This change adds a version `v` field to the `request` when the
middleware is invoked from `fetchSnapshot`. This field is populated with
the value of the requested version, and can be used to determine that
the snapshot is the result of a `fetchSnapshot` rather than a `fetch`.

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -77,6 +77,13 @@ Backend.prototype.MIDDLEWARE_ACTIONS = {
   submit: 'submit'
 };
 
+Backend.prototype.SNAPSHOT_TYPES = {
+  // The current snapshot is being fetched (eg through backend.fetch)
+  current: 'current',
+  // An historical snapshot is being fetched (eg through backend.fetchSnapshot)
+  historical: 'historical'
+};
+
 Backend.prototype._shimDocAction = function() {
   if (warnDeprecatedDoc) {
     warnDeprecatedDoc = false;
@@ -264,15 +271,7 @@ Backend.prototype._sanitizeOpsBulk = function(agent, projection, collection, ops
   }, callback);
 };
 
-Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, snapshots, request, callback) {
-  if (typeof request === 'function') {
-    callback = request;
-    request = {};
-  }
-
-  request.collection = collection;
-  request.snapshots = snapshots;
-
+Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, snapshots, snapshotType, callback) {
   if (projection) {
     try {
       projections.projectSnapshots(projection.fields, snapshots);
@@ -280,6 +279,12 @@ Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, s
       return callback(err);
     }
   }
+
+  var request = {
+    collection: collection,
+    snapshots: snapshots,
+    snapshotType: snapshotType
+  };
 
   this.trigger(this.MIDDLEWARE_ACTIONS.readSnapshots, agent, request, callback);
 };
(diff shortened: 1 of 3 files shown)


alecgibson:
Our use-case in particular is maybe a bit weird, but I think it's valid. We run a lazy migration system on our documents. So let's say I have a basic document:

```json
{
  "schemaVersion": 1,
  "title": "foo",
  "v": 1
}
```

Let's now say I want to change our JSON schema. I want to add an `info` field. The way our system works, is that we have a special migrator, that stores the ops we need to apply to a document to migrate it up to the new schema, eg:

```javascript
const v2Migration = {
  schemaVersion: 2,
  ops: [
    { p: ['schemaVersion'], od: 1, oi: 2 },
    { p: ['info'], oi: '' },
  ],
};
```

When we fetch a document, we check its `schemaVersion` against all available migrations, and if it needs migrating, we apply the pending migration ops to it, so my document now looks like:

```json
{
  "schemaVersion": 2,
  "title": "foo",
  "info": "",
  "v": 2
}
```

Cool. So why do we need to know if we're fetching a snapshot rather than a doc? The nice thing about this lazy migration system is that we can fetch old snapshots (ie ones on an old `schemaVersion`) and migrate them in exactly the same way by just using `json0.apply` on the `snapshot.data`:

```javascript
const version = 1;
share.fetchSnapshot(collection, id, version, (error, snapshot) => {
  snapshot = json0.apply(snapshot, v2Migration.ops);
  callback(snapshot);
});
```

However, in order to determine which sort of migration we need to apply (that is submit ops to the database, or just apply them directly to the snapshot), we need to know in the `readSnapshots` middleware whether we're dealing with the current real-life snapshot, or an old snapshot.

Does that all make some sort of sense? I do think that the two fetch operations are conceptually different, so there should be some way of distinguishing them.

ericyhwang:
Got it, thanks for writing up all the context!

Would the middleware you're writing need the exact snapshot version on the request object, or does it just need to know "current" vs "historical"? I like for the purpose of things to be clear, so if the latter suffices, one approach I might take would be:
```
let request = {
  snapshotReadType: 'current',  // or 'historical'
  ...
};
```

That way, the conditionals in middleware would look like `if (request.snapshotReadType === 'historical')`, which is clearer at a glance than `if (request.v)`.

Exact naming of the `snapshotReadType` property and the values is TBD, those were just the most appropriate names I thought of up front.

alecgibson:
Yeah we just need to know what type of snapshot it is, and to be fair if you really cared about the version then you should be able to get that from the snapshot itself. I'll update to add that flag. I think it should be provided for all `readSnapshots`, so I'll update the `_sanitizeSnapshots` method to accept a non-optional `snapshotType` argument rather than a generic, optional `request` argument.

lib/backend.js
```diff
@@ -77,6 +77,13 @@ Backend.prototype.MIDDLEWARE_ACTIONS = {
   submit: 'submit'
 };
 
+Backend.prototype.SNAPSHOT_TYPES = {
+  // The current snapshot is being fetched (eg through backend.fetch)
+  current: 'current',
+  // An historical snapshot is being fetched (eg through backend.fetchSnapshot)
+  historical: 'historical'
```
ericyhwang: Nate: Historical could be current version, let's change to `byVersion`

alecgibson:
@ericyhwang as discussed, I've updated this to `byVersion` instead of `historical`

ericyhwang:
Published as [sharedb@1.0.0-beta.13](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.13)

I also fixed up the PR title and description to reflect the move to `snapshotType` on the request instead of `v`.

PR #244, opened 2018-08-29
Add a `NoOpMilestoneDB` implementation

alecgibson:
This change updates our `MilestoneDB` base class to be an empty shell
that throws whenever any of its methods are called in order to help
implementers.

We then also introduce a `NoOpMilestoneDB` implementation, which
implements all the methods, but does nothing. This is so that we can
drop this implementation into the `Backend` as a sensible default that
does nothing at all when called (instead of eg using `MemoryMilestoneDB`
and having the surprising behaviour of working until you restart your
application).

As part of this, the base call signature of the `saveMilestoneSnapshot`
callback is changed to remove the `wasSaved` flag. A snapshot should be
assumed to have saved if no error is returned in that callback.

### DIFF
--- a/lib/milestone-db/index.js
+++ b/lib/milestone-db/index.js
@@ -1,5 +1,6 @@
 var emitter = require('../emitter');
 var ShareDBError = require('../error');
+var util = require('../util');
 
 module.exports = MilestoneDB;
 function MilestoneDB(options) {
@@ -39,4 +40,9 @@ MilestoneDB.prototype.saveMilestoneSnapshot = function (collection, snapshot, ca
   var error = new ShareDBError(5020, 'saveMilestoneSnapshot MilestoneDB method unimplemented');
   if (callback) return process.nextTick(callback, error);
   this.emit('error', error);
+};
+
+MilestoneDB.prototype._isValidVersion = function (version) {
+  if (version == null) return true;
+  return util.isInteger(version) && version >= 0;
 };
--- a/lib/milestone-db/index.js
+++ b/lib/milestone-db/index.js
@@ -43,6 +43,5 @@ MilestoneDB.prototype.saveMilestoneSnapshot = function (collection, snapshot, ca
 };
 
 MilestoneDB.prototype._isValidVersion = function (version) {
-  if (version == null) return true;
-  return util.isInteger(version) && version >= 0;
+  return util.isValidVersion(version);
 };
--- a/lib/milestone-db/index.js
+++ b/lib/milestone-db/index.js
@@ -1,4 +1,5 @@
 var emitter = require('../emitter');
+var ShareDBError = require('../error');
 
 module.exports = MilestoneDB;
 function MilestoneDB(options) {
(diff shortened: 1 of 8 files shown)


lib/milestone-db/memory.js
```diff
@@ -22,12 +23,17 @@ function MemoryMilestoneDB(options) {
 MemoryMilestoneDB.prototype = Object.create(MilestoneDB.prototype);
 
 MemoryMilestoneDB.prototype.getMilestoneSnapshot = function (collection, id, version, callback) {
+  if (!callback) callback = function () {};
+  if (!collection) return process.nextTick(callback, new ShareDBError(4001, 'Invalid collection'));
```
ericyhwang: I'd go with "Missing collection" and "Missing ID" here to be more precise about the error, and same in `saveMilestoneSnapshot`.

lib/milestone-db/memory.js
```diff
@@ -22,12 +23,17 @@ function MemoryMilestoneDB(options) {
...
+  if (!id) return process.nextTick(callback, new ShareDBError(4001, 'Invalid ID'));
+  if (!this._isValidVersion(version)) return process.nextTick(callback, new ShareDBError(4001, 'Invalid version'));
+
   var milestoneSnapshots = this._getMilestoneSnapshotsSync(collection, id);
 
   var milestoneSnapshot;
   for (var i = 0; i < milestoneSnapshots.length; i++) {
     var nextMilestoneSnapshot = milestoneSnapshots[i];
-    if (nextMilestoneSnapshot.v <= version || version === null) {
+    if (nextMilestoneSnapshot.v <= version || version == null) {
```
ericyhwang: Why change to `==` here?

lib/milestone-db/no-op.js
```diff
@@ -0,0 +1,23 @@
...
+ */
+module.exports = NoOpMilestoneDB;
+function NoOpMilestoneDB(options) {
+  MilestoneDB.call(this, options);
+}
+
+NoOpMilestoneDB.prototype = Object.create(MilestoneDB.prototype);
+
+NoOpMilestoneDB.prototype.getMilestoneSnapshot = function (collection, id, version, callback) {
+  process.nextTick(callback, null, undefined);
```
ericyhwang: I've noticed some other places in the code defining variables purely for argument clarity, might be worth doing here too - it took me a second to see why this was passing in `null` and then `undefined` as args.
```
var snapshot = undefined;
process.nextTick(callback, null, snapshot);
```

lib/milestone-db/index.js
```diff
@@ -23,17 +25,24 @@ MilestoneDB.prototype.close = function(callback) {
...
 MilestoneDB.prototype.saveMilestoneSnapshot = function (collection, snapshot, callback) {
-  var saved = false;
-  if (callback) return process.nextTick(callback, null, saved);
-  this.emit('save', saved, collection, snapshot);
+  var error = new ShareDBError(5020, 'saveMilestoneSnapshot MilestoneDB method unimplemented');
+  if (callback) return process.nextTick(callback, error);
+  this.emit('error', error);
+};
+
+MilestoneDB.prototype._isValidVersion = function (version) {
```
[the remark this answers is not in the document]
alecgibson: Good call. I'll commonise. There's no good reason why `==` vs `===`. Just another reason why a common method would be nice. I'll make it reject `undefined` as a version, following the maxim that `null` signifies "I meant to make this `null`".

lib/milestone-db/memory.js
alecgibson: Fine by me. These errors are sort of "dummy" errors anyway, given that nobody should really be using this implementation (hence just using the `4001` code, and not wanting to invent new error codes)

lib/milestone-db/memory.js
alecgibson: This was because of the fact that `undefined` was treated as a valid version, but now it's not, so I'll change this back.

lib/milestone-db/index.js
```diff
@@ -23,17 +25,23 @@ MilestoneDB.prototype.close = function(callback) {
...
```
ericyhwang: Optional request:

Now that this method is just a wrapper, does it still need to exist (like for testing purposes or subclass overrides), or can it be removed?

alecgibson:
@ericyhwang I've removed the `_isValidVersion` wrapper method.

alecgibson:
Actually @ericyhwang I've just seen that we don't actually export `util`, so I think I want to leave `MilestoneDB._isValidVersion` in place, so that subclasses have access to it without having to roll their own. There's no reason why what constitutes a valid version should differ from implementation to implemenation. Make sense?

ericyhwang:
Published as [sharedb@1.0.0-beta.13](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.13)

PR #245, opened 2018-08-30
Improve test coverage for `SnapshotRequest.onConnectionStateChanged`

alecgibson:
This change improves `SnapshotRequest` test coverage to 100%.

Our current test for checking that a dropped connection works didn't
actually fail properly when removing `this.sent = false` from
`SnapshotRequest._onConnectionStateChanged`, so an improved test has
been written that use middleware to wait until the request has
definitely been sent, before dropping the connection and reconnecting.

It also adds a previously uncovered test case for checking that the
connection cannot send the same request multiple times, by using a
similar setup to the rewritten dropped connection case.

This change also adds some documentation recording why we might want to
reset `this.sent = false`, because just looking at the code it's
unclear why we would ever want to do that.

### DIFF
--- a/test/client/snapshot-request.js
+++ b/test/client/snapshot-request.js
@@ -214,6 +214,15 @@ describe('SnapshotRequest', function () {
     it('can drop its connection and reconnect, and the callback is just called once', function (done) {
       var connection = backend.connect();
 
+      // Here we hook into middleware to make sure that we get the following flow:
+      // - Connection established
+      // - Connection attempts to fetch a snapshot
+      // - Snapshot is about to be returned
+      // - Connection is dropped before the snapshot is returned
+      // - Connection is re-established
+      // - Connection re-requests the snapshot
+      // - This time the fetch operation is allowed to complete (because of the connectionInterrupted flag)
+      // - The done callback is called just once (if it's called twice, then mocha will complain)
       var connectionInterrupted = false;
       backend.use(backend.MIDDLEWARE_ACTIONS.readSnapshots, function (request, callback) {
         if (!connectionInterrupted) {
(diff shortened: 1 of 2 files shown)


alecgibson:
Don't know why it claims that coverage has actually _fallen_ 🤷‍♂️ 

screen shot 2018-08-30 at 10 43 03

ericyhwang:
Re the coverage drop, Coveralls thinks this PR has one fewer "relevant line" of code compared to the code on master:

* Most recent build on master: 2276 of 2383 relevant lines covered (95.51%)
* This PR: 2275 of 2382 relevant lines covered (95.51%)

Probably due to this on line 61:
```
- } else if (!this.connection.canSend) {
+ } else {
```

That's fine. Nothing we can do unless we add coverage for unrelated code.

test/client/snapshot-request.js
```diff
@@ -214,13 +214,34 @@ describe('SnapshotRequest', function () {
     it('can drop its connection and reconnect, and the callback is just called once', function (done) {
       var connection = backend.connect();
 
-      connection.fetchSnapshot('books', 'don-quixote', function (error) {
-        if (error) return done(error);
-        done();
+      var connectionInterrupted = false;
+      backend.use(backend.MIDDLEWARE_ACTIONS.readSnapshots, function (request, callback) {
```
ericyhwang: Can you add a comment about why this needs to use the middleware?

alecgibson:
@ericyhwang I've added the comment you asked for.

ericyhwang:
Thanks! I'll merge but hold off on publishing until we've got more changes, as this is just changes to the tests (minus the very minor refactor)

PR #247, opened 2018-09-13
Adds two new doc events: 'before op batch'  and 'after op batch'

danburzo:
We have recently upgraded from ShareJS to ShareDB following the [migration guide](https://github.com/share/sharedb/blob/master/docs/upgrading-from-sharejs.md#doc-api). We've replaced `after op` with the `op` event, but we have found that `op` is called for each individual operation, as opposed to `after op` which had been called after an entire batch of operations.

The addition of two new events that get called at the start, and the end, of a batch of operations allows us perform the optimizations we were doing previously for when the client receives a batch of remote operations.

This pull request introduces these events:

* `before op batch` is called before any `before op` event in a batch
* `after op batch` is called after any `op` event in a batch

They currently receive no parameters, but we can change that if any parameters would be appropriate here.

Best regards!

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,11 +279,17 @@ same time as callbacks to `fetch` and `subscribe`.
 `doc.on('create', function(source) {...})`
 The document was created. Technically, this means it has a type. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
 
+`doc.on('before op batch'), function() {...})`
+An operation batch is about to be applied to the data. For each partial operation a pair of `before op` and `op` events will be emitted after this event.
+
 `doc.on('before op'), function(op, source) {...})`
 An operation is about to be applied to the data. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
 
 `doc.on('op', function(op, source) {...})`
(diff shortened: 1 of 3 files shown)


danburzo:
We ended up publishing our fork in the interim.

adelriosantiago:
I was looking exactly for this. the `op` event triggers several times individually. And each change is actually broadcast'ed to all users. Will this PR land master?

alecgibson:
I'm not entirely sure I understand the motivation behind this change. What's the use case?

danburzo:
When receiving a batch of events, it can be beneficial to be able to tell when the batch starts and ends, so that you may generate pause side-effects at the beginning of the batch, and only update and resume when you've finished incorporating all the (potentially hundreds of) individual `op`s.

In a previous version of sharedb, `after op` would trigger after the entire operation batch is finished, and following the suggestion from the upgrade guide to switch from `after op` to `op` incurred a significant performance penalty.

alecgibson:
@danburzo sounds fairly reasonable to me. I'll see if I can discuss with other contributors tomorrow. My one reservation about it is that it would be an event that only ever has relevance to the default type, but `json0` is something of a special case across all of `sharedb`, so I've certainly heard weirder things.

Issue #250, opened 2018-10-10
Abstract console logging


alecgibson:
At the moment there's a bunch of calls to `console.warn` and the like. It would be nice to abstract this a bit so that users can provide a `console`-like object when instantiating ShareDB so that we can:

  - silence the logger (useful for tests, eg when you're deliberately throwing errors)
  - forward logs on to a custom logger, or alerting service (eg Sentry, Rollbar)

I'd imagine it to look something like:

```javascript
const ShareDB = require('sharedb');

const myLogger = {
  log: () => {}, // I don't care about logs
  warn: message => Rollbar.warn(message),
  error: message => Rollbar.error(message),
};

const share = new ShareDB({
  logger: myLogger,
});
```

If a logger isn't provided, then we can just use the default of `console`. Maybe a nice bonus option would be if we pass `logger: null`, then ShareDB will silence all logs.

We could internally also call something like:

```javascript
const logger = Object.assign({}, console, myLogger);
```

This would make sure that the logger has sensible defaults of logging to console, so you'd just have to override the things you care about. For example:

```javascript
const share = new ShareDB({
  logger: {
    log: () => {},
  },
});
```

The above would suppress all calls to `console.log`, but would still log `warn`, `error`, etc. to the console.

What do people think?

ericyhwang:
Seems reasonable. Some logging options that seem to be common in Node:

* Nate brings up the [debug](https://www.npmjs.com/package/debug) library, used in big relatively stable libraries like Express and could be switched to with a find/replace:
  * Supports namespaces for configuring log levels on a per-library basis: https://www.npmjs.com/package/debug#usage
  * Output streams can be configured on a per-namespace + per-level, or globally: https://www.npmjs.com/package/debug#output-streams
  * Logging for a namespace can be enabled/disabled dynamically in the application entry point via a call to the library, which should also work in the browser: https://www.npmjs.com/package/debug#set-dynamically
* There's also going to structured logging, though we don't need to do so now, as that'd be more work than just switching text loggers. A few options are [Bunyan](https://github.com/trentm/node-bunyan), which I've used; or [Winston](https://github.com/winstonjs/winston) and [Roarr])(https://github.com/gajus/roarr), which I haven't.

alecgibson:
[...]

Issue #144, opened 2017-04-04
1.0 Release


curran:
The project has been in 1.0 beta for some time now.

As I understand it, the current version is actually production-ready.

What's holding back a 1.0 release?

curran:
Hi @nateps ,

It seems ShareDB is "baked", as in production-ready and hardened for use.

It may be worth showing some signs of life on the project and releasing the current codebase as 1.0.

Any blockers come to mind?

Thank you.

curran:
Greetings @nateps @rkstedman ,

Just wondering if there's anything on the 1.0 roadmap I can help out with?

rkstedman:
Hi @curran! Thanks so much for checking in! You might have a noticed a bit of activity on the repo lately -- Nate, Eric, and I have started meeting 1 hour a month to look at PRs and issues and see what progress we can make. I think we're meeting this Friday and I can surface this question to the group then!

curran:
@rkstedman Awesome! Happy to hear of the meetings and steady dedication to the project :)

curran:
Noticed there is now a milestone for 1.0! This is great https://github.com/share/sharedb/milestone/1

Closing this issue, as the 1.0 goals are there in the milestone.

curran:
I'm reopening this as ShareDB has been at 1.0 Beta for over 2 years now. Just release 1.0 already please!

curran:
FWIW when you run `npm install sharedb`, it installs `v0.11.37` from 2016.

felixk101:
https://github.com/share/sharedb/releases/tag/v1.0.0 has been released since 2019, so this issue can be closed.

PR #252, opened 2018-10-31
Use loglevel instead of console

alecgibson:
Fixes https://github.com/share/sharedb/issues/250

The [`loglevel`][1] package is a drop-in replacement for `console`,
which lets us set the level of logging we wish to see.

This is particularly useful if we want to silence noisy logs (eg in
tests where we deliberately cause failures).

This change notably removes the `debug` flag from `Connection`.
Logging should be controlled through the provided interface (see below).

## Backend

The log level is controlled on the backend through the `Backend`
constructor. For example, to enable only errors:

```js
const Backend = require('sharedb');

const share = new Backend({
  logLevel: 'error',
});
```

## Client

The client-side log level can be set through a static method exposed on
the `Client`:

```js
const Client = require('sharedb/lib/client');

Client.setLogLevel('error');
```

[1]: https://github.com/pimterry/loglevel


lib/client/connection.js
```diff
@@ -116,11 +115,11 @@ Connection.prototype.bindToSocket = function(socket) {
       var data = (typeof event.data === 'string') ?
         JSON.parse(event.data) : event.data;
     } catch (err) {
-      console.warn('Failed to parse message', event);
+      logger.warn('Failed to parse message', event);
       return;
     }
 
-    if (connection.debug) console.log('RECV', JSON.stringify(data));
+    logger.debug('RECV', JSON.stringify(data));
```
ericyhwang: As we discussed, this is _technically_ a breaking change, so I'd be more comfortable leaving the `connection.debug` flag for now and putting this on the list of candidates for 1.0 breaking changes.

That list isn't actually written yet, but I'll make a note of it for when we formally compile one.

alecgibson:
@ericyhwang I've added the `Connection.debug` shim - we should just revert that commit when releasing v1.0.

README.md
```diff
@@ -187,6 +187,25 @@ Note that only the [JSON0 OT type](https://github.com/ottypes/json0) is supporte
 `share.close(callback)`
 Closes connections to the database and pub/sub adapters.
 
+### Logging
+
+ShareDB logs some errors and warnings, which can be useful for debugging. By default, the log level is set to `warn`. This can be configured through the `logLevel` constructor argument, which can be set to any of the following:
```
ericyhwang: Looks like the default is actually set to `logger.levels.ERROR`?

curran: Great point!

lib/backend.js
```diff
@@ -48,6 +49,8 @@ function Backend(options) {
   if (!options.disableSpaceDelimitedActions) {
     this._shimAfterSubmit();
   }
+
+  logger.setLevel(options.logLevel || logger.levels.ERROR);
```
curran: The default is definitely set to `ERROR`. Please update the docs section so it matches.

lib/client/connection.js
```diff
@@ -5,6 +5,8 @@ var emitter = require('../emitter');
 var ShareDBError = require('../error');
 var types = require('../types');
 var util = require('../util');
+// TODO: Move back to the 'sharedb' logger name in v1.0
```
curran: I don't get why the `'sharedb'` logger name can't be used now. Where is the conflict?

This should probably be tracked in an issue, or added to the 1.0 TODO list.

Related https://github.com/share/sharedb/issues/144

lib/client/connection.js
```diff
@@ -56,7 +58,9 @@ function Connection(socket) {
   // state on the agent and read it in middleware
   this.agent = null;
 
+  // TODO: Remove this as a breaking changes in v1.0, and use the 'sharedb' logger and its level
   this.debug = false;
+  logger.setLevel(logger.levels.DEBUG);
```
curran: The client side logging appears not configurable at the moment?

curran:
One drawback here is that this new dependency adds 2.47 KB to the ShareDB client.

dcharbonnier:
Instead of adding a new dependency it should be an interface to plug any sort of logging system, it can default to a console.log driver. The effort to maintain a dependency up to date is not justified in this case.

lib/backend.js
alecgibson: Good catch. I'll actually update the default to `WARN` so that we aren't swallowing too much by default.

lib/client/connection.js
alecgibson: The problem is that I want to remove the `debug` flag in this class, which is sort of already trying to provide some control over log levels. In fact, leaving in the `debug` flag creates a lot of confusion, because if you set your log level to `TRACE`, but forget to set `debug = true`, then you won't get any output.

However, removing `debug` is technically a breaking change, so for now we need to give `Connection` its own logger which is always at `DEBUG`, and who output is actually controlled by the `debug` flag.

lib/client/connection.js
alecgibson: See above comment about the `debug` flag.

alecgibson:
@dcharbonnier @curran I already proposed this in my original [issue][1]. @nateps preferred to outsource to a 3rd-party library to restrict the surface area of our own API, and to avoid having to maintain more functions exposed to users.

I'm not particularly fussed myself. This seemed like the path of least resistance.

[1]: https://github.com/share/sharedb/issues/250

README.md
alecgibson: @ericyhwang setting it to `WARN` to keep in-line with the `loglevel` default, and to not swallow too much.

alecgibson:
@ericyhwang @nateps as discussed, I'll go back to the original implementation suggested in this issue: https://github.com/share/sharedb/issues/250

alecgibson:
Moving towards an override approach: https://github.com/share/sharedb/pull/258

PR #256, opened 2018-11-06
Create package-lock.json

### LINKED ISSUE #238: Check in a dependency lock file
~~It's npm's faster, safer cousin.~~

Lockfiles bring nice predictability in dependencies across multiple node versions as well.

~~It is basically a [drop in replacement for npm](https://yarnpkg.com/lang/en/docs/migrating-from-npm/) as far as devs are concerned.~~

Edited: npm is prefered over yarn, so the task here is to generate a `package-lock.json` and check it in.

curran:
Closes #238

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,1907 @@
+{
+  "name": "sharedb",
+  "version": "1.0.0-beta.14",
+  "lockfileVersion": 1,
+  "requires": true,
+  "dependencies": {
+    "@sinonjs/commons": {
+      "version": "1.3.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/commons/-/commons-1.3.0.tgz",
+      "integrity": "sha512-j4ZwhaHmwsCb4DlDOIWnI5YyKDNMoNThsmwEpfHx6a1EpsGZ9qYLxP++LMlmBRjtGptGHFsGItJ768snllFWpA==",
+      "dev": true,
+      "requires": {
+        "type-detect": "4.0.8"
+      }
+    },
+    "@sinonjs/formatio": {
+      "version": "3.0.0",
+      "resolved": "https://registry.npmjs.org/@sinonjs/formatio/-/formatio-3.0.0.tgz",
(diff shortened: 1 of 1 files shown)


alecgibson:
Thanks for getting around to doing this! While we're here, it might be good practice to update the [`engines` field][1] in `package.json`? Something like:

```json
{
  ...
  "engines": {
    "node": ">=6 <=10", // According to our Travis config
    "npm": ">=???" // It works on at least >= 5, but may go even lower...?
  }
  ...
}
```

[1]: https://docs.npmjs.com/files/package.json#engines

curran:
@alecgibson Great idea! Setting `engines` feels like it could be separate activity, with its own set of considerations. Perhaps open a new issue to discuss? I'm personally not too familiar with its effects or pros & cons.

ericyhwang:
We (Alec, Nate, and I) had a brief discussion about this today and did some research.

* Lockfiles aren't published by NPM, so users of the library will always get the newest versions that match semver, for an `npm update` or a fresh `npm install`. For that reason, one recommendation we found is "lockfiles for apps, but not libraries", so that library developers will be testing against the newest dependencies that could get installed for end users. https://github.com/sindresorhus/ama/issues/479#issuecomment-310661514
* On the other hand, Yarn recommends checking in lockfiles even for libraries. https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/#libraries-vs-applications

Nate's open to either checking in a package-lock.json or to checking in a .npmrc to disable package-lock. He says he just hasn't thought it through enough. I'm still reading up on the tradeoffs myself, with regard to lockfiles for libraries.

curran:
@ericyhwang To be honest I'm not completely sold on the added value of `package-lock`. IMO we can just leave it out and no one will be harmed. It's beneficial to have CI use the same dependencies library users will get, _precisely because_ it's more likely to break. Better if the breakage is caught in CI than by users. I'd be happy to close this PR and associated issue as "won't fix".

alecgibson:
@curran now that I've thought about it a bit, maybe I agree; there's something to be said for letting Travis break "spontaneously" with a weird, semver-non-compliant change by a dependency.

If we want to definitely not use `package-lock.json`, we should probably commit an `.npmrc` file with [`package-lock=false`][1] and an accompanying comment summarising the reasoning for omitting it.

[1]: https://codeburst.io/disabling-package-lock-json-6be662f5b97d

curran:
Closing due to inactivity.

PR #258, opened 2018-11-15
Allow custom logger overrides

alecgibson:
Fixes https://github.com/share/sharedb/issues/250
Supersedes https://github.com/share/sharedb/pull/252

This change adds the ability to override ShareDB's logging behaviour.

By default, ShareDB will still log to `console`. However, this default
can be overridden with custom methods on both the backend and in the
client.

## Supported methods

The ShareDB logger only supports the following methods:

  - `info`
  - `warn`
  - `error`

Any method that is not overridden will default to `console`.

## Backend

The backend methods can be overridden when instantiating `Backend`:

```javascript
var share = new Backend({
  logger: {
    info: () => {},                         // Silence info
    warn: () => alerts.warn(arguments),     // Forward warnings
    error: () => alerts.critical(arguments) // Map errors to critical
  }
});
```

## Client

Client methods can also be overridden:

```javascript
var ShareDB = require('sharedb/lib/client');
ShareDB.logger.setMethods({
  info: () => {},
  // etc.
});
```

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -2,7 +2,6 @@ var async = require('async');
 var Agent = require('./agent');
 var Connection = require('./client/connection');
 var emitter = require('./emitter');
-var logger = require('./logger');
 var MemoryDB = require('./db/memory');
 var NoOpMilestoneDB = require('./milestone-db/no-op');
 var MemoryPubSub = require('./pubsub/memory');
@@ -40,8 +39,6 @@ function Backend(options) {
   // The number of open agents for monitoring and testing memory leaks
   this.agentsCount = 0;
   this.remoteAgentsCount = 0;
-
-  logger.setMethods(options.logger);
 
   // The below shims are for backwards compatibility. These options will be
   // removed in a future major version
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -2,6 +2,7 @@ var async = require('async');
 var Agent = require('./agent');
 var Connection = require('./client/connection');
 var emitter = require('./emitter');
+var logger = require('./logger');
 var MemoryDB = require('./db/memory');
 var NoOpMilestoneDB = require('./milestone-db/no-op');
 var MemoryPubSub = require('./pubsub/memory');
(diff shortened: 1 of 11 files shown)


dcharbonnier:
a simple, well implemented logger class https://github.com/aurelia/logging/blob/master/src/index.js
the part giving the possibility to have more than one logger is not really interesting for sharedb but at least they define the methods, they don't overwrite the methods of an instance on the fly...

alecgibson:
@dcharbonnier I'm not sure I understand your improvement suggestion?

lib/backend.js
```diff
@@ -40,6 +41,8 @@ function Backend(options) {
   this.agentsCount = 0;
   this.remoteAgentsCount = 0;
 
+  logger.setMethods(options.logger);
```
ericyhwang: Now that I've read through the code, I do see what you mentioned around module caching - The `options.logger` is passed into the Backend constructor on instance creation but is then set onto the logger module, which is a process-level singleton. It is indeed a bit odd to call a constructor and have it change some module-level configuration.

If we don't want to pass around logger instances everywhere, perhaps a compromise would be to have the server also use a static config call, like you're doing with the client. And a name like `ShareDB.setDefaultLogger({...})` on server and client could provide a hint that it's not an instance-level logger.

alecgibson: @ericyhwang I've pushed a change that moves this to static configuration just like the client:

```javascript
Backend.logger.setMethods({...});
```

I personally think that this static accessor should make it clear to consumers that this will affect _all_ instances of `Backend`.

If you still want to change the method name, I'm not a huge fan of `setDefaultLogger`, because it implies there's some way of setting it on an instance-by-instance (non-default) way (which there isn't). I'd possibly prefer something along the lines of `setGlobalLoggingMethods` or similar.

Thoughts?

curran: How about `setLogger`?

alecgibson: I kind of think `logger.setLogger` is a bit weird, but I don't feel strongly.

alecgibson: Maybe..? But this is sort of why I landed on `setMethods`, because I thought it was nicer to be explicit about what it is you're setting.

ericyhwang: I think `setMethods` is good, thanks!

ericyhwang:
Published this as [sharedb@1.0.0-beta.15](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.15)

Happy US Thanksgiving for those who celebrate it! 🦃

curran:
Thanks @ericyhwang ! You too.

Issue #257, opened 2018-11-13
Doc.submitOp can throw errors when type.apply fails


alecgibson:
## What happens

`Doc.submitOp` internally [calls `type.apply`][1]. However, `type.apply` [could **throw an error**][2]!

This error is completely uncaught by the `Doc` class, and actually bubbles up to the caller, never calling the callback.

Example here: https://runkit.com/alecgibson/sharedb-doc-submit-throws

## What I expect to happen

I would expect `submitOp` to *always* call its callback, with an error if appropriate. I don't expect to have to "double-bag" all of my code eg:

```javascript
try {
  doc.submitOp(op, handleError);
} catch (error) {
  handleError(error);
}
```

## What I propose

We should potentially wrap the whole function in a `try`/`catch`, and call the callback with an error if we catch an error. It's probably worth thinking about where else this may happen.

[1]: https://github.com/share/sharedb/blob/a67ad42c384809eb377c9f70c7a45a7552a4be89/lib/client/doc.js#L560
[2]: https://github.com/ottypes/json0/blob/9df44f0c7958d2c98164e0682ef52f53e886ac45/lib/json0.js#L102-L117

alecgibson:
As discussed, I will:
 
  - wrap all use of `Doc._otApply` in a `try`/`catch` and attempt a `hardRollback`
  - change errors emitted inside `_otApply` to `throw` instead, so that we have consistent error behaviour

PR #259, opened 2018-11-15
Catch synchronous errors in `Doc._otApply`

alecgibson:
Fixes https://github.com/share/sharedb/issues/257

As outlined in the above issue, calling `Doc.submitOp` can sometimes
`throw` a synchronous error, for example when submitting an invalid op
that causes `type.apply` to `throw`.

This is surprising behaviour, because an error handler should already be
provided in the callback supplied to `submitOp`. What's more, the doc
could be left partially mutated and out-of-sync with the server, which
could lead to confusing behaviour.

This change wraps all internal usage of `Doc._otApply` in `try`/`catch`
blocks. If an error is thrown in this method, then we attempt to
recover with `Doc._hardRollback`, which will attempt to reset the
document and then call the callbacks of the pending ops with the error.

This change also prevents `Doc._hardRollback` from emitting an error if
at least one pending op callback was invoked, so that we don't branch
error handling behaviour.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -326,8 +326,11 @@ Doc.prototype._handleOp = function(err, message) {
   }
 
   this.version++;
-  this._otApply(message, false);
-  return;
+  try {
+    this._otApply(message, false);
+  } catch (error) {
+    return this._hardRollback(error);
+  }
 };
 
 // Called whenever (you guessed it!) the connection state changes. This will
@@ -511,8 +514,8 @@ function transformX(client, server) {
 Doc.prototype._otApply = function(op, source) {
   if (op.op) {
     if (!this.type) {
-      var err = new ShareDBError(4015, 'Cannot apply op to uncreated document. ' + this.collection + '.' + this.id);
-      return this.emit('error', err);
+      // Throw here, because all usage of _otApply should be wrapped with a try/catch
+      throw new ShareDBError(4015, 'Cannot apply op to uncreated document. ' + this.collection + '.' + this.id);
     }
 
     // Iteratively apply multi-component remote operations and rollback ops
@@ -654,8 +657,12 @@ Doc.prototype._submit = function(op, source, callback) {
     if (this.type.normalize) op.op = this.type.normalize(op.op);
   }
 
-  this._pushOp(op, callback);
-  this._otApply(op, source);
+  try {
+    this._pushOp(op, callback);
+    this._otApply(op, source);
+  } catch (error) {
+    return this._hardRollback(error);
+  }
 
   // The call to flush is delayed so if submit() is called multiple times
   // synchronously, all the ops are combined before being sent to the server.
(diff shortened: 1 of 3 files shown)


lib/client/doc.js
```diff
@@ -889,7 +900,7 @@ Doc.prototype._hardRollback = function(err) {
   this.fetch(function() {
     var called = op && callEach(op.callbacks, err);
     for (var i = 0; i < pending.length; i++) {
-      callEach(pending[i].callbacks, err);
+      called = callEach(pending[i].callbacks, err) || called;
```
alecgibson: @nateps @ericyhwang do you know if this change will have any weird side-effects? Originally it would not count a pending callback as having called something, so it would both call back, _and_ emit, which seems a bit odd?

It's forced some updates to a couple of edge case tests, which I'll highlight also. I'm assuming they're okay to change...? In them, we actually assert that the callback is called _and_ an error is emitted; is that the expected behaviour?

test/client/submit.js
```diff
@@ -628,19 +628,12 @@ describe('client submit', function() {
         if (err) return done(err);
         doc2.del(function(err) {
           if (err) return done(err);
-          doc.pause();
-          var calledBack = false;
-          doc.on('error', function() {
```
alecgibson: Is this emission actually expected, given the callback is called? (See above comment)

test/client/submit.js
```diff
@@ -657,19 +650,12 @@ describe('client submit', function() {
           if (err) return done(err);
           doc2.create({age: 5}, function(err) {
             if (err) return done(err);
-            doc.pause();
-            var calledBack = false;
-            doc.on('error', function() {
```
alecgibson: Is this emission actually expected, given the callback is called? (See above comment)

alecgibson:
Also apologies for the code coverage decrease, but I don't have any idea how to reach the other `catch` blocks without either:

  - directly calling the internal methods (which feels like cheating)
  - abusing middleware to mutate `doc.data` after `_submitOp`, but before `_handleOp` (which also feels like cheating)

alecgibson:
Okay - code coverage back up I think. I did actually manage to work out some crazy contrived (but theoretically possible) way that `type.apply` could throw after the op was successfully submitted.

alecgibson:
As discussed, we should check that **every** op has a callback. If any of them don't, then we should emit an error (so that no errors are ever swallowed).

alecgibson:
@ericyhwang @nateps I've updated the hard rollback callback logic. I've actually tweaked it a tiny bit more than discussed - the inflight op is no longer treated as "special" relative to other pending ops. Here's the commit message for clarity:

We previously followed the following logic for callbacks in a hard
rollback:

  - check there's an inflight op, and that it has a callback
  - if this isn't the case, then emit the error

This logic doesn't work very well in the situation where we don't have
an inflight op. This can happen in the following case:

  - I submit an invalid op
  - the op is added to the pending ops queue
  - we attempt to apply the op to the local document before flushing
  - `type.apply` throws, so the op is never flushed from pending to
    inflight
  - we perform a hard rollback
  - we now find we have no inflight op, but we do have a pending op

In this case, since we have no inflight op, we'd emit the error, but
also call the pending op callback, causing both a callback call _and_
an error emission, which is a bit surprising.

This change updates our logic, so that:

  - we check that there are some ops to callback
  - and that all of those ops have callbacks

If there are no callbacks, or if any of the ops don't handle the error,
then we emit, so that we can be sure of never swallowing an error. We
also avoid treating the inflight op any differently to pending ops, so
that we avoid behaviour that's dependent on where in its lifecycle an op
throws.

However, if the client handles all of their op submission errors, then
we never emit.

PR #261, opened 2018-11-30
fix leak on duplicate subscription - #260

### LINKED ISSUE #260: memory leak
Share db is leaking when multiple subscriptions to the same document.
OpStream objects are not correctly released, the problem is in the agent :

```
  var previous = streams[id];
  if (previous) previous.destroy();
  streams[id] = stream;
  [...]
  stream.on('end', function() {
    [...]
    delete streams[id];
    [...]
  });
```

When there is a previous stream, we destroy it, this destroy will trigger a "end" event when we catch the event streams[id] is not anymore the expected stream and we loose a reference to an open stream so we can't clear it when the agent is destroyed.
Two possibilities :
  - `if(streamd[id] !== stream) return;` best option for me
  - `if(stream[id].open) return`

PS : maybe that was the sense of #190 and it's not a wrong usage of sharedb it's a bug.

dcharbonnier:
When we subscribe to the same stream sharedb leak, because of the cross reference between the stream and the agent the leak is even more critical.
close #260

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -105,7 +105,7 @@ Agent.prototype._subscribeToStream = function(collection, id, stream) {
   stream.on('end', function() {
     // The op stream is done sending, so release its reference
     var streams = agent.subscribedDocs[collection];
-    if (!streams) return;
+    if (!streams || streams[id] !== stream) return;
     delete streams[id];
     if (util.hasKeys(streams)) return;
     delete agent.subscribedDocs[collection];
--- a/test/client/connection.js
+++ b/test/client/connection.js
@@ -56,6 +56,27 @@ describe('client connection', function() {
     });
   });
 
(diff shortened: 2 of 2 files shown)


dcharbonnier:
you're right, go for it, thanks

dcharbonnier:
you can also fix the same potential bug here https://github.com/share/sharedb/blob/master/lib/backend.js#L422 and and other few lines after

ericyhwang:
OK, I wrote a test that fails without your change and passes with your change. I'll get this merged and published now.

dcharbonnier:
initProjection will create a reference to the agent and if feel line after you return an error (without the stream obviously) there will be no way later to destroy the stream and the agent and stream will remain for ever in memory. I didn't observed this behavior but reading the code I see a potential problem here too.

PR #262, opened 2018-11-30
Fetch snapshot by time

alecgibson:
This change adds the ability fetch a snapshot by time. The motivation
for this is that fetching a document by time is quite a "natural" way
to think about document history, and allows us to - for example - fetch
multiple documents as they were at a given time, without having to
look up their exact version numbers first.

We add a new `Connection.fetchSnapshotByTimestamp` method, which
follows a very similar route to `Connection.fetchSnapshot`, and where
possible, as much code is re-used as possible:

  - both methods use a subclassed child of `SnapshotRequest`
  - both methods have their requests handled by the same machinery in
    `Connection`
  - both methods in the `Backend` have ops applied by a common method,
    but use their own methods for calls to middleware

In order to make this feature possible at scale, this change also adds
two new methods to the `MilestoneDB` interface:

  - `getMilestoneSnapshotAtOrBeforeTime`
  - `getMilestoneSnapshotAtOrAfterTime`

These methods are used to fetch milestone snapshots either side of the
requested timestamp, which means we only need to fetch the ops between
the two of them to reach the desired timestamp.

In the case where a milestone database is not being used, then fetching
a snapshot by timestamp is still possible, but it will fetch all the ops
for a document, and keep applying them from v0 until the timestamp is
reached, which is not particularly scalable.


lib/backend.js
```diff
@@ -637,45 +641,120 @@ Backend.prototype._fetchSnapshot = function (collection, id, version, callback)
...
+        backend._buildSnapshotFromOps(id, milestoneSnapshot, ops, callback, function shouldBreak(nextOp) {
+          var opTimestamp = nextOp && nextOp.m && nextOp.m.ts;
+          return timestamp !== null && opTimestamp > timestamp;
+        });
+      });
+    });
+  });
+};
+
+Backend.prototype._buildSnapshotFromOps = function (id, startingSnapshot, ops, callback, shouldBreak) {
```
ericyhwang: Can you document what the `shouldBreak` param does?

lib/backend.js
```diff
@@ -637,45 +641,120 @@ Backend.prototype._fetchSnapshot = function (collection, id, version, callback)
...
+    if (snapshot) from = snapshot.v;
+
+    milestoneDb.getMilestoneSnapshotAtOrAfterTime(collection, id, timestamp, function (error, snapshot) {
+      if (error) return callback(error);
+      if (snapshot) to = snapshot.v;
+
+      var options = {metadata: true};
+      db.getOps(collection, id, from, to, options, function (error, ops) {
+        if (error) return callback(error);
+        backend._buildSnapshotFromOps(id, milestoneSnapshot, ops, callback, function shouldBreak(nextOp) {
```
ericyhwang: As discussed, the `shouldBreak` was a bit confusing at first. Couple of options we came up with:

* Your idea of removing `shouldBreak`, instead pre-filtering the ops passed in.
* Passing in a `{maxOpTimestamp: ___}` options object.

I don't have a strong preference, as `_buildSnapshotFromOps` is a purely internal function.

lib/client/connection.js
```diff
@@ -634,7 +636,35 @@ Connection.prototype.fetchSnapshot = function(collection, id, version, callback)
...
 
   var requestId = this.nextSnapshotRequestId++;
-  var snapshotRequest = new SnapshotRequest(this, requestId, collection, id, version, callback);
+  var snapshotRequest = new SnapshotVersionRequest(this, requestId, collection, id, version, callback);
+  this._snapshotRequests[snapshotRequest.requestId] = snapshotRequest;
+  snapshotRequest.send();
+};
+
+/**
+ * Fetch a read-only snapshot at a given version
```
ericyhwang: Copypasta fix: version -> timestamp

alecgibson:
@nateps / @ericyhwang as discussed, I've moved the snapshot building function into `ot.js`.

ericyhwang:
LGTM from me and Nate

PR #263, opened 2019-01-16
For readSnapshots middleware, add method and parameter properties to first parameter

ericyhwang:
When monitoring for unusual response sizes, it can be useful to record which client request was associated with the response. Unfortunately, ShareDB's current "readSnapshots" middleware, which is how you get info about snapshot responses, doesn't contain information about the associated request.

This PR adds a couple additional fields to the "readSnapshots" middleware's `request` param, `method` and `parameters`.

The "readSnapshots" middleware function now has a first parameter `request` like this:

```javascript
function(request, next) {
  let {
    collection,
    snapshots,
    method,  // Backend.prototype.READ_SNAPSHOTS_METHODS: 'fetch', 'query', etc.
    parameters,  // Plain JS object, shape varies depending on method
    snapshotType,  // Backend.prototype.SNAPSHOT_TYPES
    agent
  } = request;
}
```

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -272,7 +272,7 @@ Backend.prototype._sanitizeOpsBulk = function(agent, projection, collection, ops
   }, callback);
 };
 
-Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, snapshots, snapshotType, callback) {
+Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, snapshots, requestContext, callback) {
   if (projection) {
     try {
       projections.projectSnapshots(projection.fields, snapshots);
@@ -281,13 +281,10 @@ Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, s
     }
   }
 
-  var request = {
-    collection: collection,
-    snapshots: snapshots,
-    snapshotType: snapshotType
-  };
+  requestContext.collection = collection;
+  requestContext.snapshots = snapshots;
 
-  this.trigger(this.MIDDLEWARE_ACTIONS.readSnapshots, agent, request, callback);
+  this.trigger(this.MIDDLEWARE_ACTIONS.readSnapshots, agent, requestContext, callback);
 };
 
 Backend.prototype._getSnapshotProjection = function(db, projection) {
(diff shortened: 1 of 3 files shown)


lib/backend.js
```diff
@@ -578,7 +591,16 @@ Backend.prototype._query = function(agent, request, callback) {
   var backend = this;
   request.db.query(request.collection, request.query, request.fields, request.options, function(err, snapshots, extra) {
     if (err) return callback(err);
-    backend._sanitizeSnapshots(agent, request.snapshotProjection, request.collection, snapshots, backend.SNAPSHOT_TYPES.current, function(err) {
+    var requestContext = {
+      requestMethod: 'query',  // TODO: Should this distinguish between queryFetch and querySubscribe?
```
ericyhwang: Question here for reviewers - This currently doesn't distinguish between "queryFetch" and the initial query issued by "querySubscribe". Should it?

The initial "querySubscribe" DB query is for all intents and purposes identical to the query issued by "queryFetch". The big difference is in whether a pub/sub subscription is registered.

lib/backend.js
```diff
@@ -353,20 +350,28 @@ Backend.prototype.getOpsBulk = function(agent, index, fromMap, toMap, callback)
 Backend.prototype.fetch = function(agent, index, id, callback) {
   var start = Date.now();
   var projection = this.projections[index];
-  var collection = (projection) ? projection.target : index;
+  var dbCollection = (projection) ? projection.target : index;
```
nateps: Let's leave this as `collection` for this PR. we could consider changing it, but we'd want to be consistent everywhere

lib/backend.js
```diff
@@ -353,20 +350,28 @@ Backend.prototype.getOpsBulk = function(agent, index, fromMap, toMap, callback)
...
     id: id
   };
-  backend.db.getSnapshot(collection, id, fields, null, function(err, snapshot) {
+  backend.db.getSnapshot(dbCollection, id, fields, null, function(err, snapshot) {
     if (err) return callback(err);
     var snapshotProjection = backend._getSnapshotProjection(backend.db, projection);
     var snapshots = [snapshot];
-    backend._sanitizeSnapshots(agent, snapshotProjection, collection, snapshots, backend.SNAPSHOT_TYPES.current, function(err) {
+    var requestContext = {
+      requestMethod: 'fetch',
```
nateps: let's use `method`

lib/backend.js
```diff
@@ -353,20 +350,28 @@ Backend.prototype.getOpsBulk = function(agent, index, fromMap, toMap, callback)
...
+      requestParams: {
```
nateps: let's use `parameters`

lib/backend.js
nateps: yes, let's make the method be the same name as the public method on Backend,  so `queryFetch` or `querySubscribe`

lib/backend.js
```diff
@@ -281,13 +281,10 @@ Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, s
     }
   }
 
-  var request = {
-    collection: collection,
-    snapshots: snapshots,
-    snapshotType: snapshotType
-  };
+  requestContext.collection = collection;
```
alecgibson: It might be a good optimisation to initialise these properties whenever `requestContext` is initialised. It may be beneficial to declare a `class` for clarity about what exact information we're expecting in the middleware.

lib/backend.js
```diff
@@ -366,7 +363,15 @@ Backend.prototype.fetch = function(agent, index, id, callback) {
     if (err) return callback(err);
     var snapshotProjection = backend._getSnapshotProjection(backend.db, projection);
     var snapshots = [snapshot];
-    backend._sanitizeSnapshots(agent, snapshotProjection, collection, snapshots, backend.SNAPSHOT_TYPES.current, function(err) {
+    var requestContext = {
+      method: 'fetch',
```
alecgibson: It would be nice to move these magic strings into an object, similar to the middleware actions. Just to clarify which methods are available, and to make refactoring easy, etc.

ericyhwang:
Thanks for the review, changes made! Give it another look when you've got the chance.

alecgibson:
@ericyhwang changes look good to me. Just greenify the build, and I'll approve.

ericyhwang:
I hadn't merged in your changes from last week the first time, and your tests caught the fact that I needed to support the new method too - thanks! Tests and CI doing their job. :)

ericyhwang:
https://github.com/share/sharedb/issues/269 reported that this caused a regression in `'readSnapshots'` middleware for `queryPoll()` calls specifically, so I unpublished the new version from NPM.

I'm going to revert this for now so that forks don't pick it up. I'll work on adding a test for the queryPoll+readSnapshots case, fix the issue, and create a new PR.

Issue #272, opened 2019-02-25
Document.create callback called twice


nickasd:
In the following code, both `error 1` and `error 2` get logged, which looks like the callback to `doc.create` is getting called twice, and both times the error is `Error: Delete component '0' does not match deleted text 'a'`. Some initial analysis done by @alecgibson in #162.

```
var ShareDB = require('sharedb');
var sharedb = new ShareDB();
var connection = sharedb.connect();
var doc = connection.get('files', '0');
doc.create({content: 'asdf'}, (error) => {
    if (error) {
        console.log('error 1');
        console.error(error);
    } else {
        var ops = [{p: 0, d: '0'}];
        var op = [{p: ['content'], t: 'text0', o: ops}];
        doc.submitOp(op, (error) => {
            if (error) {
                console.log('error 2');
                console.error(error);
            } else {
                console.log('success');
            }
        });
    }
});
```

alecgibson:
My thoughts replicated for completeness:

I've had a very quick dig, and it looks like it has to do with the combination of `_clearInflightOp` and `_hardRollback`.

`_clearInflightOp` tries to [call each of the callbacks][1] before clearing the `inflightOp`. In this case, that includes calling a callback which will produce an error. The `_hardRollback` then [_also_ calls the create callback][2], which - as you correctly point out - means that we call the creation op callback twice.

We've got something like this happening:

  - submit creation
  - set creation to inflight op
  - receive acknowledgement of creation
  - call the creation op callback
  - creation op callback includes a bad op submission
  - bad op submission errors
  - we attempt to hard rollback
  - hard rollback tries to call the create callback _again_

I don't have a lot of time to think about how best to handle this right now - may be best to chat with @nateps and @ericyhwang in Wednesday's meeting. Potentially best to also move this to a new issue.

My first instinct is that when we invoke a callback, we should first dissociate it from the op it was attached to, so that we don't get this repeat call (that or mark it as called, so we don't try to call it again), but I don't know what the ramifications of that behaviour might be.

Actually, now thinking about it a bit more, perhaps we _should_ be moving more towards having a light `Callback` class, which remembers if it was called or not, because we're already [trying to keep track of that sort of thing anyway][3].

[1]: https://github.com/share/sharedb/blob/70b881509da1b7c23cccb60dfec4eed98815dc74/lib/client/doc.js#L923
[...]

PR #273, opened 2019-02-28
In Doc._clearInflightOp, clear inflightOp before calling callbacks

ericyhwang:
This fixes a double-callback issue for chained op submissions, where if the second op is invalid, the first op's callback gets called twice:
https://github.com/share/sharedb/issues/272

`Doc._hardRollback` [already does takes a similar approach](https://github.com/share/sharedb/blob/v1.0.0-beta.18/lib/client/doc.js#L896-L904) of clearing local Doc state prior to calling callbacks, and Nate expressed his agreement with this fix today at the PR review meeting.

Thanks to @nickasd and @alecgibson for developing the test case that reproduces the issue!

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -920,9 +920,12 @@ Doc.prototype._hardRollback = function(err) {
 };
 
 Doc.prototype._clearInflightOp = function(err) {
-  var called = callEach(this.inflightOp.callbacks, err);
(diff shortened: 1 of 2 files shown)


ericyhwang:
Published in [sharedb@1.0.0-beta.20](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.20)

PR #276, opened 2019-03-05
Update example in README.md to use @teamwork/websocket-json-stream

gkubisa:


gkubisa:
Yes, it looks good to me.

PR #278, opened 2019-03-27
Add new "reply" middleware action

ericyhwang:
The "reply" middleware is called just before the backend is about to send a non-error reply to a client.

It's the flip side of the existing "receive" middleware action.

The middleware function gets a context with these properties:
- request
- reply
- action (always `'reply'` for this middleware)
- agent
- backend

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -209,12 +209,7 @@ Agent.prototype._reply = function(request, err, message) {
   var backend = agent.backend;
   if (err) {
     request.error = getReplyErrorObject(err);
-    var middlewareContext = {request: request, reply: request};
-    backend.trigger(backend.MIDDLEWARE_ACTIONS.reply, agent, middlewareContext, function(_err) {
-      // If we were already going to send back an error, and the reply middleware runs into
-      // another error (`_err`), just send back the original error.
-      agent.send(middlewareContext.reply);
-    });
+    agent.send(request);
     return;
   }
   if (!message) message = {};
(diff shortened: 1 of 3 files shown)


lib/agent.js
```diff
@@ -187,23 +187,34 @@ Agent.prototype._sendOps = function(collection, id, ops) {
...
-      }
-      request.error = {
-        code: err.code,
-        message: err.message
-      };
-    }
-    this.send(request);
+    request.error = getReplyErrorObject(err);
+    var middlewareContext = {request: request, reply: request};
+    backend.trigger(backend.MIDDLEWARE_ACTIONS.reply, agent, middlewareContext, function(_err) {
```
nateps: If we have an error, let's not emitted the middleware event instead.

ericyhwang:
Released in [sharedb@1.0.0-beta.21](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.21)

curran:
What is the purpose of the "reply" middleware?

I see there is no documentation at all added in this PR. The tests serve as the documentation.

ericyhwang:
@curran

> What is the purpose of the "reply" middleware?

Whatever, really - a couple example use-cases:

* Pushing metadata about replies into a structured logging store like an ELK stack
* Adding additional properties onto replies, like trace ids

It's a pretty low-level middleware, sitting just above the messaging layer, similar to the existing 'receive' middleware.

> I see there is no documentation at all added in this PR. The tests serve as the documentation.

I indeed forgot to update the README, thanks for the reminder! Just made a new PR for it: https://github.com/share/sharedb/pull/279

PR #281, opened 2019-04-06
Use @teamwork/websocket-json-stream in textarea example

### LINKED ISSUE #280: Use @teamwork/websocket-json-stream in textarea example
Since the [example in README.md uses @teamwork/websocket-json-stream](https://github.com/share/sharedb/pull/276/files), I thought it would make sense to also have the code examples use it. We can start with the textarea example, as it's been most recently updated, then perhaps update the others as well.

Although, there may be a better "best practice" nowadays, such as SockJS, I'm not really sure what the landscape looks like.

curran:
Closes  #280

Since the [example in README.md uses @teamwork/websocket-json-stream](https://github.com/share/sharedb/pull/276/files), I thought it would make sense to also have the code examples use it.

There is a slight issue at the moment with this change, I'm not sure why. It's documented here [Teamwork/websocket-json-stream: Error with ShareDB Usage](https://github.com/Teamwork/websocket-json-stream/issues/3).

If we proress here, I'd like to also propose updating the other examples in the same way, then once they work, update all their dependencies generally, which are getting out of date.

Related to discussion here https://github.com/share/sharedb/issues/275

### DIFF
--- a/examples/textarea/package.json
+++ b/examples/textarea/package.json
@@ -18,7 +18,7 @@
     "reconnecting-websocket": "^3.0.3",
     "sharedb": "^1.0.0-beta",
     "sharedb-string-binding": "^1.0.0",
-    "websocket-json-stream": "^0.0.1",
+    "@teamwork/websocket-json-stream": "^2.0.0",
     "ws": "^1.1.0"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


curran:
Now that #283 has been merged and released in [v1.0.0-beta.22](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.22), I can verify that this works well and does not suffer from the original bug observed (https://github.com/share/sharedb/issues/282) (just verified locally with fresh `npm install`).

ericyhwang:
Thanks for checking that it's all working now! I agree with making the changes elsewhere, too.

PR #283, opened 2019-04-09
Clean up closed connections

### LINKED ISSUE #282: [ERR_STREAM_DESTROYED] seems sharedb is not closing connections
Hey! I just upgraded to teamwork streamer.

ShareDb: 1.0.0-beta.21
sharedb-mongo: 1.0.0-beta.8
reconnecting-websocket: 4.1.10
@teamwork/websocket-json-stream: 2.0.0
rich-text: 3.1.0
ws: 6.2.1

node: 11.13.0
npm: 6.9.0

When I open page, everything is working right. Submitting op works without issues. Then I refresh the page and after first submitOp it throws an error.

Error what is crashing node server:
```
Error [ERR_STREAM_DESTROYED]: Cannot call write after a stream was destroyed
    at doWrite (_stream_writable.js:411:19)
    at writeOrBuffer (_stream_writable.js:399:5)
    at WebSocketJSONStream.Writable.write (_stream_writable.js:299:11)
    at Agent.send (/projects/alganep/server/node_modules/sharedb/lib/agent.js:165:15)
    at Agent._sendOp (/projects/alganep/server/node_modules/sharedb/lib/agent.js:181:8)
    at OpStream.<anonymous> (/projects/alganep/server/node_modules/sharedb/lib/agent.js:103:11)
    at OpStream.emit (events.js:193:13)
    at addChunk (_stream_readable.js:296:12)
    at readableAddChunk (_stream_readable.js:277:11)
    at OpStream.Readable.push (_stream_readable.js:232:10)
```

Last data what reached server before crash (immediately after it)
```
bubbles: false
cancelBubble: false
cancelable: false
composed: false
currentTarget: WebSocket {__zone_symbol__openfalse: null, __zone_symbol__closefalse: null, __zone_symbol__messagefalse: null, __zone_symbol__errorfalse: null, url: "ws://alganep:7000/", …}
data: "{"src":"1c7ce151b76c9abff41d499afc4640db","seq":1,"v":85,"a":"op","c":"texts","d":"17435_Tester"}"
defaultPrevented: false
eventPhase: 0
isTrusted: true
lastEventId: ""
origin: "ws://alganep:7000"
path: []
ports: []
returnValue: true
source: null
srcElement: WebSocket {__zone_symbol__openfalse: null, __zone_symbol__closefalse: null, __zone_symbol__messagefalse: null, __zone_symbol__errorfalse: null, url: "ws://alganep:7000/", …}
target: WebSocket {__zone_symbol__openfalse: null, __zone_symbol__closefalse: null, __zone_symbol__messagefalse: null, __zone_symbol__errorfalse: null, url: "ws://alganep:7000/", …}
timeStamp: 11002.040000050329
type: "message"
userActivation: null
decoded: (...)
deepCopy: (...)
encoded: (...)
__proto__: MessageEvent
index.js?9e40:25 
```

### UPDATE
I find out for every refresh is error duplicated.
So when I open page first time then the error comes only once.
When I refresh the page 10 times and let it connect everytime to ShareDb. Then the error is printed 10 times.

Seems like memory-leak?

I attached listener for stream#close and it is called everytime. 
eg.:
```
// Connect any incoming WebSocket connection to ShareDB
	const webSocket = new WebSocket.Server({server: server})
	webSocket.on("connection", function (ws, req) {
		console.warn("Incoming websocket connection")
		const stream = new WebSocketJSONStream(ws)
		stream.on('error', error => {
			console.log("----------------------------------------")
			console.log(error)
		})
		stream.on("data", message => {
			console.log("data", message)
		})
		stream.on("close", () => {
			console.log("stream closed...")
		})
		sharedb.listen(stream)
	})
```

curran:
The intention here is to resolve #282.

As of now this is a rough first pass. I have not tested or verified if this works yet.

@gkubisa Does this approach seem to make sense?

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -258,12 +258,15 @@ Agent.prototype._open = function() {
       agent._handleMessage(request.data, callback);
     });
   });
-
-  this.stream.on('end', function() {
+  
+  function cleanup() {
     agent.backend.agentsCount--;
     if (!agent.stream.isServer) agent.backend.remoteAgentsCount--;
     agent._cleanup();
-  });
+  }
+  
+  this.stream.on('end', cleanup);
+  this.stream.on('close', cleanup);
 };
 
 // Check a request to see if its valid. Returns an error if there's a problem.
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -56,7 +56,14 @@ Agent.prototype.close = function(err) {
 };
 
 Agent.prototype._cleanup = function() {
+
+  // Only clean up once if the stream emits both 'end' and 'close'.
+  if (this.closed) return;
+
   this.closed = true;
+
+  this.backend.agentsCount--;
+  if (!this.stream.isServer) this.backend.remoteAgentsCount--;
 
   // Clean up doc subscription streams
   for (var collection in this.subscribedDocs) {
@@ -259,12 +266,7 @@ Agent.prototype._open = function() {
     });
   });
   
-  function cleanup() {
-    agent.backend.agentsCount--;
-    if (!agent.stream.isServer) agent.backend.remoteAgentsCount--;
-    agent._cleanup();
-  }
-  
+  var cleanup = agent._cleanup.bind(agent);
   this.stream.on('end', cleanup);
   this.stream.on('close', cleanup);
 };
--- a/test/client/connection.js
+++ b/test/client/connection.js
@@ -106,10 +106,19 @@ describe('client connection', function() {
       var connection = backend.connect();
       connection.on('connected', function() {
         connection.socket.stream.emit('close')
-        setTimeout(function() {
-          expect(backend.agentsCount).equal(0);
-          done();
-        }, 10);
+        expect(backend.agentsCount).equal(0);
+        done();
+      });
+    });
+
+    it('updates correctly after stream emits both "end" and "close"', function(done) {
+      var backend = this.backend;
+      var connection = backend.connect();
+      connection.on('connected', function() {
+        connection.socket.stream.emit('end')
+        connection.socket.stream.emit('close')
+        expect(backend.agentsCount).equal(0);
+        done();
       });
     });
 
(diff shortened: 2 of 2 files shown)


gkubisa:
The general idea is good, however, as it is, the agent counts would be decremented twice for streams that emit both `end` and `close` events.

I'd move agent counting to `Agent#_cleanup` and start that method with `if (this.closed) return;`.

curran:
Thanks @gkubisa for the guidance. I implemented it as you suggested, and also added tests.

I believe this solves #282.

This PR, by the way, would unblock merging of https://github.com/share/sharedb/pull/281

Kindly requesting review by @ericyhwang or @nateps . Thanks!

lib/agent.js
```diff
@@ -56,8 +56,15 @@ Agent.prototype.close = function(err) {
 };
 
 Agent.prototype._cleanup = function() {
+
+  // Only clean up once if the stream emits both 'end' and 'close'.
+  if (this.closed) return;
```
ericyhwang: Nice, this will be defensive against 'end' accidentally being emitted twice by the stream too, which the old code didn't do.

lib/agent.js
```diff
@@ -258,12 +265,10 @@ Agent.prototype._open = function() {
...
   });
-
-  this.stream.on('end', function() {
-    agent.backend.agentsCount--;
-    if (!agent.stream.isServer) agent.backend.remoteAgentsCount--;
-    agent._cleanup();
-  });
+  
+  var cleanup = agent._cleanup.bind(agent);
+  this.stream.on('end', cleanup);
```
ericyhwang: Suggestion by Nate for the future, paraphrased by me:

> In the future, we should consider _not_ reacting to `'end'`, and only reacting to `'close'`. He says that Share originally reacted to `'end'` since the original stream implementation by Avital emitted both `'end'` and `'close'`, since that was the recommendation for old versions of Node.
> 
> Nowadays, we probably no longer need to do that.

ericyhwang:
This is published as [sharedb@1.0.0-beta.22](https://github.com/share/sharedb/releases/tag/v1.0.0-beta.22)

PR #286, opened 2019-04-17
WIP Presence

curran:
A continuation of #207.

While resolving merge conflicts, I encountered the following:

* `4024 - OT Type does not support presence` - I changed this to `4027`, as the code `4024` was introduced to mean something else since this PR was submitted.
* In `Doc.prototype._hardRollback` - This PR introduces a solution for not dropping errors, whereas a competing solution already existed (from [Update hard rollback callback logic](Update hard rollback callback logic) by @alecgibson ). I did my best to reconcile the two, favoring the one already presence in master.


curran:
The failing test in CI is `expires cached ops`, which appears to depend on 1ms time resolution of `setTimeout`, which makes it flaky. I will attempt to use lolex here.

curran:
I don't get the advantage of `Object.create(null)` over `{}` for this case, so I'm changing these to simplify. I'm curious @gkubisa , are there any big wins that we get from using `Object.create(null)` instead of `{}`? I understand there's a [difference in the prototype](https://stackoverflow.com/questions/15518328/creating-js-object-with-object-createnull), but that difference appears inconsequential. Probably there's no performance difference (if anything, I'd guess `Object.create(null)` would be slower due to the additional property lookup and function call).

gkubisa:
If anything, I would change those to `Map` instances because that's how they are used. `Object.create(null)` is the closes thing to `Map`. The problem with using `{}` as a map is that some properties exist on the Object prototype, so to be 100% sure that an entry exists in the "map" you'd have to use `hasOwnProperty(...)`.

I'm not sure about the performance of `Object.create(null)` vs `{}`, however, that should be inconsequential in this case. An object with a `null` prototype should be a bit faster when accessing non-existent properties because then the prototype does not need to be checked for that property.

curran:
Thanks @gkubisa for the clarification. Since these objects are often generated externally, and may need to survive `JSON.stringify` and `JSON.parse` anyway, it feels better IMO to use `{}` and live with the Object prototype.

This is ready for review by @nateps & @ericyhwang .

The only further change I'm considering is to make the implementation robust to OT types that do not implement `createPresence` and `comparePresence`, leaving more flexibility in the definition of the spec (related issue: [ottypes/docs: Presence for OT Types](https://github.com/ottypes/docs/issues/29)). At that point, the only hard requirement for OT types would be `transformPresence`.

The following could serve as default stand-ins, if these methods are not provided by OT types (from this [json0 fork](https://github.com/datavis-tech/json0/blob/master/lib/json0.js#L654)):

```js
json.createPresence = function(presenceData) {
  return presenceData;
};

json.comparePresence = function(pres1, pres2) {
  return JSON.stringify(pres1) === JSON.stringify(pres2);
};
```

Thoughts?

curran:
I moved presence logic into a new module, but this is only really used for `Doc` methods. It could also be made to house presence-related logic for `Agent`, `Connection` and `Backend` as well. Not sure whether or not this is advisable.

curran:
Stuck on this failing test:

```
  1) client presence (wrapped-presence-no-compare)
       waits for pending ops before processing future presence:
     ShareDBError: Cannot submit op. Document has not been created. dogs.fido
      at Doc._submit (lib/client/doc.js:703:17)
      at Doc.submitOp (lib/client/doc.js:814:8)
      at Context.<anonymous> (test/client/presence.js:108:22)
      at exports.callEach (lib/util.js:31:7)
      at StatelessPresence.handlePresence (lib/presence/stateless.js:106:33)
      at Doc._handlePresence (lib/client/doc.js:373:17)
      at Connection.handleMessage (lib/client/connection.js:259:20)
      at StreamSocket.socket.onmessage (lib/client/connection.js:132:18)
      at /home/curran/repos/sharedb/lib/stream-socket.js:60:12
      at processTicksAndRejections (internal/process/next_tick.js:74:9)
```

curran:
Closing in favor of https://github.com/share/sharedb/pull/288

PR #288, opened 2019-04-18
Presence 2

curran:
Summary of changes

 * This PR includes most commits from https://github.com/share/sharedb/pull/286 (hit a a dead end of red tests there)
 * Split out presence logic completely from ShareDB core.
 * The implementation of presence is exposed as `statelessPresence`, which can be used like this:

```js
var ShareDB = require('sharedb');
var statelessPresence = require('sharedb/lib/presence/stateless');
var share = new ShareDB({ presence: statelessPresence })`).
```

 * The value of the `presence` option is expected to contain implementations of the following classes:
   * `DocPresence`, which is instantiated at `doc._docPresence`. 
   * `ConnectionPresence`, which is instantiated at `doc._connectionPresence`. 
   * `AgentPresence`, which is instantiated at `doc._agentPresence`.
   * `BackendPresence`, which is instantiated at `doc._backendPresence`. 
 * `Doc`, `Connection`, `Agent` and `Backend` invoke methods on their respective presence classes.
 * By default a `dummyPresence` is used, which implements methods that mostly no-ops.
 * 100% of the presence logic is encapsulated within `statelessPresence`, so
 * it should be possible to swap out better implementations in the future, and
 * it's possible to develop third party presence implementations external to ShareDB.
 * API compatibility is mostly retained between this and the Teamwork ShareDB fork.
   * The main difference is that the presence implementation needs to be passed into the constructor.
   * Presence is exposed as `doc.presence`, as in the original implementation.
 * See this PR to compare these changes to the original Presence PR: https://github.com/Teamwork/sharedb/pull/3


curran:
OK great I found a promising path forward - migrate methods one at a time, keeping the tests green at each step.

curran:
Remaining tasks:

 * [ ] Use a `DummyPresence`, remove the `if(this.presence)` checks.

curran:
Now that presence logic is isolated within Agent and Connection, the next step is to split it out into the presence implementation. Not sure the best way to go about this, but thinking to rename the existing `Presence` class to `DocPresence`, then introduce similar classes `AgentPresence` and `ConnectionPresence` (and maybe `BackendPresence`). Instances of these would be `doc.presence`, `agent.presence`, `connection.presence`, and `backend.presence`, respectively. Having `.presence` be totally different for each thing may be confusing. It may be better to use `doc.docPresence`, `agent.agentPresence`, etc.

curran:
The only remaining task I'd like to do here is encapsulate the `Backend` presence logic as well.
Then I think this PR will be done.

curran:
Kindly requesting review @ericyhwang @nateps @alecgibson @gkubisa . Thanks in advance!

curran:
This is a good summary of the API surface area of `presence`: https://github.com/share/sharedb/pull/288/files#diff-d6dd14a85690c9e7aacb81836026efca .

curran:
Notes from today's call:

 * Probably should be possible to generalize "hooks" that get called out to presence.

curran:
Notes from today's call:
 * Good idea to replace method calls from doc.js into presence with events instead.
 * Would be good to have multiple presence instances per doc.
 * Arbitrary number of presence implementations for a single OT type.

WinstonFassett:
I recently found this PR and tried porting it into a separate module I called `sharedb-presence` in order to avoid extending ShareDB and the OT types.  Surprisingly, and despite the fact that I don't understand everything that's going on in it, it worked.  So I thought I'd share in case it is useful to anyone.  

It's a little hacky extending the ShareDB backend and client to support a new message type, but seems it can work without modifying the core ShareDB lib.  Still I found I needed to extend `Backend`, `Connection`, and `Agent`.  I didn't see a good way to extend `Doc` such that ShareDB would actually use the extended class, so instead I mix the presence-related stuff into the `Doc` instance returned by `Connection.get`.

I also pulled the presence logic out of `json0` and into a file called `transform-presence.js` that is included.

I tried to avoid extending the `Doc` class very much, so I may have left out something important that might cause some quirky behavior, but otherwise it's the same implementation using `stateless.js`.

I confess I didn't do anything with the tests in this PR because I was just fiddling around.  

Instead, to try it out, I forked json0-presence-demo and got it working:
https://github.com/WinstonFassett/json0-presence-demo

And with a few changes I got the **demo** running on glitch here:
https://winstonfassett-json0-presence-demo.glitch.me/

Hope this is helpful for someone!  Thanks for ShareDB and this PR!

curran:
Very nice! This is awesome to see. Thank you for sharing. Looking forward to digging into it.

Having presence as a separate package may ultimately make the most sense.

curran:
Action items for this PR:

 * [ ] Migrate to events where possible (e.g. hooks in Doc.js). Need good names.
 * [ ] (possibly) Split out StatelessPresence into separate package.

alecgibson:
I've only just got around to playing with the demo. It looks like this implementation potentially suffers from a similar issue to our own: there's sometimes a bit of a race between the presence updates, and the document auto-updating presence, which leads to jumpy cursor behaviour. This is especially jarring when typing at the end of a paragraph.

Oct-22-2019 15-02-20

README.md
```diff
@@ -91,6 +108,8 @@ __Options__
 * `options.pubsub` _(instance of `ShareDB.PubSub`)_
   Notify other ShareDB processes when data changes
   through this pub/sub adapter. Defaults to `ShareDB.MemoryPubSub()`.
+* `options.presence` _(implementation of presence classes)_
+  Enable user presence synchronization. The value of `options.presence` option is expected to contain implementations of the classes `DocPresence`, `ConnectionPresence`, `AgentPresence`, and `BackendPresence`. Logic related to presence is encapsulated within these classes, so it is possible develop additional third party presence implementations external to ShareDB.
```
alecgibson: I'm not sure about this API. It feels atypical both of normal JS, and also of this library. I think I'd rather pass in a single `presence` object which exposes API methods, much like the DB adapters.

README.md
```diff
@@ -308,6 +327,9 @@ Unique document ID
 `doc.data` _(Object)_
 Document contents. Available after document is fetched or subscribed to.
 
+`doc.presence` _(Object)_
+Each property under `doc.presence` contains presence data shared by a client subscribed to this document. The property name is an empty string for this client's data and connection IDs for other clients' data. The structure of the presence object is defined by the OT type of the document (for example, in [ot-rich-text](https://github.com/Teamwork/ot-rich-text#presence) and [@datavis-tech/json0](https://github.com/datavis-tech/json0#presence)).
```
alecgibson: I'm not sure I agree with handing this off to the types, because:

- I'm not sure it's strictly part of "OT", and it feels like a violation of Single Responsibility
- This relies on us being able to update the type libraries (can we?)
- There are potentially multiple desired implementations for a given type. Especially if you consider - for example - `json0`, which can sub-type other arbitrary types. What is a "generic" shape for this presence?

To me, it feels more natural for us to define some sort of separate presence transformers, probably being passed in to the `presence` constructor option?

README.md
```diff
@@ -467,6 +497,10 @@ Additional fields may be added to the error object for debugging context dependi
 * 4022 - Database adapter does not support queries
 * 4023 - Cannot project snapshots of this type
 * 4024 - Invalid version
+* 4025 - Passing options to subscribe has not been implemented
```
alecgibson: (NB: We'll need to change these to string error codes: https://github.com/share/sharedb/pull/309)

lib/backend.js
```diff
@@ -155,6 +166,13 @@ Backend.prototype.connect = function(connection, req) {
   // not used internal to ShareDB, but it is handy for server-side only user
   // code that may cache state on the agent and read it in middleware
   connection.agent = agent;
+
+  // Expose the DocPresence passed in through the constructor
+  // to the Doc class, which has access to the connection.
+  connection.DocPresence = this.presence.DocPresence;
```
alecgibson: This will only work when the client and the server are on the same box. It's not going to work for the mainline case, where the client resides on a remote machine, and we can't directly access `Doc` from `Backend`.

lib/backend.js
```diff
@@ -155,6 +166,13 @@ Backend.prototype.connect = function(connection, req) {
...
+
+  connection._connectionPresence = new this.presence.ConnectionPresence(connection);
```
alecgibson: As I stated above, I don't think this looks "normal". I'd possibly expect something more along the lines of:

```javascript
presence = this.presence.connectionPresence(connection);
```

lib/client/connection.js
```diff
@@ -255,6 +253,9 @@ Connection.prototype.handleMessage = function(message) {
       return;
 
     default:
+      if (this._connectionPresence.isPresenceMessage(message)) {
```
alecgibson: I don't like leaning on `default` for this - why can't we give it its own message type?

lib/client/doc.js
```diff
@@ -58,6 +68,8 @@ function Doc(connection, collection, id) {
   this.type = null;
   this.data = undefined;
 
+  this._docPresence = new connection.DocPresence(this);
```
alecgibson: As I've mentioned above, `Doc` won't have access to `connection.DocPresence` the way this is currently put together, because `connection.DocPresence` is set by `Backend`, which might be on a different machine altogether.

lib/backend.js
```diff
@@ -48,6 +55,10 @@ function Backend(options) {
   if (!options.disableSpaceDelimitedActions) {
     this._shimAfterSubmit();
   }
+
+  this.presence = options.presence || dummyPresence;
```
alecgibson: What's the reason for having this off by default? Performance?

lib/presence/index.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
alecgibson: What's the point of this file?

lib/presence/stateless.js
```diff
@@ -0,0 +1,539 @@
...
+
+  // The presence objects received from the server.
+  // Map of src -> presence
+  this.received = {};
+
+  // The minimum amount of time to wait before removing processed presence from this.presence.received.
+  // The processed presence is removed to avoid leaking memory, in case peers keep connecting and disconnecting a lot.
+  // The processed presence is not removed immediately to enable avoiding race conditions, where messages with lower
+  // sequence number arrive after messages with higher sequence numbers.
+  this.receivedTimeout = 60000;
```
alecgibson: Can we get some justification about why a minute is a good default amount of time? (ie is it because it's a long time compared to the "standard" lifecycle of a submit request?). How does this work if we're on a patchy connection?

Also nit: can we please write `60 * 1000`?

alecgibson:
Okay, I've tried to actually use this, but it doesn't currently work if your client isn't on the same box as your server (as outlined inline). I've also had a quick skim through the code, and here are my personal thoughts:

I like that this attempts to stay as generic as possible, but I worry that it's gone too far. I actually think that `DocPresence` should be its own class, beside `Doc` and `Query`, and that all the other logic (`AgentPresence`, `BackendPresence` and `ConnectionPresence`) should just be baked into the core. Let's be pragmatic about this: presence is non-trivial, and I'm not sure we'll have many full rewrites of _all_ of the presence logic: more just tweaks to its performance.

I think I've said in the past that I'd expected presence to act more like its own entity, and that you should be able to just do things like `connection.getPresence(collection, id)` in a similar way to `connection.get(collection, id)` for a `Doc`. Obviously, behind the scenes, the `Presence` object would be tightly coupled to `Doc`, but I think it's still conceptually its own thing.

In the future, if we wish to add pluggable behaviour, I think we should expose presence middleware hooks. For example, if you wanted to keep a stateful cache of presence on the server, then perhaps there would be a `requestPresence` hook or similar, which by default no-ops, and then requests presence from all connected clients. However, if you inject middleware into the hook, then it can query a cache, and fast return from there, rather than request from clients. This approach would both be architecturally simpler, and more similar to how ShareDB already operates.

An alternative to middleware might be that we have a PresenceCacheDB, which again by default no-ops, but can be replaced by an instance that actually serves cached presence. But - in my opinion - this is all stuff that can be added after implementing the basics (this was very much how we approached document history and milestones).

I'm also not a fan of delegating presence functionality to types. I don't think it should be their responsibility, and I think there might be many "shapes" for a presence object for a given type, depending on application. I personally think we should configure presence transformers in the `Backend` constructor, a bit like:

```javascript
new Backend({
  presence: {
    cacheDb: new MemoryPresenceCacheDB(),
    transformers: {
      'rich-text': (presence, delta, isOwnOp) => { ... },
    },
  },
});
```

curran:
All excellent ideas and feedback @alecgibson !

/cc @gkubisa , the original author of this work. All I did with this PR was to try to rearrange his work such that the PR could be accepted.

I do love the idea that presence should be a different responsibility than the OT types. It was a pain to [fork JSON0](https://github.com/datavis-tech/json0) and add presence to it.

alecgibson:
Okay, I might try to have a stab at this today with a blank piece of paper, but drawing heavily on what we've done here.

curran:
That would be amazing! Happy to review your work.

If you haven't seen it, the presence feature in https://github.com/SyncOT/SyncOT might be interesting.

gkubisa:
Just to give a bit of background, SyncOT is my take on OT with somewhat different trade-offs in some key areas. As I could design it from the ground up, I came up with a bit different way of handling presence, which you might want to adapt to ShareDB. Note that SyncOT is not complete and the ideas presented below have not been tested in real-life situations, apart from Global Presence which we're currently using in production as a stand-alone component separate from OT.

Back to presence in the context of OT... the main idea is to decouple Global Presence and Local Presence:

**Global Presence** represents user sessions/connections. It contains 3 mandatory properties: `sessionId`, `userId` and `locationId`, which could be the ID of an OT document, location within the app, etc. It also has the `lastModified` timestamp and optional custom data. Global Presence can be synchronized between clients based on `sessionId`, `userId` or `locationId`, where the last one is by far the most useful. It is cached and shared using Redis.

**Local Presence** represents users' presence within OT documents, eg caret positions, etc. It is stored as part of OT documents, updated through operations and transformed as needed. It is associated with Global Presence using `sessionId`. Having Local Presence as part of operations and snapshots greatly simplifies the whole system, as it reuses storage/caching, synchronization, conflict resolution and undo/redo.

From the above you could notice that Global Presence is transient and Local Presence is persistent, which leads to some problems:

- text selections of offline users could be visible. The solution is to simply skip displaying selections for sessions which are not active in Global Presence.
- latency/CPU/bandwidth overhead of storing presence events, eg in MongoDB. This can be minimized by throttling presence updates and optimizing storage - the current MongoDB driver in ShareDB: reads the snapshot, applies an operation, writes the operation and writes the snapshot. In SyncOT the usual process will be: apply the operation to the snapshot cashed in the local memory and write the operation to the database.
- storage overhead. This would be minimized by [MongoDB compression](https://docs.mongodb.com/manual/core/wiredtiger/#compression). Additionally, if necessary, I will implement a utility which can remove obsolete Local Presence from the database. It will be simple and safe, as Local Presence will be stored in a separate operation property as a presence snapshot - I mean for example, "setCursorPosition X", rather than "moveCursorBy X".

I hope the above makes sense and might help implement presence in ShareDB.

curran:
FWIW I'm in favor of the work over in https://github.com/share/sharedb/pull/322 . It looks a lot cleaner than this implementation at first glance.

curran:
Closing this in favor of https://github.com/share/sharedb/pull/322

PR #301, opened 2019-07-04
Add `Backend` documentation

alecgibson:
This change adds `Backend` documentation to methods that provide
functionality not offered by a client.

### DIFF
--- a/README.md
+++ b/README.md
@@ -321,6 +321,9 @@ Populate the fields on `doc` with a snapshot of the document from the server.
 Populate the fields on `doc` with a snapshot of the document from the server, and
 fire events on subsequent changes.
 
+`doc.unsubscribe(function (err) {...})`	
+Stop listening for document updates. The document data at the time of unsubscribing remains in memory, but no longer stays up-to-date. Resubscribe with `doc.subscribe`.
+
 `doc.ingestSnapshot(snapshot, callback)`
 Ingest snapshot data. The `snapshot` param must include the fields `v` (doc version), `data`, and `type` (OT type). This method is generally called interally as a result of fetch or subscribe and not directly from user code. However, it may still be called directly from user code to pass data that was transferred to the client external to the client's ShareDB connection, such as snapshot data sent along with server rendering of a webpage.
 
--- a/README.md
+++ b/README.md
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
 `query.on('extra', function() {...}))`
 (Only fires on subscription queries) `query.extra` changed.
 
+### Class: `Backend`
+
+#### `connect`
+
+```javascript
+backend.connect(connection: Connection, req: Object): Connection;
+```
+
+Connects to ShareDB and returns an instance of a [`Connection`](#class-sharedbconnection)
+
+* `connection` _Connection (optional)_: a [`Connection`](#class-sharedbconnection) instance to bind to the `Backend`
+* `req` _Object (optional)_: a connection context object that can contain information such as cookies or session data that will be available in the [middleware](#middlewares)
+
+Returns a [`Connection`](#class-sharedbconnection).
+
+#### `listen`
+
+```javascript
+backend.listen(stream: Stream, req: Object): Agent;
+```
+
+Registers a `Stream` with the backend.
+
+* `stream` _Stream_: a [`Stream`](https://nodejs.org/api/stream.html) (or `Stream`-like object) that will be used to communicate between the `Agent` and the `Backend`
+* `req` _Object (optional)_: a connection context object that can contain information such as cookies or session data that will be available in the [middleware](#middlewares)
+
+Returns an `Agent`, which is also available in the [middleware](#middlewares).
+
+#### `close`
+
+```javascript
+backend.close(callback: Function): void;
+```
+
+Disconnects ShareDB and all of its underlying services (database, pubsub, etc.).
+
+* `callback` _Function_: a callback with the signature `function (error: Error): void` that will be called once the services have stopped, or with an `error` if at least one of them could not be stopped
+
+#### `use`
+
+```javascript
+backend.use(action: string | string[], middleware: Function): Backend;
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+```
+
+Submits an operation to the `Backend`.
+
+* `agent` _Agent_: connection agent to pass to the middleware
+* `index` _string_: the name of the target collection or projection
+* `id` _string_: the document ID
+* `op` _Object_: the operation to submit
+<!-- TODO: What does this options object do? -->
+* `options` _Object_: ???
```
alecgibson: What does this `options` argument do?

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+<!-- TODO: Check that I understand this properly -->
+* `callback` _Function_: a callback with the signature `function (error: Error, ops: Object[]): void;`, where `ops` are the ops that were actually submitted to ShareDB after transforming them up to work on the latest version of the document (which may be different to the op that was originally submitted)
```
alecgibson: Is this description of the callback correct?

alecgibson: Maybe we should just remove this entire `submit` entry? Can we do anything here that we can't do from a client?

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+
+Fetch multiple document snapshots from a collection.
+
+* `agent` _Agent_: connection agent to pass to the middleware
+* `index` _string_: the name of the target collection or projection
+* `ids` _string[]_: array of document IDs
+* `options`: _Object (optional)_: options can be passed directly to the database driver's `fetchBulk` inside the `snapshotOptions` property: `{snapshotOptions: {metadata: true}}`
+* `callback`: _Function_: a callback with the signature `function (error: Error, snapshotMap: Object): void;`, where `snapshotMap` is an object whose keys are the requested IDs, and the values are the requested `Snapshot`s
+
+<!-- TODO: Is it okay to have ignored `subscribe`, `subscribeBulk`, `querySubscribe`? -->
```
alecgibson: I've ignored all subscribe-related functions, because I think they don't really make much sense without a client...?

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+
+<!-- TODO: Should we ignore `queryFetch`? -->
+
+#### `queryFetch`
```
alecgibson: Should we include this? Does it offer any extra functionality over just doing this from a client?

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+
+```javascript
+backend.queryFetch(agent: Agent, index: string, query: Object, options: Object, callback: Function): void;
+```
+
+Fetch snapshots that match the provided query.
```
alecgibson: Mention that we should use the database directly in most cases, unless they want middleware and to avoid the `Doc` overheads.

README.md
alecgibson: 👍

README.md
alecgibson: In theory, these options get passed `db.commit`, so in theory options can be passed through to any database adapter that uses them.

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
 
+### Class: `Backend`
+
+#### `connect`
+
+```javascript
+backend.connect(connection: Connection, req: Object): Connection;
+```
+
+Connects to ShareDB and returns an instance of a [`Connection`](#class-sharedbconnection)
```
ericyhwang:
```suggestion
Connects to ShareDB and returns an instance of a [`Connection`](#class-sharedbconnection). This is the Node equivalent to `new ShareDBClient.Connection(socket)` in the browser.
```

README.md
alecgibson: The ops are actually the ops that were committed between my current document version, and the version in which my own op was committed. These are used to apply to the client to transform the document up, and keep it consistent with the database.

Let's keep the `submit` entry.

README.md
alecgibson: Keep it.

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+```javascript
+backend.queryFetch(agent: Agent, index: string, query: Object, options: Object, callback: Function): void;
+```
+
+Fetch snapshots that match the provided query.
+
+* `agent` _Agent_: connection agent to pass to the middleware
+* `index` _string_: the name of the target collection or projection
+* `query` _Object_: a query object, whose format will depend on the database adapter being used
+<!-- TODO: I don't think options are used? -->
```
alecgibson: These are the same as `.submit` - mostly there for future-proofing, so that options can be sent through to the driver.

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+
+Returns a [`Connection`](#class-sharedbconnection).
+
+#### `listen`
+
+```javascript
+backend.listen(stream: Stream, req: Object): Agent;
+```
+
+Registers a `Stream` with the backend.
```
ericyhwang:
```suggestion
Registers a `Stream` with the backend. This should be called when the server receives a new connection from a client.
```

README.md
```diff
@@ -407,6 +407,165 @@ after a sequence of diffs are handled.
...
+
+* `stream` _Stream_: a [`Stream`](https://nodejs.org/api/stream.html) (or `Stream`-like object) that will be used to communicate between the `Agent` and the `Backend`
```
ericyhwang:
```suggestion
* `stream` _Stream_: a [`Stream`](https://nodejs.org/api/stream.html) (or `Stream`-like object) that will be used to communicate between the new `Agent` and the `Backend`
```

alecgibson:
We should add some stub documentation for `Agent`, mostly to document `connection.agent` to get a handle on an agent, and `agent.custom` for middleware things. We probably don't need to document the methods.

PR #302, opened 2019-07-04
Linter fixes

alecgibson:
This change addresses three things:

  1. Move us from `jshint` to `eslint` (which is the more commonly used
     linter)
  2. Fix our linting glob patterns, which currently only address files
     at the top of their respective directories (and ignore all files in
     subdirectories)
  3. Enforce stricter linting rules, based roughly on the [Google][1]
     rules

Note that the change is split into two commits:

  1. The first one adds `eslint` and fixes most of the issues
  2. The second one only addresses indentation issues in a couple of files, which would have made the change very noisy

[1]: https://github.com/google/eslint-config-google


lib/backend.js
```diff
@@ -95,7 +94,10 @@ Backend.prototype.SNAPSHOT_TYPES = {
 Backend.prototype._shimDocAction = function() {
   if (warnDeprecatedDoc) {
     warnDeprecatedDoc = false;
-    console.warn('DEPRECATED: "doc" middleware action. Use "readSnapshots" instead. Pass `disableDocAction: true` option to ShareDB to disable the "doc" action and this warning.');
+    console.warn([
+      'DEPRECATED: "doc" middleware action. Use "readSnapshots" instead.',
+      'Pass `disableDocAction: true` option to ShareDB to disable the "doc" action and this warning.'
+    ].join(' '));
```
ericyhwang: To reduce memory allocation overhead from string joining, I'd prefer this to either:
* Remain as one long inline string
* Or do the join once at module level, instead of inside the function

Same thing below for `_shimAfterSubmit`.

alecgibson: Let's move this to the top of the file

```js
"with this " +
"style"
```

lib/agent.js
```diff
@@ -333,7 +332,7 @@ Agent.prototype._handleMessage = function(request, callback) {
 };
 function getQueryOptions(request) {
   var results = request.r;
-  var ids, fetch, fetchOps;
+  var ids; var fetch; var fetchOps;
```
alecgibson: Let's force multiple lines here.

.eslintrc
```diff
@@ -0,0 +1,139 @@
+{
```
alecgibson: Let's see if we can `extend` Google and ignore ES6 style rules.

nateps: Can we figure out how to reference the google style and just override the explicit changes?

lib/agent.js
nateps: let's just use multiple lines for each statement

Issue #287, opened 2019-04-17
[1.0.0] Switch error codes from numbers to strings


ericyhwang:
Per @nateps on https://github.com/share/sharedb/issues/198:

> Fixing the error codes is something I'd like to do for a 1.0 release, and this should be a relatively quick update.
> 
> As of Node.js 8.0, Node uses the convention of `Error.code` being a string, which in retrospect makes more sense than a number. I'd like to update both the codes in ShareDB and the README to be a unique string in a similar convention to the one used by Node.js: https://nodejs.org/api/errors.html#nodejs-error-codes
> 
> I was planning to make this change, but if you'd like to suggest some codes, I'd appreciate the help!

We'll also want to document that this is a breaking change moving from 1.0.0-beta to 1.0.0.

alecgibson:
Do we have a checklist for things we need to do prior to an actual v1 release? Can we make sure we remove old middleware shims, too?

Issue #198, opened 2018-03-01
Error code documentation is out-of-date


alecgibson:
The error codes documented at the end of the `README` are out-of-date. For example, according to the [`README`][1], error `4002` means "Database adapter does not support subscribe". However, according to the [source code][2], error `4002` means "Op submit rejected".

I'm happy to raise a pull request to delete these from the `README` to at least reduce confusion, but probably don't have the time to rummage through the code to update the docs.

[1]: https://github.com/share/sharedb#4000---bad-request
[2]: https://github.com/share/sharedb/blob/ea5053b8064876538c61a2284c4f931fa8b52c79/lib/client/doc.js#L273

nateps:
Thanks for flagging this!

Fixing the error codes is something I'd like to do for a 1.0 release, and this should be a relatively quick update.

As of Node.js 8.0, Node uses the convention of Error.code being a string, which in retrospect makes more sense than a number. I'd like to update both the codes in ShareDB and the README to be a unique string in a similar convention to the one used by Node.js: https://nodejs.org/api/errors.html#nodejs-error-codes

I was planning to make this change, but if you'd like to suggest some codes, I'd appreciate the help!

PR #309, opened 2019-09-13
Replace numeric error codes with strings

alecgibson:
Fixes https://github.com/share/sharedb/issues/287
Fixes https://github.com/share/sharedb/issues/198

This is a breaking change that moves us from using numeric error codes
to string-based codes [similar to Node][1].

The motivation for this is:

  - more descriptive error codes
  - stop implying that errors can be broadly classified together (in the
    same way that HTTP errors might)

This change also exposes these codes on `ShareDBError.code`, so that
they are easily discoverable, and can be referenced by consumers for
programmatic error handling.

[1]: https://nodejs.org/api/errors.html#nodejs-error-codes


alecgibson:
I've done this PR in a relatively "dumb" way, just blindly swapping out numeric codes for strings as I saw fit. While we're here, I think it would be good to think about:

  - **specificity**: some of our errors are very specific, and some are very broad; do we want/need this to be consistent? I usually think that the message component can be used to add any extra context, but there may be cases where an error that sounds generic may want to be handled in a different way programmatically, where it would want a different error code.
    - I've notably **reduced** specificity for all the "not implemented" messages, because having different codes felt redundant. If this is wrong, let me know.
    - Are there other places to reduce specificity? What about:
      - `CREATE_DATA_MUST_BE_AN_OBJECT`
      - `DEL_MUST_BE_TRUE`
      - `DOCUMENT_ALREADY_CREATED` vs `DOCUMENT_CREATED_REMOTELY`
      - `SRC_AND_SEQ_MUST_BE_SET_TOGETHER`, `INVALID_SRC`, `INVALID_SEQ`
      - others?
    - we may need to **increase** specificity on the `INVALID_VERSION` error? We use this to refer to the doc/op/snapshot version. Do we need different errors for each of these? If not, is there a better way to name this code (to differentiate it from eg `INVALID_PROTOCOL_VERSION`)
  - **naming**: this is the hard part!
    - All the constant keys are missing the `ERR_` prefix to make consumer code a bit neater and less overloaded than `ShareDBError.code.ERR_...` (it's clear it's an error code already, I think). Is this okay?
    - Do we want a policy on frontloading? For example, should we try to always have `<subject>_<state>`? So `CLIENT_ID_INVALID` rather than `INVALID_CLIENT_ID`? I personally think that this format helps to remove some noise when looking at the error codes; helps to prevent duplicate codes; makes it a bit clearer what can be wrong with a particular thing. I thought I'd wait for opinions before actually doing this
  - **error code overlaps** - some of the error codes were overlapping (notably the `4XXX` series). I didn't do a find-replace, so there shouldn't be any breakages, but you can never be too careful
  - **default code**: I think the only other slightly controversial thing I've done is add a default code to the `ShareDBError` class. It seemed like a sensible thing to do.

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
 function ShareDBError(code, message) {
   ShareDBError.super.call(this, message);
-  this.code = code;
+  this.code = code || ShareDBError.code.UNKNOWN_ERROR;
 }
 
 makeError(ShareDBError);
 
+ShareDBError.code = {
+  BADLY_FORMED_MESSAGE: 'ERR_BADLY_FORMED_MESSAGE',
```
alecgibson: Let's prefix the keys with `ERR_`

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
 
 function ShareDBError(code, message) {
   ShareDBError.super.call(this, message);
-  this.code = code;
+  this.code = code || ShareDBError.code.UNKNOWN_ERROR;
```
alecgibson: Let's make `this.code` all caps `this.CODES`, because it's a constant, and it helps consumers to understand it's an object.

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
 
 makeError(ShareDBError);
 
+ShareDBError.code = {
+  BADLY_FORMED_MESSAGE: 'ERR_BADLY_FORMED_MESSAGE',
+  CANNOT_PROJECT_TYPE: 'ERR_CANNOT_PROJECT_TYPE',
+  CREATE_DATA_MUST_BE_AN_OBJECT: 'ERR_CREATE_DATA_MUST_BE_AN_OBJECT',
+  DATABASE_ADAPTER_NOT_FOUND: 'ERR_DATABASE_ADAPTER_NOT_FOUND',
+  DEL_MUST_BE_TRUE: 'ERR_DEL_MUST_BE_TRUE',
+  DOCUMENT_ALREADY_CREATED: 'ERR_DOCUMENT_ALREADY_CREATED',
```
alecgibson: Let's keep this, and use it in place of `DOCUMENT_CREATED_REMOTELY`, also.

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
+  DOCUMENT_CREATED_REMOTELY: 'ERR_DOCUMENT_CREATED_REMOTELY',
+  DOCUMENT_DOES_NOT_EXIST: 'ERR_DOCUMENT_DOES_NOT_EXIST',
+  DOCUMENT_WAS_DELETED: 'ERR_DOCUMENT_WAS_DELETED',
+  INVALID_CLIENT_ID: 'ERR_INVALID_CLIENT_ID',
+  INVALID_DEFAULT_TYPE: 'ERR_INVALID_DEFAULT_TYPE',
+  INVALID_METADATA: 'ERR_INVALID_METADATA',
+  INVALID_SRC: 'ERR_INVALID_SRC',
```
alecgibson: Let's group `INVALID_SRC`, `INVALID_SEQ`, and `SRC_AND_SEQ...` as `INVALID_OP_FORMAT`, and add more detail as the message.

alecgibson:
As a rough rule of thumb, let's group object formatting issues under single codes, but for any user input (eg `DEL_MUST_BE_TRUE`), let's have specific error codes.

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
+  INVALID_SEQ: 'ERR_INVALID_SEQ',
+  INVALID_STATE_TRANSITION: 'ERR_INVALID_STATE_TRANSITION',
+  INVALID_PROTOCOL_VERSION: 'ERR_INVALID_PROTOCOL_VERSION',
+  INVALID_VERSION: 'ERR_INVALID_VERSION',
+  LISTENER_NOT_ASSIGNED: 'ERR_LISTENER_NOT_ASSIGNED',
+  MAX_RETRIES: 'ERR_MAX_RETRIES',
+  MISSING_CREATE_TYPE: 'ERR_MISSING_CREATE_TYPE',
+  MISSING_OP: 'ERR_MISSING_OP',
```
alecgibson: Let's umbrella this under `INVALID_OP_FORMAT`

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
+  CREATE_DATA_MUST_BE_AN_OBJECT: 'ERR_CREATE_DATA_MUST_BE_AN_OBJECT',
+  DATABASE_ADAPTER_NOT_FOUND: 'ERR_DATABASE_ADAPTER_NOT_FOUND',
+  DEL_MUST_BE_TRUE: 'ERR_DEL_MUST_BE_TRUE',
+  DOCUMENT_ALREADY_CREATED: 'ERR_DOCUMENT_ALREADY_CREATED',
+  DOCUMENT_CREATED_REMOTELY: 'ERR_DOCUMENT_CREATED_REMOTELY',
+  DOCUMENT_DOES_NOT_EXIST: 'ERR_DOCUMENT_DOES_NOT_EXIST',
+  DOCUMENT_WAS_DELETED: 'ERR_DOCUMENT_WAS_DELETED',
+  INVALID_CLIENT_ID: 'ERR_INVALID_CLIENT_ID',
+  INVALID_DEFAULT_TYPE: 'ERR_INVALID_DEFAULT_TYPE',
+  INVALID_METADATA: 'ERR_INVALID_METADATA',
```
alecgibson: -> `INVALID_OP_METADATA`

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
+  NOT_IMPLEMENTED: 'ERR_NOT_IMPLEMENTED',
+  OP_ALREADY_SUBMITTED: 'ERR_OP_ALREADY_SUBMITTED',
+  OP_INVALID_IN_PROJECTION: 'ERR_OP_INVALID_IN_PROJECTION',
+  OP_SUBMIT_REJECTED: 'ERR_OP_SUBMIT_REJECTED',
+  OP_VERSION_NEWER_THAN_SNAPSHOT: 'ERR_OP_VERSION_NEWER_THAN_SNAPSHOT',
+  SRC_AND_SEQ_MUST_BE_SET_TOGETHER: 'ERR_SRC_AND_SEQ_MUST_BE_SET_TOGETHER',
+  SUBSCRIBE_UNSUPPORTED: 'ERR_SUBSCRIBE_UNSUPPORTED',
+  UNKNOWN_ERROR: 'ERR_UNKNOWN_ERROR',
+  UNKNOWN_TYPE: 'ERR_UNKNOWN_TYPE',
+  VERSION_MISMATCH: 'ERR_VERSION_MISMATCH'
```
alecgibson: -> `OP_SNAPSHOT_VERSION_MISMATCH`?

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
+  INVALID_VERSION: 'ERR_INVALID_VERSION',
+  LISTENER_NOT_ASSIGNED: 'ERR_LISTENER_NOT_ASSIGNED',
+  MAX_RETRIES: 'ERR_MAX_RETRIES',
+  MISSING_CREATE_TYPE: 'ERR_MISSING_CREATE_TYPE',
+  MISSING_OP: 'ERR_MISSING_OP',
+  MISSING_TRANSFORM_OPS: 'ERR_MISSING_TRANSFORM_OPS',
+  MISSING_SNAPSHOT: 'ERR_MISSING_SNAPSHOT',
+  MISSING_SNAPSHOT_VERSION: 'ERR_MISSING_SNAPSHOT_VERSION',
+  NO_OP_TO_SEND: 'ERR_NO_OP_TO_SEND',
+  NOT_IMPLEMENTED: 'ERR_NOT_IMPLEMENTED',
```
alecgibson: -> `DATABASE_METHOD_NOT_IMPLEMENTED`

alecgibson:
Let's go ahead with `<subject>_<state>`, but `subject` might be the act of doing something eg `apply`

lib/error.js
alecgibson: Also all the different version mismatches should be different error codes, to help you narrow down where the error came from.

lib/submit-request.js
```diff
@@ -240,34 +243,43 @@ SubmitRequest.prototype._shouldSaveMilestoneSnapshot = function(snapshot) {
...
 SubmitRequest.prototype.missingOpsError = function() {
   return {
-    code: 5001,
+    code: ERROR_CODE.MISSING_TRANSFORM_OPS,
     message: 'Op submit failed. DB missing ops needed to transform it up to the current snapshot version'
   };
 };
 SubmitRequest.prototype.versionDuringTransformError = function() {
-  return {code: 5002, message: 'Op submit failed. Versions mismatched during op transform'};
+  return {code: ERROR_CODE.VERSION_MISMATCH, message: 'Op submit failed. Versions mismatched during op transform'};
```
alecgibson: Maybe name this and the one below after the methods? (`during` and `after`)

lib/error.js
```diff
@@ -2,9 +2,47 @@ var makeError = require('make-error');
...
+  DOCUMENT_DOES_NOT_EXIST: 'ERR_DOCUMENT_DOES_NOT_EXIST',
+  DOCUMENT_WAS_DELETED: 'ERR_DOCUMENT_WAS_DELETED',
+  INVALID_CLIENT_ID: 'ERR_INVALID_CLIENT_ID',
+  INVALID_DEFAULT_TYPE: 'ERR_INVALID_DEFAULT_TYPE',
+  INVALID_METADATA: 'ERR_INVALID_METADATA',
+  INVALID_SRC: 'ERR_INVALID_SRC',
+  INVALID_SEQ: 'ERR_INVALID_SEQ',
+  INVALID_STATE_TRANSITION: 'ERR_INVALID_STATE_TRANSITION',
+  INVALID_PROTOCOL_VERSION: 'ERR_INVALID_PROTOCOL_VERSION',
+  INVALID_VERSION: 'ERR_INVALID_VERSION',
```
alecgibson: Let's only use the word `version` in error codes when provided with context.

lib/error.js
```diff
@@ -2,9 +2,42 @@ var makeError = require('make-error');
...
+  ERR_CLIENT_ID_BADLY_FORMED: 'ERR_CLIENT_ID_BADLY_FORMED',
+  ERR_CONNECTION_STATE_TRANSITION_INVALID: 'ERR_CONNECTION_STATE_TRANSITION_INVALID',
+  ERR_DATABASE_ADAPTER_NOT_FOUND: 'ERR_DATABASE_ADAPTER_NOT_FOUND',
+  ERR_DATABASE_DOES_NOT_SUPPORT_SUBSCRIBE: 'ERR_DATABASE_DOES_NOT_SUPPORT_SUBSCRIBE',
+  ERR_DATABASE_METHOD_NOT_IMPLEMENTED: 'ERR_DATABASE_METHOD_NOT_IMPLEMENTED',
+  ERR_DEFAULT_TYPE_MISMATCH: 'ERR_DEFAULT_TYPE_MISMATCH',
+  ERR_DOC_MISSING_VERSION: 'ERR_DOC_MISSING_VERSION',
+  ERR_DOC_ALREADY_CREATED: 'ERR_DOC_ALREADY_CREATED',
+  ERR_DOC_DOES_NOT_EXIST: 'ERR_DOC_DOES_NOT_EXIST',
+  ERR_DOC_TYPE_NOT_RECOGNISED: 'ERR_DOC_TYPE_NOT_RECOGNISED',
```
nateps: Let's use `en-us` spelling (recognized)

lib/milestone-db/memory.js
```diff
@@ -23,7 +25,9 @@ function MemoryMilestoneDB(options) {
 MemoryMilestoneDB.prototype = Object.create(MilestoneDB.prototype);
 
 MemoryMilestoneDB.prototype.getMilestoneSnapshot = function(collection, id, version, callback) {
-  if (!this._isValidVersion(version)) return process.nextTick(callback, new ShareDBError(4001, 'Invalid version'));
+  if (!this._isValidVersion(version)) {
+    return process.nextTick(callback, new ShareDBError(ERROR_CODE.ERR_UNKNOWN_ERROR, 'Invalid version'));
```
alecgibson: We shouldn't use unknown errors explicitly - let's use something like `ERR_MILESTONE_ARGUMENT_INVALID`

lib/ot.js
```diff
@@ -4,46 +4,49 @@
...
     }
     var typeName = op.create.type;
     if (typeof typeName !== 'string') {
-      return {code: 4007, message: 'Missing create type'};
+      return {code: ERROR_CODE.ERR_OT_OP_BADLY_FORMED, message: 'Missing create type'};
     }
     var type = types[typeName];
     if (type == null || typeof type !== 'object') {
-      return {code: 4008, message: 'Unknown type'};
+      return {code: ERROR_CODE.ERR_DOC_TYPE_NOT_RECOGNISED, message: 'Unknown type'};
```
alecgibson: Let's change this to `OP_BADLY_FORMED` to fit in with everything else in this  function.

alecgibson: Actually let's leave this as-is.

alecgibson:
While we're here, let's also replace all of our anonymous error objects with instances of an actual `Error`. We should create our own `Error` class that extends the base, but with its own type for `instanceof` checking. We can base this loosely on the [MDN example](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error).

In short we should:
- set the `name` property
- try to use `captureStackTrace` if available
- otherwise just construct a new `Error` and copy the `stack` from that

PR #316, opened 2019-10-18
Support query subscriptions on projections with a filter not in the projection fields

ericyhwang:
Prior to this PR, if you issue a query subscription on a projection, with a filter field that's not in the projection fields, then an op to change that filter field will not trigger the query to update.

This is because the op projection (`_sanitizeOp`) was happening prior to the `skipPoll` checks, which are what determine if an op could affect a query. A projected op would appear like it doesn't affect the query.

The fix is to defer applying projections to query ops until after the `skipPoll` checks.

I couldn't write a unit test in this repo, since the MemoryDB's query implementation always returns all docs in the collection, regardless of query filters. See below for the test I'll be adding to sharedb-mongo to cover this case.

----

Test I'll be adding to sharedb-mongo, which fails prior to this change and succeeds after this change:

```js
  describe('projection query subscribe', function() {
    var ShareDB = require('sharedb');

    var backend;
    beforeEach(function() {
      backend = new ShareDB({db: this.db});
    });

    it('query subscribe on projection will update based on fields not in projection', function(done) {
      backend.addProjection('notes_text', 'notes', {text: true});
      var connection1 = backend.connect();
      var connection2 = backend.connect();
      // Create doc and a query that doesn't match the doc.
      var doc1 = connection1.get('notes', 'doc1');
      doc1.create({color: 'blue', text: 'Hello', location: 'Work'}, null, null, function(err) {
        if (err) return done(err);
        var query = connection2.createSubscribeQuery('notes_text', {color: 'green'}, null, function(err, results) {
          if (err) return done(err);
          expect(results).to.have.length(0);

          // Submit an op on a field not in the projection, where the op should cause the doc to be
          // included in the query results.
          doc1.submitOp({p: ['color'], od: 'blue', oi: 'green'}, null, function(err) {
            if (err) return done(err);
            setTimeout(function() {
              expect(query.results).to.have.length(1);
              expect(query.results[0].id).to.equal('doc1');
              expect(query.results[0].data).to.eql({text: 'Hello'});
              done();
            }, 10);
          });
        });
      });
    });
  });
});
```

### DIFF
--- a/lib/op-stream.js
+++ b/lib/op-stream.js
@@ -22,6 +22,7 @@ OpStream.prototype._read = util.doNothing;
 OpStream.prototype.pushData = function(data) {
   // Ignore any messages after unsubscribe
   if (!this.open) return;
+  // This data gets consumed in Agent#_subscribeToStream
   this.push(data);
 };
 
--- a/lib/op-stream.js
+++ b/lib/op-stream.js
@@ -30,18 +30,10 @@ OpStream.prototype.initProjection = function(backend, agent, projection) {
   this.projection = projection;
 };
 
-OpStream.prototype.pushOp = function(collection, id, op) {
-  if (this.backend) {
-    var stream = this;
-    this.backend._sanitizeOp(this.agent, this.projection, collection, id, op, function(err) {
-      if (!stream.open) return;
-      stream.push(err ? {error: err} : op);
-    });
-  } else {
-    // Ignore any messages after unsubscribe
-    if (!this.open) return;
-    this.push(op);
-  }
+OpStream.prototype.pushOp = function(_collection, _id, op) {
+  // Ignore any messages after unsubscribe
+  if (!this.open) return;
+  this.push(op);
 };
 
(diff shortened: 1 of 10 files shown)


aconrad:
Thanks for your work on this @ericyhwang!

I wonder if https://github.com/share/sharedb/issues/199 is affected by this change and whether your PR would satisfy their concern.

lib/op-stream.js
```diff
@@ -24,33 +19,14 @@ inherits(OpStream, Readable);
...
-};
-
-OpStream.prototype.pushOps = function(collection, id, ops) {
-  for (var i = 0; i < ops.length; i++) {
-    this.pushOp(collection, id, ops[i]);
-  }
+OpStream.prototype.pushData = function(data) {
+  // Ignore any messages after unsubscribe
+  if (!this.open) return;
+  this.push(data);
```
ericyhwang:
```suggestion
  // This data gets consumed in Agent#_subscribeToStream
  this.push(data);
```

^ To make it a bit easier to follow the flow of execution

PR #318, opened 2019-10-25
Update dependencies & node versions in travis

nateps:
Update dependencies to current versions:
* deep-is => fast-deep-equal, since deep-is has some correctness bugs, is much slower, and isn't well maintained
* expect.js => chai for assertions
* instanbul => nyc for generating coverage reports
* Update all dev-dependency versions to current
* Add Node 12 & remove Node 6 from travis configuration to bring current with Node LTS versions

### DIFF
--- a/test/pubsub-memory.js
+++ b/test/pubsub-memory.js
@@ -1,6 +1,6 @@
 var MemoryPubSub = require('../lib/pubsub/memory');
 var PubSub = require('../lib/pubsub');
-var expect = require('expect.js');
+var expect = require('chai').expect;
 
 require('./pubsub')(function(callback) {
   callback(null, new MemoryPubSub());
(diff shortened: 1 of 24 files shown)


test/pubsub-memory.js
```diff
@@ -13,7 +13,7 @@ describe('PubSub base class', function() {
   it('returns an error if _subscribe is unimplemented', function(done) {
     var pubsub = new PubSub();
     pubsub.subscribe('x', function(err) {
-      expect(err).an(Error);
+      expect(err).ok;
```
ericyhwang: For this file, Chai equivalent of the old `expect(err).an(Error)` would be `expect(err).to.be.an('Error')`.

test/pubsub.js
```diff
@@ -78,7 +78,7 @@ module.exports = function(create) {
 
     it('can emit events', function(done) {
       this.pubsub.on('error', function(err) {
-        expect(err).an(Error);
+        expect(err).ok;
```
ericyhwang: Same here, `expect(err).to.be.an('Error')`

PR #320, opened 2019-10-29
Update dependencies in examples and bring them up to date

nateps:
* Update dependency versions used in server. Old version of `ws` has security vulnerabilities and we don't want people copying the old version
* Use ReconnectingWebSocket in all examples so we demonstrate reconnecting by default
* React 0.15 => 0.16 and update to more current es6 class style components
* Use express consistently in all examples


examples/leaderboard/client/PlayerList.jsx
```diff
@@ -1,30 +1,29 @@
...
 var React = require('react');
 var Player = require('./Player.jsx');
 var _ = require('underscore');
 
-var PlayerList = React.createClass({
-  propTypes: {
-    players: React.PropTypes.array.isRequired,
-    selectedPlayerId: React.PropTypes.string
-  },
+function PlayerList(props) {
```
ericyhwang: Did you intentionally leave this as a function to serve as an example of ES5 syntax, instead of converting it to a `class` like the examples higher up?

Though if it's to serve as an ES5 example, you can't use object destructuring (line 7) and spread syntax (line 14, though that might be ok if React has its own JS expression parser for template strings).

examples/leaderboard/client/PlayerSelector.jsx
```diff
@@ -1,24 +1,23 @@
+var PropTypes = require('prop-types');
 var React = require('react');
 
-var PlayerSelector = React.createClass({
-  propTypes: {
-    selectedPlayer: React.PropTypes.object
-  },
+function PlayerSelector({ selectedPlayer, onAddPoints }) {
```
ericyhwang: Same here, was leaving this a `function` intentional?

examples/leaderboard/client/PlayerList.jsx
nateps: Using just a function is how you write "stateless components" in React. In the docs they refer to it as function and class components: https://reactjs.org/docs/components-and-props.html

Here's another link about it: https://dev.to/mcarpenter/react---stateless-vs-stateful-components--5c3m

PR #321, opened 2019-10-30
Delay ops sent by query subscriptions until after polling is complete

nateps:
When clients create a query subscription, ShareDB sends all ops that match documents returned by the query as well as diffs in which documents match. These results may be live-bound in the UI to a list displayed to the user.

ShareDB doesn't guarantee the sequence in which these op & query result updates will be received, but currently the code sends the ops as soon as they are received, then issues the query against the database. At this point, both the 'op' middleware and the database poll are async operations, so they could complete in either order. However, in practice ops are probably received first, since the database poll likely takes much more time than the op middleware.

This change makes it so that ops are consistently ordered after diffs in results by delaying sending of ops until polling is complete. The main advantage is that UIs are likely to be less janky, since it is less likely that one would see a result change, then see that result disappear or get reordered after a short delay. We're still not sending the polling results and ops together, so now we might see the results order change, then the result itself change just after; however, this gap should be much shorter. In particular, I think it is a better user experience in the most noticeable case where a result no longer matches and is removed from the list.

More ideally, it would be nice to send the diff and the op together to the client in the same message or in two messages that are always sent together synchronously, but that would be a much bigger code change that I don't want to tackle right now.

### DIFF
--- a/lib/query-emitter.js
+++ b/lib/query-emitter.js
@@ -63,7 +63,12 @@ QueryEmitter.prototype._emitTiming = function(action, start) {
 };
 
 QueryEmitter.prototype._update = function(op) {
+  // Note that `op` should not be projected or sanitized yet. It's possible for
+  // a query to filter on a field that's not in the projection. skipPoll checks
+  // to see if an op could possibly affect a query, so it should get passed the
+  // full op. The onOp listener function must call backend.sanitizeOp()
   var id = op.d;
+  var pollCallback = this._defaultCallback;
 
   // Check if the op's id matches the query before updating the query results
   // and send it through immediately if it does. The current snapshot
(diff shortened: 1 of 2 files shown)


lib/query-emitter.js
```diff
@@ -90,28 +95,35 @@ QueryEmitter.prototype._update = function(op) {
   // that removed the doc from the query to cause the client-side computed
   // list to update.
   if (this.ids.indexOf(id) !== -1) {
-    // Note that this op is not projected or sanitized yet. It's possible for a
-    // query to filter on a field that's not in the projection. skipPoll checks
-    // to see if an op could possibly affect a query, so it should get passed
-    // the full op. The onOp listener function must call backend.sanitizeOp()
-    this.onOp(op);
+    var emitter = this;
+    pollCallback = function(err) {
```
ericyhwang: Nice. When I was looking at this, I was thinking of making a separate function for the query-poll code below, but this works just as well!

PR #322, opened 2019-11-04
Add Presence functionality

alecgibson:
This change adds the ability for clients to broadcast information about
"Presence" - the notion of a client's position or state in a particular
document. This might be represent a cursor in a text document, or a
highlighted field in a more complex JSON document, or any other
transient, current information about a client that shouldn't necessarily
be stored in the document's chain of ops.

The main complication that this feature solves is the issue of keeping
presence correctly associated with the version of a `Doc` it was created
at. For example, in a "naive" implementation of presence, presence
information can arrive ahead of or behind ops, which - in a text-based
example - can cause the cursor to "jitter" around the change. Using the
ShareDB implementation will ensure that the presence is correctly
transformed against any ops, and will ensure that presence information
is always consistent with the version of the document. We also locally
transform existing presence, which should help to keep (static) remote
presence correctly positioned, independent of latency.

In order to facilitate this, the feature must be used with an OT type
that supports presence. The only requirement for supporting presence is
the support of a `transformPresence` method:

```javascript
type.transformPresence(presence, op, isOwnOperation): presence;
```

* `presence` _Object_: the presence data being transformed. The type
  will define this shape to be whatever is appropriate for the type.
* `op` _Op_: the operation against which to transform the presence
* `isOwnOperation`: _boolean_: whether the presence and the op have the
  same "owner". This information can be useful for some types to break
  ties when transforming a presence, for example as used in
  [`rich-text`][1]

This work is based on the [work][2] by @gkubisa and @curran, but with
the following aims:

  - avoid modifying the existing `Doc` class as much as possible, and
    instead use lifecycle hooks
  - keep presence separate as its own conceptual entity
  - break the presence subscriptions apart from `Doc` subscriptions
    (although in practice, the two are obviously tightly coupled)
  - allow multiple presences on a single `Doc` on the same `Connection`

[1]: https://github.com/quilljs/delta#tranformposition
[2]: https://github.com/share/sharedb/pull/288


curran:
Woohoo! Thrilled to see this fresh take on presence.

alecgibson:
Let's got a step further here, and completely remove the notion of presence from `Doc` (ie `Doc` has no presence-related methods or properties). Instead, let's have an intermediary notion of `DocPresence`, which exists separately to `Doc`, but is responsible for handling all state of tracking remote presence, and also of creating instances of local presence. In this way, we can easily handle our presence subscription without needing any instances of local presence, but also have none or multiple local presences.

We can also completely decouple presence from Docs altogether, by having a notion of subscribing to presence channels (eg mouse cursor position on a screen might be a notion of presence that is not coupled to a Doc). The Doc presence should be an extension of this more generic presence.

So the API might look something like:

```javascript
const presence = connection.getDocPresence(collection, id);
// Start listening for presence
presence.subscribe((err) => {...});
// Handle incoming presence (eg update a cursor position)
presence.on('receive', (remotePresence) => {...});
// Create a instances of local presence
const localPresence1 = presence.create(id1);
const localPresence2 = presence.create(id2);
// Presences can send independently
localPresence1.submit({...});
localPresence1.destroy((err) => {...});
// Can unsubscribe, but still broadcast
presence.unsubscribe((err) => {...});
localPresence2.submit({...});
// Broadcast a removal of all local presence, and free up local memory
presence.destroy((err) => {...});
```

alecgibson:
I think the one thing I still find a little bit weird about its current incarnation is that if you're not subscribed, you won't receive any requests for your presence, so you won't inform any new subscribers. For example:

- client 1 is subscribed
- client 2 is *not* subscribed, and submits presence
- client 1 receives client 2's presence
- client 3 connects and subscribes, requesting all current presence
- client 2 is not subscribed, so doesn't receive the request, and doesn't tell client 3 its old presence (despite having submitted it)

Maybe this is an edge case we don't need to think about too hard? The mainline case will presumably be all clients being subscribed all the time.

I guess the other thing to think about is potentially garbage collecting stale clients who disconnected abruptly without a chance to update other clients. Maybe we can handle this in the `Agent` on the stream close? If the `Agent` stored all local presences associated with a connection, it could then broadcast an update to remote clients saying that the client has disconnected.

curran:
Greetings @alecgibson 

I'm looking at this work with fresh eyes and would like to try it out. Before diving into trying to craft a working example myself I thought I'd ask, do you have any working examples that use this implementation for presence features?

Thanks!

curran:
Also @nateps @ericyhwang I'm curious if there is a possibility that this implementation might get merged at some point. I know you've got a lot to juggle but just curious on a high level, is the possibility there? I'm again evaluating the `@teamwork/sharedb` vs `sharedb` packages and am tentatively switching back to mainline `sharedb` to get some of the recent fixes published. Having presence implemented here would really seal the deal!

alecgibson:
@curran I haven't got an example working at the moment, because the API was a bit unstable through a couple of PR meetings. However, if you have a look at the tests, the usage should hopefully be pretty self-evident. Note that the `README` in this PR hasn't be correctly updated (**this is also a note to self to update!**). I haven't made the last couple of meetings, and with the holidays coming up, probably won't get much more done on this until the new year, but I am trying to push ahead with this! This is 100% something we need to use, so I'll be seeing it through.

If you want to try to get something up and running, I'd love the feedback. A very quick overview of the API as it currently stands (NB: using `async` / `await` syntax out of sheer laziness for not nesting callbacks on Github):

```javascript
const doc = connection.get('books', 'northern-lights');
await doc.subscribe();

// This presence object represents all presence on the Doc. Can have multiple remote and
// local presences, available on `remotePresences` and `localPresences` properties
const presence = connection.getDocPresence('books', 'northern-lights');
presence.on('receive', (id, remotePresence) => {/** handle presence **/});
await presence.subscribe();

const localPresence = presence.create('user-1');
localPresence.submit({...});
```

The main piece of the puzzle that's missing is a `type` that supports presence. All it needs to do is support `transformPresence`. If you want to play around with `rich-text`, I've got a fork that should work with this presence PR [here](https://github.com/reedsy/rich-text/tree/presence) (note that you need to be on the `presence` branch). Usage with this type would look something like:

```js
localPresence.submit({index: 4, length: 0}); // uses the same structure as Quill
presence.on('receive', (id, remotePresence) => {
  // Handle presence.
  // remotePresence will have the same structure as above: {index: number; length: number}
});
```

Since the `transformPresence` method name is the same as https://github.com/share/sharedb/pull/288 you may also be able to use types that were compatible with that.

lib/agent.js
```diff
@@ -627,6 +673,83 @@ Agent.prototype._fetchSnapshotByTimestamp = function(collection, id, timestamp,
...
+  var stream = this.subscribedPresences[channel];
+  if (stream) stream.destroy();
+  callback(null, {ch: channel, seq: seq});
+};
+
+Agent.prototype._getPresenceChannel = function(channel) {
+  // TODO: May need to namespace this further if we want to have automatic Doc channels that don't
+  // clash with arbitrary user input (eg if a user decides to name their channel the same as a doc collection)
+  // TODO: What if a user creates a collection called _presence?
+  return '_presence.' + channel;
```
alecgibson: Do we want to namespace channels somehow? As I understand it, collections currently exist at the "root", and there would be no way to namespace them without breaking existing changes.

lib/client/doc.js
```diff
@@ -844,6 +849,28 @@ Doc.prototype.resume = function() {
   this.flush();
 };
 
+Doc.prototype.subscribeToPresence = function(callback) {
```
alecgibson: Let's remove these utility methods and keep `Doc` as clean as possible for now.

lib/client/doc.js
```diff
@@ -101,6 +101,9 @@ function Doc(connection, collection, id) {
   // the time of op submit, so it may be toggled on before submitting a
   // specifc op and toggled off afterward
   this.preventCompose = false;
+
+  this.remotePresences = {};
```
alecgibson: Let's also get rid of these instance properties.

lib/client/doc.js
```diff
@@ -571,15 +576,15 @@ Doc.prototype._otApply = function(op, source) {
 
     // The 'before op' event enables clients to pull any necessary data out of
     // the snapshot before it gets changed
-    this.emit('before op', op.op, source);
+    this.emit('before op', op.op, source, op.src);
```
alecgibson: Let's think about if we can get rid of this extra `src` emission. We basically need to know if the op is our own op in `RemoteDocPresence._handleOp`, and in `Backend.transformPresenceToLatestVersion`

alecgibson:
@ericyhwang this should be ready for review. I've updated the `README`, and added an example to the `examples` folder if you want to play around with it (cc @curran ).

I think the only outstanding issue is figuring out what to do with emitting `src` on `op` (which I'm assuming we'll discuss in the meeting next week).

curran:
Woohoo! Looking forward to giving the examples a spin. Will let you know how it goes.

curran:
@alecgibson I tried out the demo and it works well!

I would be curious to adapt your rich text demo to make a plain text demo work, using deeply nested JSON0 fields. @houshuang and I have done some work on a plain text demo of presence in https://github.com/datavis-tech/json0-presence-demo .

It would be interesting to see if it's possible to re-work that demo to use your new implementation here.

I'm ultimiately interested in implementing presence across multiple locations within a single document. For example a single ShareDG document contains multiple "files" (just strings) that users could be editing, and the UI should show presence localized to each. Looking forward to seeing if that's possible with this implementation.

Thanks for all your work on this!

alecgibson:
@curran I imagine it should be relatively simple to extend to plain text. If you don't want to publish your own type, you can always monkey-patch like I've done in this example.

lib/agent.js
```diff
@@ -29,6 +29,19 @@ function Agent(backend, stream) {
   // Map from queryId -> emitter
   this.subscribedQueries = {};
 
+  // Track which documents are subscribed to presence by the client. This is a
+  // map of channel -> stream
+  this.subscribedPresences = {};
+  // Highest seq received for a subscription request. Any seq lower than this
+  // value is stale, and should be ignored. Used for keeping the subscription
+  // state in sync with the client's desired state
+  this.presenceSubscriptionSeq = 0;
```
ericyhwang: Have you been able to find or construct a scenario where this sequence guard is needed?

My understanding is that websockets and browserchannel both guarantee in-order delivery. I'm not sure about SockJS's XHR-based transports - some brief searching was inconclusive.

Even if the transport were naive XHR, meaning messages could get delivered out of order, I think this would need to be keyed by channel. A scenario - If a client sends `ps foo.a` then `ps foo.b`, but naive XHRs are received in the opposite order, then `ps foo.b` would set a new high agent-wide watermark, causing `ps foo.a` to get ignored.

examples/rich-text-presence/client.js
```diff
@@ -0,0 +1,142 @@
...
+
+function addClient() {
+  var socket = new ReconnectingWebSocket('ws://' + window.location.host);
+  var connection = new sharedb.Connection(socket);
+  var doc = connection.get(collection, id);
+  doc.subscribe(function(err) {
+    if (err) throw err;
+    var quill = initialiseQuill(doc);
+    var color = '#' + tinycolor.random().toHex();
+    var id = 'client-' + (clients.length + 1);
```
ericyhwang: As a note, the `presenceId`s being fixed in this demo means that opening the demo in two different tabs can result in some pretty weird stuff happening.

I didn't dig into it too much, but I suspect that "client-1" in tab A and "client-1" in tab B were both trying to update the same presence, and the displayed presence in the other clients was changing based on which tab had a higher `seq`. I think we did discuss that last week, with a potential solution being to prepend a generated client id like "original src".

lib/client/presence/local-doc-presence.js
```diff
@@ -0,0 +1,109 @@
...
+    if (!this.connection.canSend) return;
+
+    this._pendingMessages.forEach(function(message) {
+      message.t = this._doc.type.uri;
+      message.v = this._doc.version;
+      this.connection.send(message);
+    }.bind(this));
+
+    this._pendingMessages = [];
+  }.bind(this));
```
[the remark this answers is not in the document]
alecgibson: Sure. I think I vaguely find the `bind` approach more readable, but probably only because I've gotten used to using arrow functions.

lib/agent.js
```diff
@@ -645,6 +705,85 @@ Agent.prototype._fetchSnapshotByTimestamp = function(collection, id, timestamp,
...
+    v: request.v,
+    t: request.t
+  };
+};
+
+Agent.prototype._subscribePresence = function(channel, seq, callback) {
+  var presenceChannel = this._getPresenceChannel(channel);
+  this.backend.pubsub.subscribe(presenceChannel, function(error, stream) {
+    if (error) return callback(error);
+    if (seq < this.presenceSubscriptionSeq) return callback(null, {ch: channel, seq: seq});
```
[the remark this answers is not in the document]
alecgibson: Good catch. I'm fixing this by just destroying the stream before the early return. Can't really think of a good way to cover this with a test, so I'm all ears if you have any ideas.

lib/agent.js
alecgibson: Ha yes, should definitely be per-channel. Good spot.

There's definitely a corner case where it's needed, even with in-order message guarantees:

1. Client subscribes
2. Client immediately, synchronously unsubscribes
3. Agent receives subscribe request, fires off an async request for a stream from the backend
4. Agent receives the unsubscribe request, which is processed synchronously (even if async, would be done in a different amount of time to the stream request)
5. Client is unsubscribed
6. Agent finally receives the subscription stream from backend, and subscribes
7. Client finishes subscribed, even though they requested the unsubscribe after the subscribe

This is obviously a super niche corner case, and I guess consumers shouldn't be _too_ surprised if they get weird behaviour when not waiting for their callbacks to resolve before making further requests, but I do think that this is the most "logical" thing to do.

If we think this is too niche for the maintenance overhead, I can just cut this behaviour.

examples/rich-text-presence/client.js
alecgibson: Ha yes I hadn't tried multiple tabs. To be fair, maybe just showing one client per tab and just adding some instructional text to open a new tab to test is easiest. It would:

- fix this issue
- simplify example code (always a good thing)
- be a bit more "realistic"

What do you think?

alecgibson: I've gone ahead and simplified it.

lib/client/connection.js
```diff
@@ -714,3 +721,52 @@ Connection.prototype._handleSnapshotFetch = function(error, message) {
...
+};
+
+Connection.prototype._handlePresence = function(error, message) {
+  var presence = util.dig(this._presences, message.ch);
+  if (presence) presence._receiveUpdate(error, message);
+};
+
+Connection.prototype._handlePresenceRequest = function(error, message) {
+  var presence = util.dig(this._presences, message.ch);
+  presence._broadcastAllLocalPresence(error, message);
```
ericyhwang: Is this intentionally not null-checked like the others?

lib/client/presence/remote-doc-presence.js
```diff
@@ -0,0 +1,151 @@
...
+};
+
+RemoteDocPresence.prototype._setPendingPresence = function() {
+  if (this._pendingSetPending) return;
+  this._pendingSetPending = true;
+  var presence = this;
+  this._doc.whenNothingPending(function() {
+    presence._pendingSetPending = false;
+    if (!presence._pending) return;
+    if (presence._pending.seq < presence.seq) return presence._pending = null;
```
ericyhwang: Since a client connection's `seq` currently resets to 1 (or 0?) upon reconnecting, I'm worried about buggy behavior in scenario:
1. Client A does some ops and presence pushing, ending up at a `seq` of 20.
2. Client B is subscribed to Client A's presence, so its RemotePresence for Client A has a `presence.seq` of 20.
3. Client A disconnects and reconnects, so its `seq` resets. It starts sending presence again with `seq` <20.
4. Client B ignores those messages from Client A until the `seq` is over 20 again.

As we discussed, without the assumption of persistent src+seq for a client session, we'll either have to take this out, or to add a presence version.

If we go the presence versioning route, we'll similarly want to be cautious about two sessions using the same `presenceId`, otherwise two sessions with the same `presenceId` can overwrite each others' presence.

lib/client/connection.js
alecgibson: Good catch!

lib/client/presence/remote-doc-presence.js
alecgibson: I've addressed the `seq` issue by just adding a `pv` presence version field to the message instead. (NB: subscribe / unsubscribe still uses a `seq`, but that's derived in `Presence`, separately from the `Connection` `seq` so should be fine to use. Maybe could do with a better name...).

As far as `presenceId` being unique goes, I think I'm not feeling very comfortable with suffixing the ID. I think having two different IDs for something is quite confusing, both in the codebase (I've already tripped over the channel namespacing a few times, for example), and from the consumer's point of view (eg when directly accessing a local presence by ID in `presence.localPresences`).

A possible middle ground could be providing a sensible default random ID if the presence ID is not provided (ie make the ID optional), so that we get sensible default behaviour, and if the client doesn't care about what the presence ID represents, then they don't have to care (they can naturally still query `localPresence.presenceId` to find out what random ID has been assigned to them).

curran:
Excited to see activity here! Hoping to carve out some time to implement a plain text presence example on top of this work as proof of concept, for inclusion alongside the rich text demo.

alecgibson:
@ericyhwang I've rebased and updated to use `Agent._src`; should be good to review now

lib/client/doc.js
alecgibson: Let's keep this, but bury it one level deep in an object with a `src` property for forwards-compatibility and self-documentation. We should also update the `README`.

lib/agent.js
```diff
@@ -369,6 +419,19 @@ Agent.prototype._handleMessage = function(request, callback) {
         return this._fetchSnapshot(request.c, request.d, request.v, callback);
       case 'nt':
         return this._fetchSnapshotByTimestamp(request.c, request.d, request.ts, callback);
+      case 'p':
+        var presence = this._createPresence(request);
```
ericyhwang: Might be a good idea to put this Agent presence-handling code behind a flag, so it's explicitly opt-in.

We know that this implementation doesn't necessarily scale to hundreds or thousands of clients on a single doc, so it's nice for a server to not have to worry about presence unless it's opted in.

alecgibson: Let's hide this behind a feature flag for now. Will need to update the docs, saying that it's behind a flag and sort of beta-y, with potential for fan-out issues with lots of presence clients.

alecgibson:
@ericyhwang I've hidden presence behind a feature flag. Note that the feature flag disables presence broadcasts and subscribe requests, but will still let users unsubscribe in case the flag is toggled while the server is running.

alecgibson:
@ericyhwang I've also made the unique local presence ID an optional argument; if consumers don't provide one, we'll just assign the local presence with one for them.

gkubisa:
Great work guys! I'm really happy to see this feature complete. :tada:

curran:
Amazing! Really really excited about this. Thank you everyone for your hard work here.

neilcarv:
Hey guys, this work is brilliant. A quick question, I can see a PR for json0 and an issue open for json1 to support a transformPresence method. Would that be something that will be merged/added soon? Also,  in general, should json0 or json1 be used?

Thanks, Neil

curran:
Hi @neilcarv ! This is something near and dear to my heart and I'm glad you asked.

Current status is: there's no demo code for the new implementation of ShareDB presence used with JSON0. I plan to work on one soon, but haven't been able to dig into it yet.

Here's a PR agains JSON0 that implements the OLD version of presence https://github.com/ottypes/json0/pull/31/files . I would love to see this updated to the presently published API. I'm not sure how far it will diverge, but hopefully getting it to work should be relatively painless.

I'd also like to see this demo ported to use the new presence implementation https://github.com/datavis-tech/json0-presence-demo

presence

selections

I hope to take a stab at this some time but if anyone else has interest, feel free to have a crack at it!

Issue #308, opened 2019-09-13
Modifying a document during a fetch


alecgibson:
Following on from the meeting on Wednesday, I thought I'd write up some notes and propose a documentation update, or maybe even an API change.

There are a couple of cases in our ShareDB usage where we would like to modify a document _before_ being fetched by a client. For example:

  - fetch or create: when the client performs a fetch, they will _always_ get a sensible document back, and won't need to know anything about how to initialise a document)
  - lazy migration of documents: when a client fetches a document with an old schema version, the document is migrated to the newer version

Below is a proposal for how to deal with this case. If this is a sensible proposal, I think we should add it to the documentation, or perhaps add a helper method.

```javascript
backend.use('readSnapshots', (request, callback) => {
  const connection = backend.connect();
  let pendingModifications = 0;

  for (let i = 0; i < request.snapshots.length; i++) {
    const snapshot = request.snapshots[i];
    if (shouldModify(snapshot)) {
      pendingModifications++;
      const modificationOp = modify(snapshot);
      const doc = connection.get(request.collection, snapshot.id);

      // Ingest rather than fetch to avoid an infinite loop
      // of readSnapshots middleware
      doc.ingestSnapshot(snapshot, (error) => {
        if (error) return callback(error);
        doc.submitOp(modificationOp, (error) => {
          if (error) return callback(error);

          // Update the snapshot so the client gets the modified
          // version of the document
          snapshot.type = doc.type.uri;
          snapshot.v = doc.version;
          snapshot.data = doc.data;

          // Check if all snapshots have updated, and finally
          // unblock the fetch
          pendingModifications--;
          if (!pendingModifications) callback();
        });
      });
    }
  }
});
```

If this approach is acceptable, then we could potentially put together some sort of helper function on the `readSnapshots` `request` object. The client code might look something like:

```javascript
backend.use('readSnapshots, (request, callback) => {
  const modifier = (snapshot, request, modifierCallback) => {
    // Do something to our snapshot and return an op to update it.
    // We pass in the request in case we want other contextual
    // information, like the collection, or agent.custom
    if (snapshot.schemaVersion < 2) {
      return modifierCallback(null, {p: ['schemaVersion'], oi: 2});
    }

    if (somethingBadHappened()) {
      return modifierCallback(new Error('Uh-oh!'));
    }

    // Returning a falsy op means we won't bother 
    // updating the snapshot
    return modifierCallback();
  };

  // This method will take care of:
  // - iterating over all snapshots in the request
[...]

PR #331, opened 2019-12-03
Add ability for "readSnapshots" middleware to reject reads of specific snapshots for bulk fetch/subscribe

ericyhwang:
## Background

The "readSnapshots" middleware can be used to reject snapshot reads, such as for access control. However, it's currently all or nothing. When doing a bulk fetch or subscribe, the middleware receives many snapshots at once, and you currently can't reject reads for specific snapshots while allowing others through.

Bulk fetch/subscribe is automatically used on reconnection, when multiple fetches/subscribes by id are issued while a client is offline. It can also be manually triggered via the currently-undocumented `Connection#startBulk` and `#endBulk` methods.

## Changes

1. Add ability for "readSnapshots" middleware to reject reads of specific snapshots for bulk fetch/subscribe
2. Fix client handling of errors in replies to bulk actions. Such errors were getting silently ignored before.

### Add ability for "readSnapshots" middleware to reject reads of specific snapshots for bulk fetch/subscribe

This is accomplished via calling `context.rejectSnapshotRead(snapshot, error)` or `context.rejectSnapshotReadSilent(snapshot, errorMessage)`.

```js
backend.use('readSnapshots', (context, callback) => {
  if (context.collection === 'dogs') {
    for (const snapshot of context.snapshots) {
      if (snapshot.data.trainerId !== context.agent.custom.userId) {
        context.rejectSnapshotRead(snapshot, new Error('Cannot see dogs for other trainers'));
      }
    }
  }
  callback();
});
```

Setting a snapshot error has different effects depending on the situation:

* For single-doc reads, which is the typical case, a snapshot-specific error is "promoted" to an error on the entire request, as if the middleware had called `callback(error)`. This enables backwards compatibility with older clients, and it's just more useful to have the unwrapped error in a single-doc read case.
* For bulk fetch/subscribe of multiple docs by id, snapshot-specific errors will cause those specific docs' data to _not_ be sent to the client. Instead, the client docs will each receive their specific corresponding error. Other non-error snapshots will be sent to the client as usual.
* For queries, any snapshot-specific error will cause the entire query to receive an error. This is because queries have no clean way of handling an error in a specific doc in the query's results.
* `rejectSnapshotReadSilent` will cause the ShareDB client to treat it as a silent rejection, not passing the error back to user code. ShareDB itself will still cancel fetches/subscribes as usual.

### Fix client handling of errors in replies to bulk actions

In the client, an incoming `message.error` gets extracted and deleted from the message:
https://github.com/share/sharedb/blob/ca35a5c9faf1d11b827302fb328a7b82ed714eb9/lib/client/connection.js#L177-L184

However, `_handleBulkMessage` then tries to use that `message.error`, which is always undefined at that point.
https://github.com/share/sharedb/blob/master/lib/client/connection.js#L272-L283

That means errors in bulk action replies were always getting silently ignored.

This PR fixes that issue, so such errors get properly emitted now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -602,16 +602,16 @@ An `Agent` is the representation of a client's `Connection` state on the server.
 The `Agent` will be made available in all [middleware](#middlewares) requests. The `agent.custom` field is an object that can be used for storing arbitrary information for use in middleware. For example:
 
 ```javascript
-backend.useMiddleware('connect', function (request, callback) {
+backend.useMiddleware('connect', (request, callback) => {
   // Best practice to clone to prevent mutating the object after connection.
   // You may also want to consider a deep clone, depending on the shape of request.req.
   Object.assign(request.agent.custom, request.req);
   callback();
 });
 
-backend.useMiddleware('readSnapshots', function (request, callback) {
-  var connectionInfo = request.agent.custom;
-  var snapshots = request.snapshots;
+backend.useMiddleware('readSnapshots', (request, callback) => {
+  const connectionInfo = request.agent.custom;
+  const snapshots = request.snapshots;
 
   // Use the information provided at connection to determine if a user can access the snapshots.
   // This should also be checked when fetching and submitting ops.
@@ -619,9 +619,7 @@ backend.useMiddleware('readSnapshots', function (request, callback) {
     return callback(new Error('Not allowed to access collection ' + request.collection));
   }
   // Check each snapshot individually.
-  // In ES6, this could be `for (const snapshot of snapshots)`
-  for (var i = 0; i < snapshots.length; i++) {
-    var snapshot = snapshots[i];
+  for (const snapshot of snapshots) {
     if (!userCanAccessSnapshot(connectionInfo, request.collection, snapshot)) {
       request.rejectSnapshotRead(snapshot,
         new Error('Not allowed to access snapshot in ' request.collection));
@@ -635,10 +633,10 @@ backend.useMiddleware('readSnapshots', function (request, callback) {
 // potentially making some database request to check which documents they can access, or which
 // roles they have, etc. If doing this asynchronously, make sure you call backend.connect
 // after the permissions have been fetched.
-var connectionInfo = getUserPermissions();
+const connectionInfo = getUserPermissions();
 // Pass info in as the second argument. This will be made available as request.req in the
 // 'connection' middleware.
-var connection = backend.connect(null, connectionInfo);
+const connection = backend.connect(null, connectionInfo);
 ```
 
 ### Logging
--- a/README.md
+++ b/README.md
@@ -172,6 +172,7 @@ Register a new middleware.
     * `stream`: The duplex Stream provided to `share.listen` (for 'connect')
     * `query`: The query object being handled (for 'query')
     * `snapshots`: Array of retrieved snapshots (for 'readSnapshots')
+    * `rejectSnapshotRead(snapshot, error)`: Reject a specific snapshot read (for 'readSnapshots')
     * `data`: Received client message (for 'receive')
     * `request`: Client message being replied to (for 'reply')
     * `reply`: Reply to be sent to the client (for 'reply')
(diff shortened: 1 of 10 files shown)


alecgibson:
Setting an `error` on `Snapshot` maybe feels a bit magical to me? A possible other approach is to return an array of errors, where each error maps to the array of snapshots passed in? This also isn't particularly transparent, though, so maybe your approach is fine.

ericyhwang:
Yeah, it does feel a bit magical, but it's the cleanest option I've thought of so far, usage-wise.

I tried an array of errors in addition to an object "map" of errors, but as I was writing the tests, I realized actually using it that way was a bit awkward. You'd either have to remember to push `undefined` for no error, or use an indexed for loop and assign errors by index. You'd also have to keep a count of errors so that you can call back with `undefined` if there isn't actually an error. That all means two new local variables to track that state.

ericyhwang:
Nate and I talked about this a bit, and he brought up the same concern around mutating the snapshot, mentioning the slight performance hit for V8 to change object shapes.

He suggested using a map on the context object. I went with a method `context.rejectSnapshotRead(snapshot, error)` instead, and he's on board with that approach since there is apparently a semi-similar precedent in SubmitRequest.

I've made the changes and updated the description and README to reflect the new way of doing things.

Thoughts on the new approach?

alecgibson:
Yeah I think I'd had a similar idea for [mutating snapshots][1].

[1]: https://github.com/share/sharedb/issues/308

README.md
```diff
@@ -612,10 +613,19 @@ backend.useMiddleware('readSnapshots', function (request, callback) {
...
+  // Use the information provided at connection to determine if a user can access the snapshots.
   // This should also be checked when fetching and submitting ops.
-  if (!userCanAccessSnapshots(connectionInfo, snapshots)) {
-    return callback(new Error('Authentication error'));
+  if (!userCanAccessCollection(connectionInfo, request.collection)) {
+    return callback(new Error('Not allowed to access collection ' + request.collection));
+  }
+  // Check each snapshot individually.
+  // In ES6, this could be `for (const snapshot of snapshots)`
+  for (var i = 0; i < snapshots.length; i++) {
```
nateps: Since this example code is server only, we can write in ES6 syntax. I wouldn't worry about the fact that sharedb is still in ES5 style

lib/readSnapshots-request.js
```diff
@@ -0,0 +1,49 @@
+module.exports = ReadSnapshotsRequest;
```
nateps: This filename should be "read-snapshots-request.js". The convention in sharedb is that all filenames are lowercase and all multi-word names are dash separated.

lib/backend.js
```diff
@@ -252,15 +253,59 @@ Backend.prototype._sanitizeSnapshots = function(agent, projection, collection, s
...
+ * Returns an overall error from "readSnapshots" based on the snapshot-specific errors.
+ *
+ * - If there's exactly one snapshot and it has an error, then that error is returned.
+ * - If there's more than one snapshot and at least one has an error, then an overall
+ *   "ERR_SNAPSHOT_READS_REJECTED" is returned, with an `idToError` property.
+ *
+ * @param {Snapshot[]} snapshots
+ * @param {{[docId: string]: string | Error}} idToError
+ */
+function getReadSnapshotsError(snapshots, idToError) {
```
nateps: I think this function should be moved into ReadSnapshotsRequest. It needs to access the id map, which should really be private to the class.

lib/agent.js
```diff
@@ -517,8 +517,24 @@ Agent.prototype._fetchOps = function(collection, id, version, callback) {
 Agent.prototype._fetchBulk = function(collection, versions, callback) {
   if (Array.isArray(versions)) {
     this.backend.fetchBulk(this, collection, versions, function(err, snapshotMap) {
-      if (err) return callback(err);
-      callback(null, {data: getMapData(snapshotMap)});
+      var isPartialError = err != null && err.code === ERROR_CODE.ERR_SNAPSHOT_READS_REJECTED;
```
nateps: I don't think a partial error should produce an error in this callback. Could we simply have the error information in the snapshotMap? I wonder if we even need to change this code in agent if we do that

lib/error.js
```diff
@@ -41,6 +41,25 @@ ShareDBError.CODES = {
...
+   * A "readSnapshots" middleware rejected the reads of specific snapshots.
+   *
+   * This error code is mostly for server use and generally will not be encountered on the client.
+   * Instead, each specific doc that encountered an error will receive its specific error.
+   *
+   * The one exception is for queries, where a "readSnapshots" rejection of specific snapshots will
+   * cause the client to receive this error for the whole query, since queries don't support
+   * doc-specific errors.
+   */
+  ERR_SNAPSHOT_READS_REJECTED: 'ERR_SNAPSHOT_READS_REJECTED',
```
nateps: I'd like to get rid of this error and only have the individual errors on the snapshots themselves

lib/agent.js
ericyhwang: Yes, it's possible, though I did have to update the agent code to serialize errors from the new location.

lib/error.js
ericyhwang: This error is still needed, as it's sent to the client for queries, when at least one snapshot in the query results gets rejected, as the comment indicates.

curran:
Thanks for the work on this! It solves an issue I was just facing with the errors getting silently ignored. Great work!

PR #338, opened 2020-01-09
Silence logger in tests

alecgibson:
At the moment, the logger will log to console during tests, which is
noisy and undesirable.

This change adds a test setup file, which silences the logger by setting
all of its methods to no-ops.

This behaviour can be overridden with the `LOGGING` environment
variable:

```bash
LOGGING=true npm test
```

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,3 +1,4 @@
 --reporter spec
 --check-leaks
 --recursive
+--file test/setup.js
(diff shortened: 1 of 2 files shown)


ericyhwang:
Sounds reasonable to me.

Could be useful to add an environment variable to allow turning the logging back on in the tests, in case you're debugging something and want the more verbose logs.

Also, I believe this will only work when running everything under the `test/` directory at once. If we wanted this to apply when running a specific test file, we could `--require` this file in the mocha.opts.

alecgibson:
@ericyhwang good thinking. Done.

Issue #337, opened 2020-01-08
Idea: Protocol plugins, client/server capability handshake


ericyhwang:
These are some of my thoughts around enabling more extensibility of ShareDB and more rapid development of new ideas. I'm hoping to find some time in the next month or two to write a more fleshed out spec, but in the meantime, comments and more ideas are welcome!

## ShareDB message protocol plugins

Over the past couple years, there have been new features that require extending the ShareDB message protocol with different `request.a` "actions":

* `fetchSnapshotByTimestamp`
* Presence, still in progress

These are handled in large switch statements:

* On the server in Agent: https://github.com/share/sharedb/blob/v1.1.0/lib/agent.js#L334
* On the client in Connection: https://github.com/share/sharedb/blob/v1.1.0/lib/client/connection.js#L183

I've been thinking it would be good to implement a framework for message protocol plugins, and refactor more optional features like queries, `fetchSnapshotByTimestamp`, and presence into plugins. This has some nice advantages:

1. ShareDB users could choose exactly what features they want, keeping code and bundle sizes smaller.
2. We can develop new protocol-level features more rapidly as independent plugins, compared to developing them in ShareDB core. They can also be independently semver'd.
3. ShareDB users could author their own protocol plugins without needing to touch ShareDB core.

One of the requirements to accomplishing this is that clients and servers need to be able to do some sort of capability handshake, so they can agree on what protocol plugins to use for the session and at what versions.

## Client/server capability handshake

I haven't put much thought into this yet. It could work like a simpler version of a TLS handshake. It would need to include at least:

* Capability names
* Capability versions

Potential use cases for a capability handshake:

* ShareDB message protocol plugins
* Making it easier to do major semver updates to ShareDB core
* Negotiating a default OT type or supported OT types in general: https://github.com/share/sharedb/issues/284
* Custom application-specific capabilities defined by ShareDB users

curran:
It's a great idea! ShareDB needs this.

alecgibson:
Vague high-level brain dump:

We'll need to think about what to do in non-happy handshake cases. For example, it may be acceptable to run without some plugins, but not without others. Even here, the server and client may disagree about what's important. In the case where a plugin isn't loaded, but this is deemed acceptable, it would be nice to inform the client, so they can eg alert the user about reduced functionality / prompt them to reload the page, etc. Similarly if an older version of a plugin is agreed upon, the client may want to handle this in some way.

[...]

PR #341, opened 2020-01-30
Add basic handshake negotiation

alecgibson:
This change updates how the client and server communicate when first
connecting, in a non-breaking way.

## Current behaviour

  - client creates and connects to a socket
  - client constructs a `Connection` object with the socket
  - the backend binds to the socket and constructs an `Agent`
  - the `Agent` sends an `init` message down the socket when constructed
  - the `Connection` receives the `init` message and initialises itself
    using information in the message

## Motivation for change

There are currently a few things that would be nice to "negotiate" when
a client connects. For example, clients may wish to:

  - the same `src` value when reconnecting, so that it persists as a
    stable "session ID" (which has been implemented here)
  - set their default type
  - establish the exact versions of OT types being used
  - establish plugins and versions to use

More detail can be found here: https://github.com/share/sharedb/issues/337

## New behaviour

  - client creates and connects to a socket
  - client constructs a `Connection` object with the socket
  - the backend binds to the socket and constructs an `Agent`
  - the `Agent` sends an `init` message down the socket when constructed
  - the `Connection` checks the `init` message to see if the backend is
    capable of handshaking
  - if the backend _cannot_ handshake (ie it's old), then the client
    initialises itself as before
  - **new behaviour starts here**
  - if the backend _can_ handshake, then it will disregard the `init`
    message, and send its own `hs` "handshake" message to the server
  - the backend receives the handshake message and sends a response
  - the `Connection` receives the new `hs` message and initialises itself
    using information in the message (like before, but with more or
    different information information)

## Other changes

As a convenience for some tests, which had to be tweaked (because they
assumed the connection was established immediately), this change also
adds an optional `callback` parameter to `Backend.connect`, which will
be invoked once the connection handshake completes.

We also add a guard for arbitrarily large `seq`, which will no longer be
reset on reconnection, so might hit integer overflow on particularly
long-lived `Connection` objects.


lib/backend.js
```diff
@@ -125,6 +137,8 @@ Backend.prototype.connect = function(connection, req) {
  * (The agent is available through all middleware)
  */
 Backend.prototype.listen = function(stream, req) {
+  // TODO: When a Connection reconnects with a persisted src, it will change the
+  // Agent's ID to be something different than what it had in the connection middleware
```
alecgibson: Reminder that we should discuss this. I can't think of a particularly good way of avoiding it (we could potentially set `Agent` to initially have a `null` ID? At least that would be less surprising than it changing?).

We could potentially only trigger the middleware on the `hs` event, but obviously that would break for older clients, so probably not a good idea. We could add an _extra_ middleware, and deprecate this one? But again can't think of a good way of shimming without triggering middlewares twice.

lib/client/connection.js
```diff
@@ -305,8 +285,6 @@ Connection.prototype._handleBulkMessage = function(err, message, method) {
 };
 
 Connection.prototype._reset = function() {
-  this.seq = 1;
-  this.id = null;
```
alecgibson: Note that I've deleted this (so `id` is never `null` after initialisation). The only place this seems to matter is in `Doc` (which I've fixed). I know we spoke about leaving this as-is, but it seems a little bit "messy" to stash the id in another property just to put it back here? I'm not sure I see the need.

That said, is this potentially a breaking change? ie might consumers _assume_ this behaviour on a disconnect?!

lib/backend.js
```diff
@@ -100,7 +100,12 @@ Backend.prototype.close = function(callback) {
   finish();
 };
 
-Backend.prototype.connect = function(connection, req) {
+Backend.prototype.connect = function(connection, req, callback) {
+  if (typeof connection === 'function') {
```
alecgibson: This overload is possibly a bit contentious? Just because I've not handled all overload cases. It was mostly for convenience. We can just get rid of it if we think it's too much faff.

lib/client/connection.js
```diff
@@ -714,3 +692,43 @@ Connection.prototype._handleSnapshotFetch = function(error, message) {
   delete this._snapshotRequests[message.id];
   snapshotRequest._handleResponse(error, message);
 };
+
+Connection.prototype._handleLegacyInit = function(message) {
+  // If the minor protocol version has been set, we can ignore this legacy
+  // init message, and continue with the updated handshake protocol.
+  if (message.protocolMinor) return this._initializeHandshake();
```
alecgibson: I couldn't think of a good way of adding a test case for legacy clients. I assume that's okay?

lib/backend.js
ericyhwang: Let's not do the overload, then, since we can keep it backwards-compatible without the overload.

lib/client/connection.js
```diff
@@ -714,3 +692,43 @@ Connection.prototype._handleSnapshotFetch = function(error, message) {
   delete this._snapshotRequests[message.id];
   snapshotRequest._handleResponse(error, message);
 };
+
+Connection.prototype._handleLegacyInit = function(message) {
```
alecgibson: Let's move `_initializeHandshake` out of here to save half a round trip, and make it easier to remove legacy behaviour in future.

lib/client/connection.js
ericyhwang: Let's move the `_initializeHandshake()` call to initial connect time, so that it's not dependent on receiving the legacy init message from the server. That way, in the future, we can remove the legacy init message without breaking "protocol 1.1" clients.

lib/backend.js
ericyhwang: Options we discussed:
- Keep connect as-is, null out the ID. Breaking change, if people are looking at the id.
- Add a "client init" or "handshake" middleware. Problem is, the agent id changing after connect could still be considered a breaking change, if someone has stored the id somewhere.
- Delay emitting the connect event until we know there's a confirmed id. Problem - this is tricky to do for older clients that don't go through the handshake. We could potentially do something like emit "connect" just prior to processing the first message and/or after some timeout, but that gets messy.

Feels like the best way longer term is to do some form of breaking change, but I'm not sure if there's some fully backwards-compatible shim we can do in the meantime, as you say. Let's sleep on this for a day or two.

alecgibson: This week we discussed:

- add an `agent.src` field, which is initialised to `null`, and then given the connection ID provided by the new handshake
- `agent.src` will remain `null` for legacy clients (who never send an updated ID)
- `agent.clientId` is still set on construction, and never changes
- any internal ShareDB code that references `agent.clientId` should be updated to use some sort of getter method, which will provide `agent.src || agent.clientId`

This way, we can keep our change non-breaking, but still allow the client to suggest an ID. Anyone who cares can look out for the new `src` field as the "canonical" ID.

lib/client/connection.js
alecgibson: As discussed, let's just ignore the idea of this being "breaking"; it's arguably an implementation concern.

lib/client/connection.js
alecgibson: Let's add a special little test suite that runs using its own `package.json` that has a dependency on the old version of ShareDB.

PR #347, opened 2020-02-24
handshake immediately if socket is already open

dead-horse:
In our situation, we will connect to the server and authorize the request first, so if socket is already open, we must do handshake immediately because `onopen` event will not be emitted again.

### DIFF
--- a/lib/client/connection.js
+++ b/lib/client/connection.js
@@ -139,6 +139,10 @@ Connection.prototype.bindToSocket = function(socket) {
     }
   };
 
+  // If socket is already open, do handshake immediately.
+  if (socket.readyState === 1) {
+    connection._initializeHandshake();
+  }
   socket.onopen = function() {
     connection._setState('connecting');
     connection._initializeHandshake();
--- a/test/client/connection.js
+++ b/test/client/connection.js
@@ -1,3 +1,4 @@
+var sinon = require('sinon');
 var expect = require('chai').expect;
 var Backend = require('../../lib/backend');
 var Connection = require('../../lib/client/connection');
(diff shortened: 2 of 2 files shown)


lib/client/connection.js
```diff
@@ -139,10 +139,16 @@ Connection.prototype.bindToSocket = function(socket) {
...
 
-  socket.onopen = function() {
-    connection._setState('connecting');
+  // If socket is already open, do handshake immediately.
+  // Otherwize do handshake when socket open
+  if (socket.readyState === 1) {
     connection._initializeHandshake();
-  };
+  } else {
+    socket.onopen = function() {
```
alecgibson: I'm not sure about hiding this in an `else` statement. For example, I'm pretty sure if you use a reconnecting websocket, you'll still want the `onopen` listener bound so that if your connection drops and gets re-established, this is correctly called.

dead-horse: hmmm. I'm not sure if `_initializeHandshake` is idempotent or not?

alecgibson: It should eventually just trigger `Connection._initialize` (after a round-trip), which was the behaviour before the handshake was added (previously simply triggered by the `Agent` sending an `init` message on construction, when binding to a new socket).

test/client/connection.js
```diff
@@ -142,11 +142,18 @@ describe('client connection', function() {
...
-    it('initial connection.state is connecting, if socket.readyState is OPEN', function() {
-      // https://html.spec.whatwg.org/multipage/web-sockets.html#dom-websocket-open
-      var socket = {readyState: 1};
+    it('initial connection.state is connecting and init handshake, if socket.readyState is OPEN', function() {
+      var send = false;
+      var socket = {
+        readyState: 1,
+        send: function(msg) {
+          expect(JSON.parse(msg).a).equal('hs');
+          send = true;
```
alecgibson: Setting flags like this feels a little weird. Can we please use a spy instead? Like:

```javascript
var socket = {
  readyState: 1,
  send: sinon.spy(),
};

expect(socket.send.calledOnce).to.be.true;
const message = JSON.parse(socket.send.getCall(0).args[0]);
expect(message.a).to.equal('hs');
```

dead-horse: fixed

alecgibson: 🙌 thanks!

alecgibson:
@ericyhwang this looks go to me — can you please sense-check?

ericyhwang:
Published in [sharedb@1.2.1](https://github.com/share/sharedb/releases/tag/v1.2.1)

PR #353, opened 2020-03-18
Define server plugin interface, move Agent queries code into a built-in plugin

ericyhwang:
We want to develop support for ShareDB plugins that can add support for new actions on top of ShareDB's functionality - https://github.com/share/sharedb/issues/337

This is one part of that work, adding the ability for the server `Agent#_handleMessage` to call message handling functions in registered plugins.

As a feasibility check, this PR moves the core query code from the server Agent class into a built-in server plugin, QueryServerPlugin, with no functionality changes. There are also no changes to client code in this PR.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -6,6 +6,9 @@ var ShareDBError = require('./error');
 
 var ERROR_CODE = ShareDBError.CODES;
 
+/** @typedef {import('./backend')} Backend */
+/** @typedef {import('./backend').ServerPlugin<unknown, unknown, unknown>} ServerPlugin */
+
 /**
  * Agent deserializes the wire protocol messages received from the stream and
  * calls the corresponding functions on its Agent. It uses the return values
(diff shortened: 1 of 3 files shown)


lib/agent.js
```diff
@@ -379,19 +362,22 @@ Agent.prototype._checkRequest = function(request) {
 // Handle an incoming message from the client
 Agent.prototype._handleMessage = function(request, callback) {
   try {
+    var plugin = this.actionToPlugin[request.a];
+
     var errMessage = this._checkRequest(request);
     if (errMessage) return callback(new ShareDBError(ERROR_CODE.ERR_MESSAGE_BADLY_FORMED, errMessage));
+    if (plugin) {
+      try {
+        plugin.checkRequest(request);
```
alecgibson: Maybe this method should be optional? Or do we want to enforce that all plugins validate requests?

lib/query-server-plugin.js
```diff
@@ -0,0 +1,229 @@
...
+function QueryServerPlugin() {
+  this.name = 'sharedb.query';
+  this.requestHandlers = {
+    qf: queryFetch,
+    qs: querySubscribe,
+    qu: queryUnsubscribe
+  };
+}
+
+QueryServerPlugin.prototype.close = function(callback) {
```
alecgibson: Should this also be optional?

PR #356, opened 2020-03-31
Fix presence re-subscribe after destroy

alecgibson:
This change fixes the following edge case:

  1. Create a `Presence` object
  2. Subscribe the presence
  3. Destroy the presence
  4. Create a new `Presence` object for the same channel on the same
     `Connection` object
  5. Attempt to re-subscribe
  6. `Agent` ignores the re-subscribe request, because the `seq` was
     reset to `1` on the new `Presence` object, but is still `2` on the
     `Agent`

We fix this by using a universal `_presenceSeq` counter on the
`Connection`, which is never reset, meaning that all presence messages
will have a strictly higher number than one another.

### DIFF
--- a/lib/client/connection.js
+++ b/lib/client/connection.js
@@ -57,6 +57,9 @@ function Connection(socket) {
   // A unique message number for the given id
   this.seq = 1;
 
+  // A unique message number for presence
+  this._presenceSeq = 1;
+
   // Equals agent.src on the server
   this.id = null;
 
--- a/lib/client/presence/presence.js
+++ b/lib/client/presence/presence.js
@@ -20,7 +20,6 @@ function Presence(connection, channel) {
   this.subscribed = false;
   this.remotePresences = {};
   this.localPresences = {};
-  this.seq = 1;
 
   this._remotePresenceInstances = {};
   this._subscriptionCallbacksBySeq = {};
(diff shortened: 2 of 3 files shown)


alecgibson:
No, `presenceSubscriptionSeq` is never cleared. It's kind of hard to know when it's safe to do so. The reason it's there is to figure out the correct target subscription state of the client, because subscribe and unsubscribe requests to the backend take some arbitrary amount of time.

For example, we **can't** clear on unsubscribe, because then it opens up this case:

  1. Client subscribes, sets `seq = 1`, waits for response
  2. Client unsubscribes, sets `seq = 2`, waits for response
  3. Client subscribes with `seq = 3`, but doesn't wait for response
  4. Client unsubscribes with `seq = 4`, and again doesn't wait for response
  5. Server receives and processes unsubscribe first (say), resetting `seq = null`
  6. Server receives and process the earlier subscribe setting `seq = 3`
  7. Client finishes subscribed, even though their desired state is unsubscribed

I'm all ears for an alternative approach. If we assume that all messages arrive in order (which I'm not sure we are? I always forget where we are with that), then we could actively send a reset request from the client when we call `presence.destroy()`, but if messages don't have guaranteed order we still have the problem of determining if later messages were issued before or after the destroy.

Issue #357, opened 2020-04-01
Synchronous `doc.unsubscribe` leaves me subscribed


alecgibson:
See test case here: https://github.com/share/sharedb/commit/1a3edc926523f45660530b5329db1fe00fd2e1cc

If I call:

```javascript
doc.subscribe();
doc.unsubscribe();
```

synchronously, then I'd expect to finish unsubscribed, but I finish subscribed.

ericyhwang:
From https://github.com/share/sharedb/pull/356:
> Ops do not have this issue since only one op (list of op components) is in-flight at once, with the rest held in pendingOps.
>
> Medium to long term, we're thinking of having doc sub/unsub locally queued too such that there's only one in-flight, which will fix the issue with the agent remaining subscribed while the client thinks it's not. Then perhaps doing the same for presence too.

alecgibson:
Okay running into some complication around fixing this. Mostly around connection state change. This is sort of a brain dump of my thinking so far, largely for organising my own thoughts, but also open to any suggestions.

# Summary

I think it boils down to a decision between:

  - keep requests true to the order in which they were requested (a "full flush"), even if it forces the client through some transient state that doesn't match their desired subscription state
  - try to keep subscription state as "current" as possible: the client has requested a state, and that's the state we should force ourselves to be in as soon as possible, even if it means a confusing order of callbacks, and callbacks being fired with an inconsistent state to the `Doc`

The key thing is that there can only ever be **one** inflight request at a time, so that client state stays in sync with server state.

# Example scenarios

## Stable connection

1. Connection established
1. Call `subscribe()` then `unsubscribe()`

I'd expect the following sequence of events to happen:
1. Send subscription request
1. Receive subscription callback
1. Send unsubscribe request
1. Receive unsubscribe callback

## Eager client

1. Start disconnected
1. Call `subscribe()` then `unsubscribe()`
1. Establish connection

Then we could do one of the below, but which?

### Full queue flush

Process the whole queue as in the "Stable connection" scenario:

1. Send subscription request
1. Receive subscription callback
1. Send unsubscribe request
1. Receive unsubscribe callback

### Short-circuit queue flush

We want to finish unsubscribed, which we already are, because we submitted both requests before establishing a connection and having the opportunity to subscribe. So

1. Trigger subscribe callback (without bothering to send the request)
2. Trigger unsubscribe callback (again, without sending the request)

[...]

PR #358, opened 2020-04-03
Only allow a single subscribe/unsubscribe request to be in-flight

alecgibson:
Fixes https://github.com/share/sharedb/issues/357

This change is a rework of our subscribe/unsubscribe logic. At the
moment, subscribe and unsubscribe messages are handled completely
separately, which means that we can hit race conditions when
simultaneous subscribe and unsubscribe requests are submitted.

This currently manifests itself in this bug:

  1. `subscribe()`
  2. `unsubscribe()` synchronously (ie before receiving the `subscribe`
     response
  3. Expect to finish unsubscribed. Indeed, `Doc` claims that I'm
     unsubscribed, but actually - because subscribe happens to take a
     bit longer on the server - I'm really still subscribed

This change reworks subscribe and unsubscribe requests into a queue:

  - there may only ever be a single in-flight subscribe or unsubscribe
    request
  - other requests are queued and only submitted once we complete the
    in-flight request

# Special cases

In order to maintain existing behaviour, we also handle a couple of
special cases:

  - if there is _only_ a single subscription request (inflight, or
    pending), and a duplicate request is submitted by the client, we
    just roll the callback into the existing request
  - if the next request is an unsubscribe, we will immediately call back
    if disconnected (but we no longer call them if there is a preceding
    subscribe waiting)

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -63,9 +63,9 @@ function Doc(connection, collection, id) {
 
   // Array of callbacks or nulls as placeholders
   this.inflightFetch = [];
-  this.inflightSubscribe = [];
-  this.inflightUnsubscribe = [];
+  this.inflightSubscribe = null;
   this.pendingFetch = [];
+  this.pendingSubscribe = [];
 
   // Whether we think we are subscribed on the server. Synchronously set to
   // false on calls to unsubscribe and disconnect. Should never be true when
@@ -229,9 +229,9 @@ Doc.prototype.hasPending = function() {
     this.inflightOp ||
     this.pendingOps.length ||
     this.inflightFetch.length ||
-    this.inflightSubscribe.length ||
-    this.inflightUnsubscribe.length ||
-    this.pendingFetch.length
+    this.inflightSubscribe ||
+    this.pendingFetch.length ||
+    this.pendingSubscribe.length
   );
 };
 
@@ -266,33 +266,39 @@ Doc.prototype._emitResponseError = function(err, callback) {
   this.emit('error', err);
 };
 
-Doc.prototype._handleFetch = function(err, snapshot) {
+Doc.prototype._handleFetch = function(error, snapshot) {
+  var callbacks = this.pendingFetch;
+  this.pendingFetch = [];
   var callback = this.inflightFetch.shift();
-  if (err) return this._emitResponseError(err, callback);
+  if (callback) callbacks.push(callback);
+  if (callbacks.length) {
+    callback = function(error) {
+      util.callEach(callbacks, error);
+    };
+  }
+  if (error) return this._emitResponseError(error, callback);
   this.ingestSnapshot(snapshot, callback);
   this._emitNothingPending();
 };
 
-Doc.prototype._handleSubscribe = function(err, snapshot) {
-  var callback = this.inflightSubscribe.shift();
-  if (err) return this._emitResponseError(err, callback);
-  // Indicate we are subscribed only if the client still wants to be. In the
-  // time since calling subscribe and receiving a response from the server,
-  // unsubscribe could have been called and we might already be unsubscribed
-  // but not have received the response. Also, because requests from the
-  // client are not serialized and may take different async time to process,
-  // it is possible that we could hear responses back in a different order
-  // from the order originally sent
-  if (this.wantSubscribe) this.subscribed = true;
-  this.ingestSnapshot(snapshot, callback);
+Doc.prototype._handleSubscribe = function(error, snapshot) {
+  var request = this.inflightSubscribe;
+  this.inflightSubscribe = null;
+  var callbacks = this.pendingFetch;
+  this.pendingFetch = [];
+  if (request.callback) callbacks.push(request.callback);
+  var callback;
+  if (callbacks.length) {
+    callback = function(error) {
+      util.callEach(callbacks, error);
+    };
+  }
+  if (error) return this._emitResponseError(error, callback);
+  this.subscribed = request.wantSubscribe;
+  if (this.subscribed) this.ingestSnapshot(snapshot, callback);
+  else if (callback) callback();
   this._emitNothingPending();
-};
-
-Doc.prototype._handleUnsubscribe = function(err) {
-  var callback = this.inflightUnsubscribe.shift();
-  if (err) return this._emitResponseError(err, callback);
-  if (callback) callback();
-  this._emitNothingPending();
+  this._flushSubscribe();
 };
 
 Doc.prototype._handleOp = function(err, message) {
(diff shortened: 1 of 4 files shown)


lib/client/doc.js
```diff
@@ -266,33 +266,39 @@ Doc.prototype._emitResponseError = function(err, callback) {
   this.emit('error', err);
 };
 
-Doc.prototype._handleFetch = function(err, snapshot) {
+Doc.prototype._handleFetch = function(error, snapshot) {
+  var callbacks = this.pendingFetch;
```
alecgibson, line 270: This is something I've done which I _think_ should be okay (and all the tests pass), but I'd like a sense check. The `pendingFetch` callbacks _used_ to be handled by `Doc._resubscribe`, where they'd be queued on the appropriate call.

However, I've sort of just "short-cutted" that behaviour, and now just grab these in both `_handleFetch` and `_handleSubscribe`. It seemed simpler, and I can't seem to think of any downsides...?

If I understand the semantics of this correctly, it's just an array of callbacks that should be called once we receive a snapshot, so grabbing them directly in the response handlers should be the same as trying to smoosh them into the `inflightFetch` callback?

alecgibson: The old behaviour was actually that each `fetch` would still trigger a message sent to the server, which would then only call a _single_ callback (`inflightFetch.shift()`), whereas we currently early-return all of them.

We should at the very least mimic the old behaviour (assume that all requests are responded to in order), or — better yet — store callbacks by `seq` like in Presence, and we can invoke the "correct" callback.

lib/client/doc.js
```diff
@@ -413,30 +410,66 @@ Doc.prototype.fetch = function(callback) {
...
+  this._queueSubscribe(wantSubscribe, callback);
+};
+
+Doc.prototype._queueSubscribe = function(wantSubscribe, callback) {
+  var nextRequest = this.inflightSubscribe || this.pendingSubscribe[0];
+  var hasOtherRequests = this.pendingSubscribe.some(function(request) {
+    return request !== nextRequest;
+  });
+  if (nextRequest && nextRequest.wantSubscribe === wantSubscribe && !hasOtherRequests) {
+    var callbacks = [];
```
alecgibson, line 430: Put this code block into a named function to clarify what we're doing

lib/client/doc.js
```diff
@@ -413,30 +410,66 @@ Doc.prototype.fetch = function(callback) {
...
+  var wantSubscribe = false;
+  this._queueSubscribe(wantSubscribe, callback);
+};
+
+Doc.prototype._queueSubscribe = function(wantSubscribe, callback) {
+  var nextRequest = this.inflightSubscribe || this.pendingSubscribe[0];
+  var hasOtherRequests = this.pendingSubscribe.some(function(request) {
+    return request !== nextRequest;
+  });
+  if (nextRequest && nextRequest.wantSubscribe === wantSubscribe && !hasOtherRequests) {
```
alecgibson, line 429: Think about a good variable name for this condition?

lib/client/doc.js
```diff
@@ -413,30 +410,66 @@ Doc.prototype.fetch = function(callback) {
...
+      // so let's just set ourselves to unsubscribed now.
+      this.subscribed = false;
+      this.connection.sendUnsubscribe(this);
+    }
+
     return;
   }
-  if (callback) process.nextTick(callback);
+
+  if (!this.pendingSubscribe[0].wantSubscribe) {
```
ericyhwang, line 466: Would be nice to add a short comment explaining this case - if the client offline and requests to unsubscribe, immediately call back since offline means it's not subscribed.

lib/client/doc.js
ericyhwang: Per our investigation, old behavior was that, if the connection is online, a reply to fetch will only call the earliest-queued callback in the non-bulk case.

lib/client/doc.js
alecgibson: Also instead let's check the _last_ request here instead of the first, and then we can de-duplicate eg:

- subscribe
- unsubscribe
- subscribe
- subscribe (<-- should be combined with previous subscribe)

alecgibson: (Also should be no need to check if we have other requests)

ericyhwang, line 429: This will only coalesce if there's a single next request. As discussed, it should be OK to coalesce in the case of `s -> us -> s -> s`, by using the latest request instead of the next request. Then we don't need to check for other requests, either.

lib/client/doc.js
alecgibson: @ericyhwang I've thought about it again, and I think that actually the refactored behaviour _is_ still correct.

`pendingFetch` is only ever dealt with when we go offline. Here's how the old behaviour worked:

## Always online

We never touch `pendingFetch`.

## Subscribe whilst offline

- Start offline
- Try to subscribe
- Callback pushed onto `pendingFetch`
- Connection opened
- Call `_resubscribe()`
- Pending subscribe callbacks all combined and used as a single callback for a single subscribe request (or a fetch, if we don't want to be subscribed)

## Lose connection waiting for subscription

- Start online
- Send subscribe request
- Callback pushed onto `inflightSubscribe`
- Connection closed

lib/client/doc.js
```diff
@@ -451,6 +477,14 @@ function pushActionCallback(inflight, isDuplicate, callback) {
   }
 }
 
+function combineCallbacks(callbacks) {
+  callbacks = callbacks.filter(Boolean);
```
ericyhwang, line 481: This is to filter out null/undefined items in `callbacks`?

lib/client/doc.js
```diff
@@ -413,30 +410,59 @@ Doc.prototype.fetch = function(callback) {
...
-  this.subscribed = false;
+  var wantSubscribe = false;
+  this._queueSubscribe(wantSubscribe, callback);
+};
+
+Doc.prototype._queueSubscribe = function(wantSubscribe, callback) {
+  var lastRequest = this.pendingSubscribe[this.pendingSubscribe.length - 1] || this.inflightSubscribe;
+  var isDuplicateRequest = lastRequest && lastRequest.wantSubscribe === wantSubscribe;
+  if (isDuplicateRequest) {
+    lastRequest.callback = combineCallbacks([lastRequest.callback, callback]);
```
ericyhwang, line 428: 👍 to the improvement in readability

lib/client/doc.js
alecgibson: Yes. I can change to a `function` if you'd prefer. I tend to use this shorthand a lot personally (and I especially prefer it when arrow functions aren't available), but I can see why it can be confusing if you haven't come across it before. It's basically just abusing the fact that:

```javascript
Boolean(null) // false
Boolean(undefined) // false
Boolean(function() {}) // true
```

Could also just make some sort of util function called `truthy(arg) { return !!arg }` and use that for improved readability.

PR #359, opened 2020-04-03
Fix `Presence._callEachOrEmit`

alecgibson:
At the moment, there's a bug where we:

  1. Start disconnected
  2. Submit a presence subscribe request with **no callback**
  3. Connect
  4. The presence then subscribes and calls each callback, but doesn't
     check whether the callback is truthy, which throws

We could check callbacks when we first assign them to the internal map,
but it's quite nice for debugging to see an `undefined` callback
attached to a `seq`.

Therefore, we should handle callbacks not existing when actually
calling them, which `Doc` already does.

This change moves the `Doc` `callEach` method into `utils`, where we can
also use it in `Presence`.


alecgibson:
(I think the code coverage decrease is because I moved lines of code out of `doc.js`)

Issue #362, opened 2020-04-16
Is the textarea example working?


danijar:
It's not quite clear to me what the textarea example is supposed to do. Is it supposed to allow collaborative editing of the text at the same time? Running it locally, it shows a textarea but there is no synchronization between tabs and it doesn't remember the text on page reload, either. Those both work for the quill rich-text example.

PR #363, opened 2020-04-16
Fix textarea example

alecgibson:
Fixes https://github.com/share/sharedb/issues/362

The textarea example currently doesn't work, because it overrides a
bunch of our WebSocket event handler functions.

This change fixes that by attaching these methods as listeners, rather
than overriding the handler functions.

### DIFF
--- a/examples/textarea/client.js
+++ b/examples/textarea/client.js
@@ -11,20 +11,20 @@ var statusSpan = document.getElementById('status-span');
 statusSpan.innerHTML = 'Not Connected';
 
 element.style.backgroundColor = 'gray';
-socket.onopen = function() {
+socket.addEventListener('open', function() {
   statusSpan.innerHTML = 'Connected';
   element.style.backgroundColor = 'white';
-};
+});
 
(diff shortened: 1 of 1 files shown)


curran:
Nice fix! Thank you @alecgibson

examples/textarea/client.js
```diff
@@ -11,17 +11,23 @@ var statusSpan = document.getElementById('status-span');
 statusSpan.innerHTML = 'Not Connected';
 
 element.style.backgroundColor = 'gray';
+var onopen = socket.onopen;
 socket.onopen = function() {
```
ericyhwang, line 15: It would be cleaner to use the [WebSocket events](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket#Events) here, since these aren't intending to override behavior, they're just looking to add additional UI when something happens.
```js
socket.addEventListener('open', ...);
```

alecgibson: Ha sorry old IE habit

PR #367, opened 2020-05-13
Run `sharedb-mongo` tests

alecgibson:
This change runs the [`sharedb-mongo`][1] test suite as part of the
Travis build, since `sharedb-mongo` pulls some tests directly from this
library.

[1]: https://github.com/share/sharedb-mongo

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 language: node_js
+services:
+  - mongodb
 node_js:
   - "12"
   - "10"
   - "8"
 script: "npm run lint && npm run test-cover"
 # Send coverage data to Coveralls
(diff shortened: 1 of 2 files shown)


alecgibson:
@ericyhwang this is the build failing correctly: https://travis-ci.org/github/share/sharedb/jobs/686670802

(I just didn't `npm install` for `sharedb-mongo`, and it exited with a non-zero code)

alecgibson:
Also this is a successful failure before rebasing onto the fixed tests: https://travis-ci.org/github/share/sharedb/jobs/686673709

So I'm now confident that we'll correctly fail the build if we break something that breaks `sharedb-mongo`

Issue #373, opened 2020-06-16
Composing the last pending Op and the submitted Op can result in `null`


antoinelyset:
Hi!

While testing ShareDB with the JSON1 OT, I encountered an interesting bug. When we compose the last pending Op and the submitted Op before sending it to the server, it can result in `null` (with JSON1 at least).

https://github.com/share/sharedb/blob/cea06f8b2e6a4553eda9146782cddd24481f71dd/lib/client/doc.js#L808-L813

This raises a lot of different problems and crashes because `op.op` is falsy.

We don't send the actual Op (`null`) see :
https://github.com/share/sharedb/blob/cea06f8b2e6a4553eda9146782cddd24481f71dd/lib/client/connection.js#L455-L463

Here we return `undefined`: 

https://github.com/share/sharedb/blob/cea06f8b2e6a4553eda9146782cddd24481f71dd/lib/agent.js#L824-L833

And maybe more.

I have the resources to fix this bug but I wanted to know your opinion on how to fix it first.

Should we prevent sending it client-side? If yes, how can we express a value as skippable? (Should it be defined type-wise? Should it be a definable function called before sending the value?)

Should we just send `null` as a regular value and modify every call site to something like `if (op.op !== undefined) `?

Thanks for reading me.

alecgibson:
I haven't personally touched much of the OT stuff, but my first thought is that if we don't already handle this case, then a falsy return from `type.compose` is invalid, and should be fixed in the `json1` library. The [docs](https://github.com/ottypes/docs#optional-properties) aren't particularly thorough, but they do mention this of `type.compose`:

> compose(op1, op2) -> op: (optional) Compose op1 and op2 to produce a new operation

I don't think it's crazy to assume that `compose` should always return a valid op (even in the event of a no-op).

That being said, I _also_ don't think it's crazy to handle a `null` no-op. I'd start with raising this in the `json1` library, and checking that a `null` return from `compose` is expected.

@josephg maintains both `json1` and the `docs` libraries, so might have more insights?

curran:
It would be interesting to put together some unit tests for this case for both `json1` (your failing case) and the equivalent ops in `json0`.

antoinelyset:
I don't know if it's only related to `json1`, we could imagine a compose that returns `false` or `[]`.

Let's wait for @josephg answer then.

josephg:
[...]

PR #374, opened 2020-06-18
Allow an op to be falsy

antoinelyset:
Hi!

This is a fix for https://github.com/share/sharedb/issues/373

I chose to use `'op' in op` as the check to know if we are doing an edit.

I've added a simple test to verify that it works (it was actually really useful to find all the places I needed to change).

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -257,7 +257,7 @@ Agent.prototype._sendOp = function(collection, id, op) {
     src: op.src,
     seq: op.seq
   };
-  if (op.op) message.op = op.op;
+  if ('op' in op) message.op = op.op;
   if (op.create) message.create = op.create;
   if (op.del) message.del = true;
 
@@ -826,7 +826,7 @@ function createClientOp(request, clientId) {
   // such as a resubmission after a reconnect, but it usually isn't needed
   var src = request.src || clientId;
   // c, d, and m arguments are intentionally undefined. These are set later
-  return (request.op) ? new EditOp(src, request.seq, request.v, request.op) :
+  return ('op' in request) ? new EditOp(src, request.seq, request.v, request.op) :
     (request.create) ? new CreateOp(src, request.seq, request.v, request.create) :
       (request.del) ? new DeleteOp(src, request.seq, request.v, request.del) :
         undefined;
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -528,7 +528,7 @@ function transformX(client, server) {
   }
 
   // Ignore no-op coming from server
-  if (!server.op) return;
+  if (!('op' in server)) return;
 
   // I believe that this should not occur, but check just in case
   if (client.create) {
@@ -567,7 +567,7 @@ function transformX(client, server) {
  * @private
  */
 Doc.prototype._otApply = function(op, source) {
-  if (op.op) {
+  if ('op' in op) {
     if (!this.type) {
       // Throw here, because all usage of _otApply should be wrapped with a try/catch
       throw new ShareDBError(
@@ -713,7 +713,7 @@ Doc.prototype._submit = function(op, source, callback) {
   if (!source) source = true;
 
   // The op contains either op, create, delete, or none of the above (a no-op).
-  if (op.op) {
+  if ('op' in op) {
     if (!this.type) {
       var err = new ShareDBError(
         ERROR_CODE.ERR_DOC_DOES_NOT_EXIST,
@@ -800,14 +800,14 @@ Doc.prototype._tryCompose = function(op) {
 
   // Compose an op into a create by applying it. This effectively makes the op
   // invisible, as if the document were created including the op originally
-  if (last.create && op.op) {
+  if (last.create && 'op' in op) {
     last.create.data = this.type.apply(last.create.data, op.op);
     return last;
   }
 
   // Compose two ops into a single op if supported by the type. Types that
   // support compose must be able to compose any two ops together
-  if (last.op && op.op && this.type.compose) {
+  if ('op' in last && 'op' in op && this.type.compose) {
     last.op = this.type.compose(last.op, op.op);
     return last;
   }
@@ -925,7 +925,7 @@ Doc.prototype._rollback = function(err) {
   // working state, then call back
   var op = this.inflightOp;
 
-  if (op.op && op.type.invert) {
+  if ('op' in op && op.type.invert) {
     op.op = op.type.invert(op.op);
 
     // Transform the undo operation by any pending ops.
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -504,7 +504,7 @@ Doc.prototype.flush = function() {
 
 // Helper function to set op to contain a no-op.
 function setNoOp(op) {
-  op.op = null;
+  delete op.op;
   delete op.create;
   delete op.del;
 }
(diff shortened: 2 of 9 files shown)


lib/agent.js
```diff
@@ -257,7 +257,7 @@ Agent.prototype._sendOp = function(collection, id, op) {
     src: op.src,
     seq: op.seq
   };
-  if (op.op) message.op = op.op;
+  if ('op' in op) message.op = op.op;
```
curran, line 260: Wait, so `op` here can be `null`, right?

But, `in` doesn't work on `null`...

image

In any case, since the solution will need to be duplicated so many places across the codebase, it might make sense to introduce a helper function, like `hasOp(op)`, so that the actual checking logic could be modified in just one place.

antoinelyset: Here `op` will always be an object I think. (The old condition was based on this too). So more

Screenshot 2020-06-18 at 16 43 51

Calyhre: Yep, @antoinelyset is right here, `op` is an object. It's `op.op` in this code that might be `null`

lib/client/doc.js
```diff
@@ -528,7 +528,7 @@ function transformX(client, server) {
   }
 
   // Ignore no-op coming from server
-  if (!server.op) return;
+  if (!('op' in server)) return;
```
alecgibson, line 531: As I mentioned in the issue, can we please update `setNoOp` to return a no-op consistent with this new shape (ie `{op: null}`), and then keep this check as it was before.

test/client/submit.js
```diff
@@ -1196,5 +1198,34 @@ module.exports = function() {
         });
       });
     });
+
+    describe('type.json1', function() {
```
alecgibson, line 1202: Can we please move json1 tests into their own file for neatness?

test/client/submit.js
```diff
@@ -1196,5 +1198,34 @@ module.exports = function() {
...
+      it('ops submitted sync get composed even if the composition returns null', function(done) {
+        var doc = this.backend.connect().get('dogs', 'fido');
+        doc.create({age: 3}, json1Type.type.uri);
+        doc.submitOp(json1Type.insertOp(['color'], 'gold'));
+        doc.submitOp(json1Type.removeOp(['color']), function(err) {
+          if (err) return done(err);
+          expect(doc.data).eql({age: 3});
+          // Version is 1 instead of 3, because the create and ops got composed
+          expect(doc.version).eql(1);
```
alecgibson, line 1211: Can we break the `create` case into its own test?

test/client/submit.js
```diff
@@ -1196,5 +1198,34 @@ module.exports = function() {
...
+            if (err) return done(err);
+            expect(doc.data).eql({age: 3});
+            // Ops get composed
+            expect(doc.version).eql(2);
+            doc.submitOp(json1Type.insertOp(['size'], 1));
+            doc.del(function(err) {
+              if (err) return done(err);
+              expect(doc.data).eql(undefined);
+              // del DOES NOT get composed
+              expect(doc.version).eql(4);
```
alecgibson, line 1223: Again, separate test, please.

lib/client/doc.js
antoinelyset: Sorry, I'm a bit confused by this. Right now `setNoOp` `delete` the `op` key from the `op`. You want it to be updated to?

```javascript
function setNoOp(op) {
  op.op = null
  delete op.create;
  delete op.del;
}
```

alecgibson: Yes, please. Just so that a no-op looks the same everywhere.

antoinelyset: Oh okay. I try this change (and putting back `if (!server.op)`) and I've got some failing specs. I think it comes from the fact that I need to update this check in `_otApply` : 

https://github.com/antoinelyset/sharedb/blob/28f1faa7818d5af0ce9f610262a180f6651f6f61/lib/client/doc.js#L570

If I don't, `length` will fail here: 

https://github.com/sharedb/sharedb/blob/28f1faa7818d5af0ce9f610262a180f6651f6f61/lib/client/doc.js#L594

Right? But now we're back to the problem of `null` being skipped?

I'm a bit lost by this `setNoOp`, `null` is only a NoOp in JSON1 and ShareDB should not know this and continue as if it was a regular operation.

alecgibson: That second link is broken but, I assume you meant to point [here](https://github.com/antoinelyset/sharedb/blob/28f1faa7818d5af0ce9f610262a180f6651f6f61/lib/client/doc.js#L594)? `length` shouldn't fail, because of the `defaultType` check (where the `defaultType` is always `json0`).

antoinelyset: Yes sorry, it points here. the `op` can be a noOp and still be in `json0` no? This is what I have in the specs.

antoinelyset: I've pushed my branch, if you want to take a look. I think I didn't quite understand.

alecgibson: I think it's fine to update the check and just skip that block if `op.op === null`. All the block is doing is fragmenting the `json0` op and applying it one component at a time. Given that a no-op has no components, it's fine to just skip, and the no-op will be "applied" [below](https://github.com/antoinelyset/sharedb/blob/28f1faa7818d5af0ce9f610262a180f6651f6f61/lib/client/doc.js#L621).

antoinelyset: Hmm I've update `_otApply` check for a `op.op === null` but now I've got a lot of other tests failing. I'm missing something again I think.

antoinelyset:
I've converted it to a draft. I will work on the other tasks (adding more tests related to the JSON1 format and to the server side)

antoinelyset:
Ok, so I've added a `submit-json1.js` test file, split the tests, and added one for server-side composition.

I've put back the change with `delete op.op` not because I'm stubborn (I can be sometimes hahaha), but because I don't quite know how to do it with `op.op = null`.

josephg:
We could add a isNoOp method in the json1, json0, etc and use that. They’re trivial methods to implement.

antoinelyset:
That could be interesting indeed. It could be done in another PR right?

curran:
Love it. `isNoOp` Sounds like a great thing.

alecgibson:
@antoinelyset what issues were you having using `op.op = null` instead of `delete op.op`? I think it's best to keep our "shape" of a no-op as consistent as possible.

antoinelyset:
I encountered a lot of errors linked to timeout and un-equalities. I think I didn't update enough places. (you can take a look here https://github.com/share/sharedb/pull/374/commits/ea40a735b30525e496bc23da4d027ad6f26b2cd7)

The way I think about it is that the no-op shape for ShareDB should be `delete op.op` . `op.op = null` is a no-op in JSON1 but it should be seen as a regular op by ShareDB. So we should keep the `delete op.op` and keep the updated checks `'op' in 'op'`.

If we want to avoid sending `op.op === null` in JSON1, we should implement a per type `isNoOp` check (`null` could be a valid and useful op in another type).

alecgibson:
@antoinelyset makes sense. We can always change it later if we need. The one other thing we mentioned in the meeting is that it's a bit weird that we check `'op' in op` but not `'create' in op` (I guess in theory there could be a no-op create...?), but that's probably just me nit-picking, so let's go ahead and merge.

Thanks for your work on this! 👏

curran:
This is great!

@alecgibson FWIW, the last commit in `master` seems to have broken CI.

image

Only Node 10.

image

Looks like it's a new test that's failing.

image

alecgibson:
Hopefully fixed in https://github.com/share/sharedb/pull/376

PR #384, opened 2020-07-23
Held ops

alecgibson:
# Introduction

This change adds support for a "held op". Held ops are ops which are:

 - Only applied locally
 - Not necessarily submitted to the server
 - May be submtted to the server at some later time
 - Will be lost at the end of a "session" if not flushed (since they
   haven't been submitted to the server)

# Motivation

Held ops allow for some level of disconnect with the server, whilst
keeping the `Doc` "live". The primary use case this was designed for is
for when a `Doc` is in a "confirming" state: that is, some change has
been made to the `Doc`, but we're waiting for some confirmation that
this change should be comitted to the server (eg user input).

This sort of thing can be achieved with `pause()` and `resume()`, but:

 - Pausing a document will stop all ops from being submitted, even if
   those other ops don't need confirmation
 - Resuming a `Doc` is not fool-proof. It would be a much nicer failure
   mode to only fail to commit the changes that require confirmation,
   rather than all subsequent changes

We could try to address these issues by having a side-channel: a second
`Connection` which we funnel all ops that need confirmation, which we
can safely pause, but this has its own issues:

 - Can't flush individual ops (unless we keep opening more
   side-channels)
 - We now have two clients to handle, which can cause confusion around
   how to merge data, and what constitutes a "local" op (ie everything
   submitted down the pauseable connection, will look remote to the
   mainline channel, even though it came from our own machine)

Held ops attempts to solve this issue by providing a mechanism for
submitting everything through a single client, and holding ops that can
be individually flushed to the server.

Note that one other possible (ab)use for held ops is to simply have
local-only data, which benefits from type transformations, although it
should be possible to achieve something like that with Presence.

# API

We add a new method for submitting a held op:

```js
var flush = doc.submitHeldOp(op, options);
```

Where `op` and `options` are the same as for `submitOp`. We could
potentially add a flag to `options`, like `{held: true}`, and just use
`submitOp`, but `submitHeldOp` returns a value, which `submitOp` does
not.

The returned value `flush` is a function that takes a callback:

```js
flush(function(error) {})
```

This method will flush that particular held op to the server, and then
call the provided callback on acknowledgement, as if we'd used
`submitOp`.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -88,6 +88,9 @@ function Doc(connection, collection, id) {
   // This is a list of {[create:{...}], [del:true], [op:...], callbacks:[...]}
   this.pendingOps = [];
 
+  this._heldOpId = 1;
+  this._heldOps = {};
+
   // The OT type of this document. An uncreated document has type `null`
   this.type = null;
 
@@ -342,13 +345,15 @@ Doc.prototype._handleOp = function(err, message) {
     return;
   }
 
-  if (this.inflightOp) {
-    var transformErr = transformX(this.inflightOp, message);
-    if (transformErr) return this._hardRollback(transformErr);
+  var opsToTransform = [];
+  if (this.inflightOp) opsToTransform.push(this.inflightOp);
+  opsToTransform = opsToTransform.concat(this.pendingOps);
+  for (var id in this._heldOps) {
+    opsToTransform.push(this._heldOps[id]);
   }
 
-  for (var i = 0; i < this.pendingOps.length; i++) {
-    var transformErr = transformX(this.pendingOps[i], message);
+  for (var i = 0; i < opsToTransform.length; i++) {
+    var transformErr = transformX(opsToTransform[i], message);
     if (transformErr) return this._hardRollback(transformErr);
   }
 
@@ -712,18 +717,9 @@ Doc.prototype._submit = function(op, source, callback) {
   // Locally submitted ops must always have a truthy source
   if (!source) source = true;
 
-  // The op contains either op, create, delete, or none of the above (a no-op).
   if ('op' in op) {
-    if (!this.type) {
-      var err = new ShareDBError(
-        ERROR_CODE.ERR_DOC_DOES_NOT_EXIST,
-        'Cannot submit op. Document has not been created. ' + this.collection + '.' + this.id
-      );
-      if (callback) return callback(err);
-      return this.emit('error', err);
-    }
-    // Try to normalize the op. This removes trailing skip:0's and things like that.
-    if (this.type.normalize) op.op = this.type.normalize(op.op);
+    var error = this._normalizeOp(op);
+    if (error) return this._callbackOrEmit(error, callback);
   }
 
   try {
@@ -739,6 +735,19 @@ Doc.prototype._submit = function(op, source, callback) {
   process.nextTick(function() {
     doc.flush();
   });
+};
+
+// Try to normalize the op. This removes trailing skip:0's and things like that.
+Doc.prototype._normalizeOp = function(op) {
+  if (!this.type) {
+    var err = new ShareDBError(
+      ERROR_CODE.ERR_DOC_DOES_NOT_EXIST,
+      'Cannot submit op. Document has not been created. ' + this.collection + '.' + this.id
+    );
+    return err;
+  }
+
+  if (this.type.normalize) op.op = this.type.normalize(op.op);
 };
 
 Doc.prototype._pushOp = function(op, callback) {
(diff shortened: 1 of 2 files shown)


lib/client/doc.js
```diff
@@ -88,6 +88,9 @@ function Doc(connection, collection, id) {
   // This is a list of {[create:{...}], [del:true], [op:...], callbacks:[...]}
   this.pendingOps = [];
 
+  this._heldOpId = 1;
+  this._heldOps = {};
```
alecgibson, line 92: I know that the other props on this class aren't prefixed, but I'd like to keep things "private" where possible

lib/client/doc.js
```diff
@@ -342,13 +345,15 @@ Doc.prototype._handleOp = function(err, message) {
     return;
   }
 
-  if (this.inflightOp) {
-    var transformErr = transformX(this.inflightOp, message);
-    if (transformErr) return this._hardRollback(transformErr);
+  var opsToTransform = [];
+  if (this.inflightOp) opsToTransform.push(this.inflightOp);
+  opsToTransform = opsToTransform.concat(this.pendingOps);
+  for (var id in this._heldOps) {
```
alecgibson, line 351: This is where my understanding of OT starts to get pushed a bit — does the transformation ordering matter here? I'd have normally assumed no, but I think `transformX` mutates the server op, so order **does** matter (I confirmed this trivially by moving the inflight op to after pending, and a couple of tests broke). I just don't know how (because held ops are temporally separate from pending ops). It could be that it's only important for `create` and `del`?

lib/client/doc.js
```diff
@@ -741,6 +737,19 @@ Doc.prototype._submit = function(op, source, callback) {
   });
 };
 
+// Try to normalize the op. This removes trailing skip:0's and things like that.
+Doc.prototype._normalizeOp = function(op) {
+  if (!this.type) {
```
alecgibson, line 742: This isn't really "normalising", but it's common code. Open to naming improvements on the method.

lib/client/doc.js
```diff
@@ -760,6 +769,12 @@ Doc.prototype._pushOp = function(op, callback) {
   op.type = this.type;
   op.callbacks = [callback];
   this.pendingOps.push(op);
+
+  if (op.del) this._heldOps = {};
+  for (var id in this._heldOps) {
+    var transformErr = transformX(op, this._heldOps[id]);
```
alecgibson, line 775: Again, my understanding of `transformX` isn't great. It's going to mutate `op`. But we're at the end of the function, so I assume that's okay?

test/client/doc.js
```diff
@@ -418,4 +421,151 @@ describe('Doc', function() {
...
+      flush(errorHandler(done));
+      var count = 0;
+      doc2.on('op', function() {
+        if (++count < 2) return;
+        expect(doc2.data.title).to.equal('Nineteen Eighty-four: A Novel');
+        done();
+      });
+    });
+
+    it('agrees with a remote client about a conflicting local insert', function(done) {
```
alecgibson, line 495: Should also add a test case where the submit op has an index higher than the held op. This currently fails, because the submitted op should be transformed against held ops when being submitted, but they aren't.

alecgibson:
I've tried looking at this again, but I think I've hit a snag where we need the type to be invertible. Consider this case:

1. Start with document `{text: 'Hi!'}`
2. Submit held op: `[{p: ['text', 0], sd: 'Hi'}, {p: ['text', 0], si: 'Hello}]`
3. Submit (non-held) op: `[{p: ['text', 5], si: ' world'}]`
4. Local document reads `Hello world!`
5. Remote document reads `Hi! world` (or potentially errors, given that `5` is longer than the string length?)

This is because we'd like the client to really send `[{p: ['text', 2], si: ' world'}]` — that is, the op as if the held op hadn't been applied.

However, as far as I'm aware, this is only achievable by transforming our submitted op by the **inverse** of the held op (or another way of thinking it might be to "untransform" by the op). 

Only supporting invertible types is a definite shame (and on a personal note, wouldn't be acceptable for our own use case).

The only solution I can think of is pretty horrible: force consumers to call `submitOp` as if they're submitting to the remote doc (ie they ignore their own held ops), and then we can correctly transform the submitted op when applying locally. I don't know if I'm missing something better and cleverer?

curran:
This is cool. Related to https://github.com/datavis-tech/vizhub-issue-tracker/issues/445

alecgibson:
As discussed in today's meeting, I'm going to close this for now. I don't have a lot of reason to work on it if it doesn't support non-invertible types, and it means we'd probably have code lurking that nobody's using.

If anyone's particularly keen to add this feature, please feel free to pick up the work!

PR #391, opened 2020-09-18
fix: replace action name with backend.MIDDLEWARE_ACTIONS

akumatus:
Since `backend.MIDDLEWARE_ACTIONS` is defined in `Backend.prototype`, it's better to use it as an action name.


akumatus:
@coveralls

alecgibson (approved):
Seems sensible to me.

NB: build failure is because of https://github.com/share/sharedb-mongo/issues/90

Issue #396, opened 2020-10-13
receive op  data had split more times with not local？


aeroyu:
i submit op like 
[
  {
    "sd": "checkable-list-item",
    "p": [
      "blocks",
      0,
      "type",
      0
    ]
  },
  {
    "si": "unstyled",
    "p": [
      "blocks",
      0,
      "type",
      0
    ]
  }
]
and local receive op event op is the same with send data，
but with other client receive op，has split two time，[
  {
    "sd": "checkable-list-item",
    "p": [
      "blocks",
      0,
      "type",
      0
    ]
  }] and [{
    "si": "unstyled",
    "p": [
      "blocks",
      0,
      "type",
      0
    ]
  }
]

how can i receive op normal

alecgibson:
Could you please elaborate on your use case? Why is it important to have these delivered as a single object?

aeroyu:
> Could you please elaborate on your use case? Why is it important to have these delivered as a single object?

i use sharedb with draft js，when the op is split and setState，will not work，ex `[
  {
    "sd": "unstyled",
    "p": [
      "blocks",
      0,
      "type",
      0
    ]
  },
  {
    "si": "checkable-list-item",
    "p": [
      "blocks",
      0,
      "type",
      0
    ]
  }
]` 
the first op （delete type unstyled） merge into local ，but it is not my expect，because ，draft js block type can not be empty，and will reset type to unstyled，after apply the second op , the type result is “checkable-list-itemunstyled”

aeroyu:
> Could you please elaborate on your use case? Why is it important to have these delivered as a single object?

i have solve it，i solution is set the uuid and op_size to op model，when receive by other side，i get uuid and op_size to merge the ops  and do one setState，it work！any best  solution？

alecgibson:
Glad to hear you solved it. Honestly off the top of my head I'm not sure there's any (easy) way around it. If you're interested, here's the code that's responsible for that behaviour: https://github.com/share/sharedb/blob/c711cfcb777213d193b1f4a101125e8f6e8e6864/lib/client/doc.js#L584-L620

You'll see it performs this check:

```js
if (!source && this.type === types.defaultType && op.op.length > 1)
```

and then "shatters" the op.

So one potential way around this would be to register second `json0` type, which _isn't_ the default type, and use that, but it's pretty hacky.

You could also try switching to the [`json1`][1] type, which seems to have more semantic operations (including replace).

[...]

PR #401, opened 2020-11-04
Add batch op events

alecgibson:
This change adds two events:

 - `before op batch` - fired once before any set of ops is applied.
 - `op batch` - fired once after any set of ops is applied. This may
   follow multiple `op` events if the op had multiple `json0` components

This is a non-breaking change that should allow clients to process ops
in their entirety.

There has already been some discussion around this:

 - https://github.com/share/sharedb/pull/129
 - https://github.com/share/sharedb/issues/396

This is a much simpler approach than the existing pull request. Here we
try not to change existing behaviour, and only add new, non-breaking
events.

Motivation for such an event would include clients applying some form of
validation logic, which doesn't make sense if an op is shattered.

For example, consider a client that wants to ensure a field
`mustBePresent` is always populated:

```js
doc.on('op', () => {
  if (!doc.data.mustBePresent) throw new Error('invalid');
});

remoteDoc.submitOp([
  {p: ['mustBePresent', 0], sd: 'existing value'},
  {p: ['mustBePresent', 0], si: 'new value'},
]);
```

In the above example, the submitted op is clearly attempting to perform
a replacement. However, the receiving `doc` only receives this
replacement in parts, so it looks like the document reaches an invalid
state, when actually the submitted op is perfectly valid.

In this case we `throw`, but we could have also attempted to populate
with a default value, which could interfere with the desired value.

This change fixes the above issue, because now we can just listen for
the `op batch` event, and consider the document once all the components
of a given op have been applied.

### DIFF
--- a/README.md
+++ b/README.md
@@ -364,6 +364,12 @@ An operation is about to be applied to the data. `source` will be `false` for op
 `doc.on('op', function(op, source) {...})`
 An operation was applied to the data. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
 
+`doc.on('before op batch'), function(op, source) {...})`
+A potentially multi-part operation is about to be applied to the data. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
+
+`doc.on('op batch'), function(op, source) {...})`
+A potentially multi-part operation was applied to the data. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
+
 `doc.on('del', function(data, source) {...})`
 The document was deleted. Document contents before deletion are passed in as an argument. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
 
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -581,6 +581,10 @@ Doc.prototype._otApply = function(op, source) {
       );
     }
 
+    // NB: If we need to add another argument to this event, we should consider
+    // the fact that the 'op' event has op.src as its 3rd argument
+    this.emit('before op batch', op.op, source);
+
     // Iteratively apply multi-component remote operations and rollback ops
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -364,6 +364,12 @@ An operation is about to be applied to the data. `source` will be `false` for op
 `doc.on('op', function(op, source) {...})`
 An operation was applied to the data. `source` will be `false` for ops received from the server and defaults to `true` for ops generated locally.
 
+`doc.on('before op batch'), function(op, source) {...})`
```
alecgibson, line 367: As usual, naming is definitely up for debate. I'd have preferred to follow what was proposed in the [other PR](https://github.com/share/sharedb/pull/129), and rename the existing events to something like `op component`, but that would have been breaking.

alecgibson:
Potential things to discuss:

- Do we emit this for _every_ op, or just ops that _will_ be shattered? My gut tells me that the "simpler" approach is better (ie fire for everything), so that clients can just react to `batch op` if they just know they always want the atomic change. They can use heuristics if needed to determine if it's about to be shattered (or we could pass as an argument to the event?)
- Are we happy with keeping the event arguments vaguely the same as the `op` event? I've omitted the `op.src` for now, because it's easier to be conservative, but we could add it for consistency.
- Do we want to hide this behind a flag? I kind of feel like a flag is pointless; if you're not subscribed to the events, then nothing happens, and by subscribing to the events you're signalling that you actively want to know anyway.
- We could potentially avoid adding an extra event, and add a flag that signals whether you want the shattering behaviour instead (again, this was proposed in the other PR). This feels a little "fiddlier" to me, but it's certainly an option, and can similarly be made in a non-breaking way.

ericyhwang (approved):
Looks good as-is! We all discussed the questions and think this is the cleanest approach without having to make a breaking change.

PR #408, opened 2020-11-16
Send client handshake when server binds

alecgibson:
At the moment, it's possible for the server to miss the client's `hs`
handshake message. Since it never gets the handshake, it never sends one
back, and the client remains uninitialised.

This can happen when performing asynchronous, long-running tasks between
opening the websocket, and calling `backend.listen` to bind the socket
to ShareDB.

This change fixes this issue by having the client send the handshake
message twice:

 - Once when the `Connection` is initialised. If the server is ready,
   but the client connected late, we'll still handshake.
 - Once when the client receives the server's `init` message. If the
   server bound to the socket late, and missed the first handshake
   message, this prompts us to send it again.

### DIFF
--- a/lib/client/connection.js
+++ b/lib/client/connection.js
@@ -705,9 +705,11 @@ Connection.prototype._handleSnapshotFetch = function(error, message) {
 };
 
 Connection.prototype._handleLegacyInit = function(message) {
-  // If the minor protocol version has been set, we can ignore this legacy
-  // init message, and wait for a response to our handshake message.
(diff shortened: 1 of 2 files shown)


ericyhwang (approved):

alecgibson:
@ericyhwang this PR "broke" a test on Node 14. More details in the commit message. Could you please sense-check?

ericyhwang (approved):
Still LGTM

PR #409, opened 2020-11-16
Move to Github Actions

alecgibson:
The build on Travis has been pretty slow lately. This change moves our
config to use Github Actions, which is also free for public
repositories.

While we're here, we also update our Node test matrix to include
Node 14, which is the current version.


ericyhwang (approved):
Good idea! It seems like Travis has had large continuous backlogs for open-source projects recently:

Screen Shot 2020-11-25 at 11 44 56 AM

PR #410, opened 2020-11-26
Bump mocha

alecgibson:
This change bumps `mocha` to the latest version. v8 [dropped support][1]
for `mocha.opts`, so we replace it with a `.mocharc.yml` config file
instead.

[1]: https://github.com/mochajs/mocha/blob/master/CHANGELOG.md#boom-breaking-changes


pypmannetjies (approved):
👌

PR #411, opened 2020-11-26
Fix MilestoneDB tests not exiting

alecgibson:
On newer test runners, the MilestoneDB tests may not exit properly when
backed by a real database.

This is because - in some of the nested tests - we `create` a new
database without closing the first database, which never gets destroyed
and therefore the tests never exit, since a socket to the database is
left open.

This change removes these nested `create` calls to avoid this hanging
database state.


pypmannetjies (approved):
Makes sense

PR #412, opened 2020-12-02
Test for calling `db.close()` multiple times

alecgibson:
We should be able to call `db.close()` even after the database has
already been closed, and not get an error.

This change adds a test for this case.

### DIFF
--- a/test/db.js
+++ b/test/db.js
@@ -36,6 +36,16 @@ module.exports = function(options) {
       });
     });
 
+    it('calls db.close() multiple times', function(done) {
+      var db = this.db;
+      db.close(function(error) {
+        if (error) return done(error);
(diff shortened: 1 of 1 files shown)


pypmannetjies:
Are you adding the test case in order to make a fix? Because it looks like it will throw an error right now.

alecgibson:
It's a potential fix for part of https://github.com/share/sharedb-mongo/issues/99

The "core" build passes; it just fails on testing ShareDB Mongo. We'd need to merge this first: https://github.com/share/sharedb-mongo/pull/100

I wasn't sure what the expected behaviour was, but since there's not a test case in the "opposite" direction, this felt more sensible to me.

ericyhwang (approved):

PR #416, opened 2020-12-16
Fix flaky test

alecgibson:
This change is a speculative fix for a flaky test. The `sharedb-mongo`
build has sometimes been failing because the following test sometimes
[fails][1]:

```
subscribe bulk with readSnapshots rejectSnapshotRead special ignorable error
```

The test fails in the `after each` hook, because we try to write a
snapshot to Mongo after Mongo has already been closed (ie `done()` has
already been called).

A potential cause for this is the fact that we don't wait for the
`submitOp` calls to complete before exiting the test. It's possible that
in a fraction of runs, the following happens:

 - Submit op on `fido`
 - Submit op on `spot`
 - `spot` op committed
 - `'receive'` called
 - nextTick, but `fido` op **still** not committed yet
 - Run assertions, and call `done()`
 - Mongo closed in the `afterEach` hook
 - `fido` finally tries to commit, but Mongo has been closed already

This change attempts to guard against this case by wrapping everything
in a `parallel` call, so we wait for the `submitOp` calls to complete
before exiting the test.

[1]: https://github.com/share/sharedb-mongo/runs/1565317742?check_suite_focus=true

### DIFF
--- a/test/client/subscribe.js
+++ b/test/client/subscribe.js
@@ -272,19 +272,27 @@ module.exports = function() {
                 done(new Error('fido on connection2 should not have received any ops, got:' +
                   JSON.stringify(op)));
               });
-              // Issue some operations on connection1.
-              connection.get('dogs', 'fido').submitOp([{p: ['age'], na: 1}]);
-              connection.get('dogs', 'spot').submitOp([{p: ['age'], na: 1}]);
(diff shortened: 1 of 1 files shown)


curran:
LGTM!

ericyhwang (approved):

PR #418, opened 2020-12-19
remove JSON.stringify

justinpinili:
- pass in the raw chunk data instead of stringifying it, to avoid stringify limitations


curran:
I'm curious, what are the "stringify limitations"?

distracteddev:
@curran String length in Node 10+ is 1 GB. This means that the current sharedb client cannot support query results larger than that.

PR #419, opened 2020-12-22
subscribe return meta

aeroyu:


alecgibson:
Thanks for the Pull Request @aeroyu ! This is a nice feature. I won't have time to look at this properly until the new year, but off the top of my head we'll need at least the following features to merge this:

## Metadata opt-in

Getting metadata should probably be opt-in. Some consumers may not want their clients to have access to potentially secret metadata, so we should:

 - Add a feature flag to the backend, so that consumers don't accidentally start leaking data by bumping their `sharedb` dependency
 - Force the client to also opt-in to metadata by setting a flag on `Doc`, so that the client can declare whether it cares about meta or not

##  Reset `meta`

There are various cases where we _unset_ `doc.data`. We should also unset the `meta` in these cases (search through `doc.js` for `data = ` to get a rough idea). I can't remember off the top of my head, but it'll be things like creating, destroying or resetting a doc.

aeroyu:
> Thanks for the Pull Request @aeroyu ! This is a nice feature. I won't have time to look at this properly until the new year, but off the top of my head we'll need at least the following features to merge this:
> 
> ## Metadata opt-in
> Getting metadata should probably be opt-in. Some consumers may not want their clients to have access to potentially secret metadata, so we should:
> 
> * Add a feature flag to the backend, so that consumers don't accidentally start leaking data by bumping their `sharedb` dependency
> * Force the client to also opt-in to metadata by setting a flag on `Doc`, so that the client can declare whether it cares about meta or not
> 
> ## Reset `meta`
> There are various cases where we _unset_ `doc.data`. We should also unset the `meta` in these cases (search through `doc.js` for `data = ` to get a rough idea). I can't remember off the top of my head, but it'll be things like creating, destroying or resetting a doc.

that will be great

PR #426, opened 2021-01-11
Optionally send `source` to the server

alecgibson:
This change adds the capability to send an op's `source` to the server,
for access in middleware.

# Motivation

The use-case that inspired this change was the ability to record some
usage stats based on users' ops. It's possible to record these stats
through separate HTTP calls, but recording stats through ShareDB would:

 - keep stats in sync with committed ops
 - allow servers to dictate stat definitions (enabling immediate
   roll-out; consistency; etc.)
 - leverage existing socket connections, which prevents having to open
   another socket; make lots of HTTP calls; or resort to
   [`sendBeacon`][1]

However, the type of stats we wish to record require more transient
information about the source of the op (eg keyboard; clipboard; mouse;
etc.).

It's clearly impractical to set this at connection time (where we set
other more "permanent" metadata, such as user ID), so we need some way
of signalling transient metadata attached to individual ops.

This can be communicated within the same client using the `source`
argument on `submitOp`, but this information never makes it to the
server.

# Implementation

## Usage

This new feature is opt-in, and is enabled on a per-`Doc` basis by
setting `doc.submitSource = true`.

No further configuration is required, and `source` can be set as it was
before in `submitOp`.

The `source` will be available in middleware at `request.extra.source`.

Of particular note is the fact that this feature supports complex
objects as `source`s, so it's possible to send complex metadata
structures to the server (although please note the side-effect mentioned
below).

## `op.x`

This change adds a new `x` property to the structure of a wrapped `op`.
This `x` represents "extra" information, which will not be committed to
the database, but will be made available in middleware (if consumers
want to commit it themselves, they are free to move data from `x` to `m`
in the middleware).

For now, the only property `x` can have is `source`, which will be
transmitted from the client.

## Side-effect

When using this feature, there is a necessary side-effect: any ops that
don't have a matching `source` will not be composed (since we don't know
how to disentangle these ops in the middleware).

This should be unimportant in the majority of cases (particularly if
`source` has not been actively set).

[1]: https://volument.com/blog/sendbeacon-is-broken

### DIFF
--- a/lib/client/connection.js
+++ b/lib/client/connection.js
@@ -454,11 +454,13 @@ Connection.prototype.sendOp = function(doc, op) {
     d: doc.id,
     v: doc.version,
     src: op.src,
-    seq: op.seq
+    seq: op.seq,
+    x: {}
   };
   if ('op' in op) message.op = op.op;
   if (op.create) message.create = op.create;
   if (op.del) message.del = op.del;
+  if (doc.submitSource) message.x.source = op.source;
   this.send(message);
 };
 
--- a/test/middleware.js
+++ b/test/middleware.js
@@ -2,6 +2,7 @@ var Backend = require('../lib/backend');
 var expect = require('chai').expect;
 var util = require('./util');
 var types = require('../lib/types');
+var errorHandler = util.errorHandler;
 
 describe('middleware', function() {
   beforeEach(function() {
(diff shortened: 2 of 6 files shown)


lib/client/connection.js
```diff
@@ -454,11 +454,13 @@ Connection.prototype.sendOp = function(doc, op) {
...
     v: doc.version,
     src: op.src,
-    seq: op.seq
+    seq: op.seq,
+    x: {}
   };
   if ('op' in op) message.op = op.op;
   if (op.create) message.create = op.create;
   if (op.del) message.del = op.del;
+  if (doc.submitSource) message.x.source = op.source;
```
pypmannetjies, line 463: I wonder if `submitSource` reads clear enough as a flag?

Though I must admit I struggle to think of a better one 🤔 

Maybe `recordSource` or `trackOpSource`

test/middleware.js
```diff
@@ -314,4 +315,91 @@ describe('middleware', function() {
...
+
+      this.backend.use('commit', function(request) {
+        expect('source' in request.extra).to.be.false;
+        done();
+      });
+
+      doc.submitOp([{p: ['tricks'], oi: ['fetch']}], {source: 'trainer'}, errorHandler(done));
+    });
+
+    it('composes ops with the same source', function(done) {
```
pypmannetjies, line 374: Might be good to add a test for 'composes ops with different sources if `submitSource` is false' though not sure how likely that scenario is

pypmannetjies (commented):
It LGTM but will defer to Eric for approval

lib/client/connection.js
alecgibson: Yeah I mean it kind of sounds like a method I guess. Could be `shouldSubmitSource`/`sourceIsSubmitted`? But starts getting verbose

test/middleware.js
alecgibson: Done.

lib/client/connection.js
pypmannetjies: `keepTrackOfSourceWhenSubmittingOps`? 😄

ericyhwang (approved):
LGTM on the code changes

Please update the documentation for `options.source` in the submit methods to indicate that if `submitSource` is true, the `source` will get sent to the server but not persisted or sent other clients

PR #429, opened 2021-01-25
Add hook for end of op lifecycle

alecgibson:
This change adds a new `after submit` event at the end of an op
submission, *regardless of success*.

At the moment, we have:

 - `submit`: called when an op is submitted
 - `afterWrite`: called after an op is successfully committed to the
   database
 - `reply`: called when the `Agent` is about to send a non-errored
   reply back to the client

However, it's currently impossible to react to ops that error, or to
ensure that you catch an event whenever an op is finally "done"
processing, regardless of success or failure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,6 +628,16 @@ Fetch snapshots that match the provided query. In most cases, querying the backi
 * `options` _Object_: an object that may contain a `db` property, which specifies which database to run the query against. These extra databases can be attached via the `extraDbs` option in the `Backend` constructor
 * `callback` _Function_: a callback with the signature `function (error: Error, snapshots: Snapshot[], extra: Object): void;`, where `snapshots` is an array of the snapshots matching the query, and `extra` is an (optional) object that the database adapter might return with more information about the results (such as counts)
 
+#### `on`
+
+```javascript
+backend.on('submitRequestEnd', callback): void;
+```
+
+A submit request is about to return to the client. This is called whether the request was a success or a failure.
+
+* `callback` _Function_: callback for handling the event: `function (error, request): void;`
+
 ### Class: `ShareDB.Agent`
 
 An `Agent` is the representation of a client's `Connection` state on the server. If the `Connection` was created through `backend.connect` (ie the client is running on the server), then the `Agent` associated with a `Connection` can be accessed through a direct reference: `connection.agent`.
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -204,11 +204,17 @@ Backend.prototype.trigger = function(action, agent, request, callback) {
 // Submit an operation on the named collection/docname. op should contain a
 // {op:}, {create:} or {del:} field. It should probably contain a v: field (if
 // it doesn't, it defaults to the current version).
-Backend.prototype.submit = function(agent, index, id, op, options, callback) {
+Backend.prototype.submit = function(agent, index, id, op, options, originalCallback) {
+  var backend = this;
+  var request = new SubmitRequest(this, agent, index, id, op, options);
+
+  var callback = function(error, ops) {
+    backend.emit('submitRequestEnd', error, request);
+    originalCallback(error, ops);
+  };
+
   var err = ot.checkOp(op);
   if (err) return callback(err);
-  var request = new SubmitRequest(this, agent, index, id, op, options);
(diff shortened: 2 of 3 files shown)


alecgibson:
This is a small and unassuming PR, but probably needs a lot of discussion:

# Event vs Middleware

I'd first thought this would make sense as a middleware hook, which would also make this consistent with all our other lifecycle stuff. *However*, this lifecycle hook is different, because no other middleware ever runs after the lifecycle has errored. The entire point of this hook is that it would run in both success and failure cases. If we do that, then we have to either:

 - support multiple errors being returned from middleware (`readSnapshots` potentially already does something like this...?); or
 - disallow errors being returned in the `readSnapshots` callback (which would also make this feel and act differently to other middleware anyway)

Rather than go with some confusing middleware hook that acts differently to all the others, I've currently implemented this as an event, showing that this is a purely informational hook, and that consumers aren't intended to interact with the request.

That being said, I'm still on the fence myself. It feels very odd adding an event, where all the other events on the `Backend` are currently to do with timing. Perhaps if we added a corresponding `submit` event at the start of the lifecycle, it would feel less weird, but then that hook would be made completely redundant by the existence of the `submit` middleware.

We could potentially also look into other alternatives, like turning the `SubmitRequest` class into an emitter, and allowing middlewares to subscribe to the request object itself. This might be a more "sensible"-looking API: `request.on('end', () => {})`, but isn't particularly in keeping with the rest of the lifecycle hooks.

# Naming

`after submit` is a [bad name](https://github.com/share/sharedb/pull/314), and we should definitely rename it. I've used it as a placeholder name for now. Possible suggestions:

 - `submit done`
 - `submit request done`
 - `submit end`
 - `submit request destroy`
 - ???

The other thing to decide is wither we use spaces or camelCase:

 - `Doc` events use spaces (like `before op`)
 - `Backend` middleware hooks are camelCased (like `afterWrite`), and were actively moved away from spaces

lib/backend.js
```diff
@@ -204,11 +204,17 @@ Backend.prototype.trigger = function(action, agent, request, callback) {
 // Submit an operation on the named collection/docname. op should contain a
 // {op:}, {create:} or {del:} field. It should probably contain a v: field (if
 // it doesn't, it defaults to the current version).
-Backend.prototype.submit = function(agent, index, id, op, options, callback) {
+Backend.prototype.submit = function(agent, index, id, op, options, originalCallback) {
+  var backend = this;
+  var request = new SubmitRequest(this, agent, index, id, op, options);
+
+  var callback = function(error, ops) {
+    backend.emit('after submit', error, request);
```
pypmannetjies, line 212: Like we discussed, I think using an "-ing" suffix would help convey that this occurs while the submit is busy finishing up and not actually afterwards. So for example: 

```suggestion
    backend.emit('submit finalizing', error, request);
```

pypmannetjies (commented):
After discussing this in the weekly meeting, it LGTM. Makes sense that it's not really Middleware. The other option is to change SubmitRequest into an event emitter which neither of us preferred as it will complicate things for the caller. 

Can we also add this new event to the documentation?

lib/backend.js
alecgibson: I think I'm leaning towards `submit end` personally?

pypmannetjies: TIL about "gerund"

README.md
```diff
@@ -628,6 +628,16 @@ Fetch snapshots that match the provided query. In most cases, querying the backi
...
 
+#### `on`
+
+```javascript
+backend.on('submitRequestEnd', callback): void;
+```
+
+A submit request is about to return to the client. This is called whether the request was a success or a failure.
+
+* `callback` _Function_: callback for handling the received presence: `function (error, request): void;`
```
ericyhwang, line 639:
```suggestion
* `callback` _Function_: callback for handling the error: `function (error, request): void;`
```

ericyhwang (approved):

alecgibson:
As discussed, we'll go with an event, because middleware behaviour would be poorly defined. We'll go with `submitRequestEnd` as the event name to avoid confusion with whether this happens after the `submit` middleware or the whole request. We go with camelCasing because Node does this with eg `unhandledRejection`.

Issue #255, opened 2018-11-05
Support browser environment without Browserify


curran:
Currently some client code accesses `process.nextTick`, in [doc.js line 663](https://github.com/share/sharedb/blob/5f4bd902d5679a1fa87f7e954f1c407d18a9049d/lib/client/doc.js#L663).

This is shimmed to `setTimeout` when building with Browserify, but other build tools such as Rollup do not automatically translate `process.nextTick` to `setTimeout`.

My current workaround is to shim it myself as follows:

```js
window.process = {

  // Required for ShareDB client. 
  nextTick: callback => {
    setTimeout(callback, 0);
  }
};
```

Would a PR that changes `process.nextTick` to `setTimeout` in the original source be welcome?

I don't see any disadvantages to making this change.

gkubisa:
`process.nextTick` in nodejs schedules a micro-task. `setTimeout` schedules a task. See https://jakearchibald.com/2015/tasks-microtasks-queues-and-schedules/ for a difference between tasks and micro-tasks.

In my opinion a micro-task would be more appropriate wherever ShareDB uses `process.nextTick`, as we just want to perform an action asynchronously as soon as possible and `setTimeout` introduces unnecessary delay. So, I'd prefer if the shim was based on `MutationObserver` or `Promise`, rather than `setTimeout`.

ericyhwang:
My comments and thoughts:

1. Should we change `process.nextTick` to `setTimeout` in the source code?
    - No, as they do different things in NodeJS.
2. Should ShareDB itself shim `process.nextTick`?
    - I haven't really thought too deeply about if Share should do it, or to leave it to the browser bundling tool.
    - One option is to recommend a shim if your bundler doesn't do it for you. Looks like browserify uses the [`process`](https://www.npmjs.com/package/process) module as a shim.

curran:
I researched a bit more and found the following to be the implementation of `nextTick` in the Browserify bundle:

``js
process.nextTick = function (fun) {
    var args = new Array(arguments.length - 1);
    if (arguments.length > 1) {
        for (var i = 1; i < arguments.length; i++) {
            args[i - 1] = arguments[i];
        }
    }
    queue.push(new Item(fun, args));
    if (queue.length === 1 && !draining) {
        runTimeout(drainQueue);
    }
};
```

Indeed this comes from the [process module implementation](https://github.com/defunctzombie/node-process/blob/master/browser.js#L134). Good to know!

[...]

Issue #427, opened 2021-01-23
Client crashing on submitOp - process is not defined


miking-the-viking:
From my research process is not available in the browser. Has anyone else run into this problem?

![image](https://user-images.githubusercontent.com/31659809/105612840-9154f200-5d8c-11eb-8cf5-2594d1d02ecd.png)

alecgibson:
Probable duplicate of https://github.com/share/sharedb/issues/255

miking-the-viking:
Yeah that looks bang on @alecgibson

miking-the-viking:
For reference I found a viable Webpack solution to setup the process polyfill, it was removed in Webpack 5 by default.

```
// webpack needs to be explicitly required
const webpack = require('webpack')

module.exports = {

/* ... rest of the config here ... */

  plugins: [
    // fix "process is not defined" error:
    // (do "npm install process" before running the build)
    new webpack.ProvidePlugin({
      process: 'process/browser',
    }),
  ]
}
```

from https://stackoverflow.com/a/64553486/5129500

alecgibson:
Just looking into a Webpack 5 bump myself. I do think we should address this. It's pretty weird that we force consumers to polyfill all of `process` (not to mention also `stream`).

As a first step, we could potentially just include these polyfills in ShareDB itself, especially since our advice is to use Browserify, or to use Webpack (which now recommends using Browserify modules anyway).

PR #434, opened 2021-02-10
Ponyfill `nextTick`

alecgibson:
Fixes https://github.com/share/sharedb/issues/255 & https://github.com/share/sharedb/issues/427

At the moment, ShareDB uses the Node.js `process.nextTick` everywhere,
including in client-side code.

ShareDB's current advice is to use a modern build-chain tool, such as
Webpack or Browserify to add support for these Node.js functions.
However, Webpack v5 has [removed][1] their Node.js polyfills. Webpack
actively recommends package authors to "provide alternative
implementations/dependencies for the browser."

It's now quite weird that ShareDB forces consumers to polyfill the
entire `process` module, when we really just need `nextTick`. In order
to keep ShareDB usage as simple as possible, this change ponyfills
`process.nextTick`:

 - If `process.nextTick` is already defined, we just use this directly
 - Otherwise use `setTimeout` (which is essentially the underlying
   behaviour in the recommended [`process`][2] module)

[1]: https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed
[2]: https://github.com/defunctzombie/node-process/blob/master/browser.js#L134-L145

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -28,6 +28,8 @@ const SHAREDB_RULES = {
   'no-unused-vars': ['error', {vars: 'all', args: 'after-used'}],
   // It's more readable to ensure we only have one statement per line
   'max-statements-per-line': ['error', {max: 1}],
+  // ES3 doesn't support spread
+  'prefer-spread': 'off',
   // as-needed quote props are easier to write
   'quote-props': ['error', 'as-needed'],
   'require-jsdoc': 'off',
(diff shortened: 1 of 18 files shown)


pypmannetjies (commented):
It LGTM. I would suggest a simple test case to confirm that if `process.nextTick` is present it is not replaced (still equals the original) and if it is not present that `util.nextTick` is defined and is a function. But not sure how easy it will be to stub.

alecgibson:
@pypmannetjies I've added a nasty unit test that deletes `process.nextTick` 🙈  Seems to do the trick, though.

pypmannetjies (approved):
LGTM

PR #435, opened 2021-02-17
Remove dependence on `util`

alecgibson:
At the moment, ShareDB uses the Node.js `util` package just to use
`inherits()`.

Although this is only used server-side, it would be nice to keep the
code client-side compatible (eg for running unit tests in Karma, or
similar browser-based test environments).

Webpack v5 recently [removed][1] their Node.js polyfills, and it would
be nice if we didn't rely on Node.js modules where possible.

Since `util.inherits()` is just a [very thin wrapper][2] around
`Object.setPrototypeOf()`, this change simply swaps this usage for the
native method, which has support even back to Internet Explorer 11.

[1]: https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed
[2]: https://github.com/nodejs/node/blob/8353854ed7c658ae3692fb65f12adadf6f38307c/lib/util.js#L163-L181

### DIFF
--- a/lib/op-stream.js
+++ b/lib/op-stream.js
@@ -1,4 +1,3 @@
-var inherits = require('util').inherits;
 var Readable = require('stream').Readable;
 var util = require('./util');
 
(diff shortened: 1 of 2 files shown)


alecgibson:
Actually just realised we'd need to shim the `stream` module as well — maybe this isn't worth it?

alecgibson:
As discussed, `Stream` and `Duplex` would also need polyfilling, and it's reasonable to expect a consumer to polyfill Node.js modules if they're trying to run server-side code in the browser.

PR #438, opened 2021-03-03
📝 Add Jekyll documentation

alecgibson:
This change attempts to move us away from our `README.md`, which we've
outgrown, into a series of Jekyll-compatible Markdown files, which
focus on better segmentation of the documentation, and having more
headings that can be linked to.

The aim of this documentation is to be hosted on Github Pages.


docs/api/agent.md
```diff
@@ -0,0 +1,28 @@
...
+
+1. TOC
+{:toc}
+
+An `Agent` is the representation of a client's [`Connection`]({% link api/connection.md %}) state on the server.
+
+The `Agent` instance will be made available in all [middleware]({% link middleware/index.md %}) contexts, where [`agent.custom`](#custom--object) can be particularly useful for storing custom context.
+
+{: .info }
+If the `Connection` was created through [`backend.connect()`]({% link api/backend.md %}#connect()) (ie the client is running on the server), then the `Agent` associated with a `Connection` can be accessed through [`connection.agent`]({% link api/connection.md %}#agent--agent).
```
pypmannetjies, line 18: nit:

```suggestion
If the `Connection` was created through [`backend.connect()`]({% link api/backend.md %}#connect()) (i.e. the client is running on the server), then the `Agent` associated with a `Connection` can be accessed through [`connection.agent`]({% link api/connection.md %}#agent--agent).
```

docs/api/backend.md
```diff
@@ -0,0 +1,596 @@
...
+
+Optional
+{: .label .label-grey }
+
+> Default: `null`
+
+> An instance of a ShareDB [milestone adapter]({% link adapters/milestone.md %}) that provides the data store for [milestone snapshots]({% link document-history.md %}#milestone-snapshots), which are historical snapshots of documents stored at a specified version interval.
+
+{: .info }
+> If this option is omitted, milestone snapshots will **not** be enabled, but historic snapshots *may* still be fetched.
```
pypmannetjies, line 68: nit: Can we explain a bit more? I don't really understand how the two are different.

docs/api/backend.md
```diff
@@ -0,0 +1,596 @@
...
+
+> The document ID
+
+`from` -- number
+
+> The first op version to fetch. If set to `null`, then ops will be fetched from the earliest version
+
+`to` -- number
+
+> The last op version. This version will *not* be fetched (ie `to` is non-inclusive). If set to `null`, then ops will be fetched up to the latest version
```
pypmannetjies, line 346: nit: 
```suggestion
> The last op version. This version will *not* be fetched (i.e. `to` is non-inclusive). If set to `null`, then ops will be fetched up to the latest version
```

docs/api/doc.md
```diff
@@ -0,0 +1,470 @@
...
+> The document contents
+
+{: .warn }
+> The data will only be available after a document has been fetched or subscribed to.
+
+### `preventCompose` -- boolean
+
+> Default: `false`
+
+> Set to `true` to prevent ops from being composed together. This is read at the time of calling [`submitOp()`](#submitop), so it may be toggled on before submitting a specific op, and toggled off again afterwards
```
pypmannetjies, line 56: Have we explained elsewhere what it means if ops are composed together?

docs/api/doc.md
```diff
@@ -0,0 +1,470 @@
...
+### del()
+
+Delete the document locally and send a delete operation to the server.
+
+```js
+doc.del([options [, callback]])
+```
+
+{: .info }
+ShareDB documents and their ops are never truly deleted from the database. Instead, they will be tombstoned.
```
pypmannetjies, line 280: Maybe we can explain what a tombstone looks like for us 🤔 
Here is a suggestion but I'm not sure I'm capturing all of it. 

```suggestion
ShareDB documents and their ops are never truly deleted from the database. Instead, they will be tombstoned. A tombstone is a document that has no data, but retains some metadata such as the original `_id`. ShareDB will ignore these tombstones when queried, but they are still present in the underlying database. 
```

docs/api/sharedb-error.md
```diff
@@ -0,0 +1,98 @@
...
+
+> This is non-critical, and part of normal control flow, and is sent as an error in order to short-circuit the op processing. It is eventually swallowed by the server, and shouldn't need further handling.
+
+### `ERR_SUBMIT_TRANSFORM_OPS_NOT_FOUND`
+
+> The ops needed to transform the submitted op up to the current version of the snapshot could not be found.
+
+> If a client on an old version of a document submits an op, that op needs to be transformed by all the ops that have been applied to the document in the meantime. If the server cannot fetch these ops from the database, then this error is returned.
+
+> The most common case of this would be ops being deleted from the database. For example, let's assume we have a TTL set up on the ops in our database. Let's also say we have a client that is so old that the op corresponding to its version has been deleted by the TTL policy. If this client then attempts to submit an op, the server will not be able to find the ops required to transform the op to apply to the current version of the snapshot.
```
pypmannetjies, line 50:
```suggestion
> The most common case of this would be ops being deleted from the database. For example, let's assume we have a TTL set up on the ops in our database. Let's also say we have a client that is so out of date that the op corresponding to its version has been deleted by the TTL policy. If this client then attempts to submit an op, the server will not be able to find the ops required to transform the op to apply to the current version of the snapshot.
```

docs/api/sharedb-error.md
```diff
@@ -0,0 +1,98 @@
...
+
+> The specified document [type]({% link types/index.md %}) has not been registered with ShareDB.
+
+> This error can usually be remedied by remembering to [register]({% link types/index.md %}#installing-other-types) any types you need.
+
+### `ERR_DEFAULT_TYPE_MISMATCH`
+
+> The default type being used by the client does not match the default type expected by the server.
+
+> This will typically only happen when using a different default type to the built-in `json0` used by ShareDB by default (eg if using a fork). The exact same type must be used by both the client and the server, and should be registered as the default type:
```
pypmannetjies, line 80:
```suggestion
> This will typically only happen when using a different default type to the built-in `json0` used by ShareDB by default (e.g. if using a fork). The exact same type must be used by both the client and the server, and should be registered as the default type:
```

docs/faq.md
```diff
@@ -1,9 +0,0 @@
-# FAQ
-
-## Is it possible to completely delete documents from the db?
-
-No, it is not possible to use the ShareDB API to fully delete data. In addition, the operation log is kept forever by default.
-
-Maintaining persistence of snapshots and ops means that ShareDB can correctly deal with all cases where ops have been removed. Permanently removing the snapshot document could result in a corrupt state in some edge cases by not maintaining the current document version, which must be incremented on each commit. As well, if you delete ops and then a client reconnects needing those ops, you will break that client and it will be unable to submit any pending changes or bring itself up to date from its current state. If your use case calls for complete deletion of operations, you'll need to ensure that no clients will ever need them again or deal with the error appropriately.
```
pypmannetjies, line 7: Maybe we should keep this "justification" somewhere else like in the documentation for the `del` operation

docs/projections.md
```diff
@@ -0,0 +1,33 @@
...
+# Projections
+
+Some [types]({% link types/index.md %}) support exposing a projection of a real collection, with a specified set of allowed fields.
+
+{: .info }
+Currently, only [`json0`]({% link types/json0.md %}) supports projections.
+
+Once configured, the projected collection looks just like a real collection -- except documents only have the fields that have been specified.
+
+Operations on the projected collection work, but only a small portion of the data can be seen.
```
pypmannetjies, line 16:
```suggestion
Operations on the projected collection work, but only a small portion of the data can be seen and altered.
```

pypmannetjies (commented):
Just a couple really minor suggestions from me. As discussed in the weekly meeting, we can move ahead with this and make changes as required in future.

docs/api/doc.md
alecgibson: No, but the original `README` doesn't mention it either. I'll add a TODO next to this to add some documentation on composing.

docs/faq.md
alecgibson: See comment above.

PR #440, opened 2021-03-14
fix for process being undefined in browser environment

ChrisWojcik:
Still seeing an issue with Webpack v5 even after #434.

`process` is undefined in a browser environment, so `if (process) ...` will throw a reference error. A check with `typeof` needs to be used instead.

Couldn't come up with a way to unit test this differently.


alecgibson (approved):
Of course. Good catch.

Issue #442, opened 2021-03-15
Do not `invert` automatically in `Doc.prototype._rollback`


antoinelyset:
Hey ShareDB's team!

While using https://github.com/ottypes/json1 with ShareDB, I found a small problem in `Doc.prototype._rollback`. It automatically uses `JSON1.type.invert` but since we don't use invertible operations (our remove ops do not contain the removed data), it does not work.
https://github.com/share/sharedb/blob/fa9179dd565b11e129c1f005de0a7542ef352363/lib/client/doc.js#L950-L957

Moreover JSON1 support for invertible operations and composition is known to not be [without flaws](https://github.com/ottypes/json1/issues/20).

Do you think we could introduce a configuration option to deactivate this functionality?

alecgibson:
Quick hacky fix is to just:

```js
delete types.json1.invert;
```

But obviously that's pretty ugly, and you may or may not want to use `invert` elsewhere.

In general, I'd expect if a type implements `invert` for the ops to be invertible. That being said, Pull Requests are always welcome!

curran:
Is this a duplicate of https://github.com/share/sharedb/issues/451?

hsource:
I did a more targeted and backwards-compatible fix in #520 - Fall back to doing a _hardRollback when op.type.invert fails. Let me know if you have any feedback on that! We're running that in a patched production build already

antoinelyset:
Hi! I Just saw this change.
For the record, it won't fix the issue with `ot-json1` since `invert` exists and will not fail, it will just invert incorrectly see https://github.com/share/sharedb/pull/443#issuecomment-802432950.

alecgibson:
Noted. I think I'm in favour of deprecating "soft" rollback as default behaviour. I'm not sure `invert()` is particularly bullet-proof in any of the widely-used OT types.

happy15:
One more possible case: 

Say we have op1, op2, op3, op4, and all of op2 ~ op4 depend on the existence of op1 to be meaningful. Op1 is rejected by server and _rollback successful, which leaves op2 ~ op4 transformed against invert(op1). This results in transformed op2', op3', op4'. 

My question is: will op2', op3', op4' still be meaningful? Ideally, op2 ~ op4 should be removed along with op1. In the current framework of sharedb, the best bet may be to make the transform result of op2~op4 against invert(op1) to be no-op.

PR #443, opened 2021-03-16
Add preventInvertInRollback option

antoinelyset:
My try at adding an option to improve support for JSON1. https://github.com/share/sharedb/issues/442

I heavily relied on a previous spec, so it may be possible to simplify it.

### DIFF
--- a/test/client/doc.js
+++ b/test/client/doc.js
@@ -474,5 +474,74 @@ describe('Doc', function() {
         }
       ], done);
     });
+
+    describe('with an op collision', function() {
+      function testScenario(preventInvertInRollback, done) {
+        var doc1 = this.backend.connect().get('dogs', 'snoopy');
+        var doc2 = this.backend.connect().get('dogs', 'snoopy');
+        doc2.preventInvertInRollback = preventInvertInRollback;
+
+        var pauseSubmit = false;
+        var fireSubmit;
+        var invertHaveBeenCalled = false;
+
+        this.backend.use('submit', function(request, callback) {
+          if (pauseSubmit) {
+            fireSubmit = function() {
+              pauseSubmit = false;
+              callback();
+            };
+          } else {
+            fireSubmit = null;
+            callback();
+          }
+        });
+        async.series([
+          doc1.create.bind(doc1, {colours: ['white']}),
+          doc1.whenNothingPending.bind(doc1),
+          doc2.fetch.bind(doc2),
+          doc2.whenNothingPending.bind(doc2),
+          function(next) {
+            // Wrap invert to test that it has been called
+            var originalInvert = doc2.type.invert;
+            doc2.type.invert = function(op) {
+              invertHaveBeenCalled = true;
+              return originalInvert(op);
+            };
+            next();
+          },
+          doc1.submitOp.bind(doc1, {p: ['colours'], oi: 'white,black'}),
+          function(next) {
+            pauseSubmit = true;
+            doc2.submitOp({p: ['colours', '0'], li: 'black'}, function(error) {
+              expect(error.message).to.equal('Referenced element not a list');
+              next();
+            });
+
+            doc2.fetch(function(error) {
+              if (error) return next(error);
+              fireSubmit();
+            });
+          },
+          function(next) {
+            expect(invertHaveBeenCalled).to.eql(!preventInvertInRollback);
+            expect(doc2.data).to.eql(doc1.data);
+            next();
+          }
+        ], done);
+      }
+
+      it('calls doc type invert by default', function(done) {
+        var preventInvertInRollback = false;
+        var testScenarioBound = testScenario.bind(this);
+        testScenarioBound(preventInvertInRollback, done);
+      });
+
+      it('does not call doc type invert if preventInvertInRollback is `true`', function(done) {
+        var preventInvertInRollback = true;
+        var testScenarioBound = testScenario.bind(this);
+        testScenarioBound(preventInvertInRollback, done);
+      });
+    });
   });
 });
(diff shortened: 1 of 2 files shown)


antoinelyset:
Thanks for your quick answer. I will ask what we can do in the JSON1 repo.

josephg:
Hi! json1 and sharedb original author here. I think can do better than those options.

The problem isn't that json1 is broken. The problem is that json1 has both invertible and non-invertible operations. Non-invertible operations come from two sources:

1. Creating remove / replace operations which don't specify the removed / replaced content. Eg `json1.removeOp(['a'])` produces a non-invertible operation, while `json1.removeOp(['a'], "deleted content")` produces an invertible operation. (When inverted, it will insert "deleted content" into doc.a)
2. Calling transform will remove invertibility information in many cases. This is unfortunately unavoidable. Consider op1 moves doc.a to doc.b.c, and op2 deletes doc.b. transform(op1, op2) will produce an operation that deletes doc.a, but the transform function does not know the original contents of doc.a, and thus can't make the result invertible. There used to be a bug where compose would produce non-invertible operations as well - I think thats been fixed but its a long time since I looked at the code. Issues / patches welcome!

To make matter worse, there's no flag in an operation which specifies whether or not its invertible. This is a flaw in the design.

json1 has two methods to help here:

- `json1.makeInvertible(op, doc) -> op` adds the missing information back into an operation
- `invertWithDoc(op, doc) -> doc` - which inverts an operation given the document state before the operation was originally applied. (Now I think about it isn't super useful because if you knew that, you probably don't need to call invert.) `invertWithDoc` is just a wrapper is just a wrapper around `invert(makeInvertible(op, doc))`.

I had sharedb in mind when I wrote json1 (and if it existed when sharedb was written, we would have used it). Its been a long time since I looked at the code, but sharejs should be able to use the makeInvertible method in json1 to restore invertibility information in the undo / rollback stack, since the state is known by sharedb when the operation is created.

```javascript
> doc = {a:'hi'}
> op = json1.removeOp(['a'])
[ 'a', { r: true } ]
> json1.makeInvertible(op, doc)
[ 'a', { r: 'hi' } ]
```

alecgibson:
All in all, it sounds like inversion is a bit "unreliable" from the perspective of ShareDB consuming an API — that is ShareDB has no idea of whether or not it should be able to apply an inversion to any given op. This is also [true of `json0`](https://github.com/ottypes/json0/pull/23).

@josephg we could potentially track an in-memory undo stack (this would incidentally make `rich-text` psuedo-invertible, too, since it also implements a method that can invert an op with the doc). However, I feel like that would be a lot of effort to go to for error recovery (when we could just hard rollback instead, which — in my opinion — would be more robust anyway).

In an ideal world, my suggestion would be to default to hard rollback, and make invertible fallback opt-in. However, that would be a breaking change, so we probably can't go down that route.

Perhaps this PR is then the most pragmatic solution. @ericyhwang had raised the concern that it's a bit awkward to use as a consumer: you have to set whether you want an inverted rollback on each document. However, on reflection, I think it's possible that some consumers might use different documents/collections in different ways (perhaps they use invertible ops on one, but not on another?). It's a bit awkward, but applying it at this level gives consumers the greatest granularity of control.

My only other thought is that we could perhaps offer an event before rollback (which might also be nice for logging). This would also let consumers decide per-op if they want to soft rollback or not. A potential API might look like:

```js
doc.on('before rollback', (event) => {
  if (consumerNotWantInvert(event.op)) {
    event.hardRollback();
  }
});
```

josephg:
Sure - happy to defer to your opinion on what sharedb should do. The rollback infrastructure is from after my time, and I wasn't in those design meetings.

Otherwise if there's any changes you'd like to see in json1, let me know / file an issue.

test/client/doc.js
```diff
@@ -474,5 +474,74 @@ describe('Doc', function() {
...
+
+    describe('with an op collision', function() {
+      function testScenario(preventInvertInRollback, done) {
+        var doc1 = this.backend.connect().get('dogs', 'snoopy');
+        var doc2 = this.backend.connect().get('dogs', 'snoopy');
+        doc2.preventInvertInRollback = preventInvertInRollback;
+
+        var pauseSubmit = false;
+        var fireSubmit;
+        var invertHaveBeenCalled = false;
```
curran, line 486: I can't help but notice a grammar issue. Suggest to simplify this name.
```suggestion
        var invertCalled = false;
```

test/client/doc.js
```diff
@@ -474,5 +474,74 @@ describe('Doc', function() {
...
+        async.series([
+          doc1.create.bind(doc1, {colours: ['white']}),
+          doc1.whenNothingPending.bind(doc1),
+          doc2.fetch.bind(doc2),
+          doc2.whenNothingPending.bind(doc2),
+          function(next) {
+            // Wrap invert to test that it has been called
+            var originalInvert = doc2.type.invert;
+            doc2.type.invert = function(op) {
+              invertHaveBeenCalled = true;
```
curran, line 508:
```suggestion
              invertCalled = true;
```

test/client/doc.js
```diff
@@ -474,5 +474,74 @@ describe('Doc', function() {
...
+              next();
+            });
+
+            doc2.fetch(function(error) {
+              if (error) return next(error);
+              fireSubmit();
+            });
+          },
+          function(next) {
+            expect(invertHaveBeenCalled).to.eql(!preventInvertInRollback);
```
curran, line 527:
```suggestion
            expect(invertCalled).to.eql(!preventInvertInRollback);
```

curran (changes requested):
Minor name change suggestions.

Does this change effectively implement **Default to hard rollback** #451?

Is there anything missing here for addressing #451 or is this PR good to go?

PR #446, opened 2021-03-19
Add the custom values on the agent onto the snapshot options when getSnapshot is called

pypmannetjies:
It turns out that the `sharedb-mongo` [beforeSnapshotLookup](https://github.com/share/sharedb-mongo/blob/534143c110777304669b52f3920a5aec98e0c5d2/src/middleware/actions.js#L8) middleware is pretty useless, because there is no `sharedb` middleware that is triggered right before `getSnapshot` is called. We only have `readSnapshots` but it is triggered after `getSnapshot` is called. 

If we want to add some shard key onto the snapshot options, we would need to do it in such a middleware, similar to the approach taken with `beforeOverwrite`

However, @ericyhwang and I thought it might be generally more useful to always have some information passed to the `sharedb-<db>` adapter's middleware so that it is easier in general to implement. Otherwise this PR will turn into adding new middleware for sharedb just in order to pass some information from the agent to the db middleware. 

I know there were some concerns about passing the agent / request to the db implementation, which is why I thought a good approach is to just pass the `request.agent.custom`

For now this is just for `getSnapshot` and `getSnapshotBulk` - will have to do `commit` in a separate PR as this change is becoming a bit large.


alecgibson (commented):
I think this generally feels okay to me. For the record, my concerns with passing in the _whole_ `Agent` object were:

 - weird inversion of control: `Agent` knows about `Backend`, which should call the adapter, and not the other way around
 - opens up a wide API surface area, when adapters probably don't really need to know about most of the things inside `Agent`
 - `Agent` is a very "`sharedb`" concept, and adapters shouldn't really have any knowledge of `sharedb` internals at all

I have a couple of reservations about using `custom` like this:

 - In the majority of cases, `custom` will most probably contain information that has nothing to do with the DB adapter
 - I've always vaguely thought of `custom` as an immutable object that gets set on connection, and mutating the object as it moves through other middlewares feels like you're making your `custom` weirdly stateful (eg I assume it's easy to fall into traps where you set state for an op submit cycle, and then don't reset it before the next one begins?)

That being said, I probably have those hangups just because I use `custom` in a specific way, and really `custom` was intended as a generic place to dump consumer data, which this is, so it's probably fine.

pypmannetjies:
> I think this generally feels okay to me. For the record, my concerns with passing in the _whole_ `Agent` object were:
> 
> * weird inversion of control: `Agent` knows about `Backend`, which should call the adapter, and not the other way around
> * opens up a wide API surface area, when adapters probably don't really need to know about most of the things inside `Agent`
> * `Agent` is a very "`sharedb`" concept, and adapters shouldn't really have any knowledge of `sharedb` internals at all
> 
> I have a couple of reservations about using `custom` like this:
> 
> * In the majority of cases, `custom` will most probably contain information that has nothing to do with the DB adapter
> * I've always vaguely thought of `custom` as an immutable object that gets set on connection, and mutating the object as it moves through other middlewares feels like you're making your `custom` weirdly stateful (eg I assume it's easy to fall into traps where you set state for an op submit cycle, and then don't reset it before the next one begins?)
> 
> That being said, I probably have those hangups just because I use `custom` in a specific way, and really `custom` was intended as a generic place to dump consumer data, which this is, so it's probably fine.

Hi Alec

We discussed this and were wondering if you think it would be better to add a new property on `Agent` - something like `dbOptions` or even `dbMiddlewareOptions` so that it is clear and separate from Custom? It makes our use case a bit harder but it could help address some of these concerns as well as protect from "custom drift" losing the original intent of the Custom property.

alecgibson:
> would [it] be better to add a new property on `Agent`

It might be nicer. I was unsure myself, which is why I didn't suggest it. I don't know if it actually just adds a bit more confusion and needing to know exactly about how ShareDB internals work down to the adapter level (although I suppose you'd _need_ to know by the time you're mucking about in adapter middleware).

Or we can just leave it to consumers to structure their own `custom` data sensibly.

Would we actively empty this object on submit?

Another possible approach: can we abuse the `snapshotOptions` object to just include some payload?

pypmannetjies:
> Would we actively empty this object on submit?

No I don't think it makes sense to edit the object within sharedb, that seems like it could be surprising to consumers. What would the benefits be of emptying it? 

> Another possible approach: can we abuse the `snapshotOptions` object to just include some payload?

If you mean the existing `options.snapshotOptions` I don't think so because it doesn't exist currently on all of the call sites for `getSnapshot` and `getSnapshotBulk`. Also - I'm not sure where the value is populated from 🤔

alecgibson:
> No I don't think it makes sense to edit the object within sharedb, that seems like it could be surprising to consumers. What would the benefits be of emptying it?

I think I'm confused about how this object is going to be used. Could you please provide a bit of example consumer code, showing how you'd expect to use this?

pypmannetjies:
> > No I don't think it makes sense to edit the object within sharedb, that seems like it could be surprising to consumers. What would the benefits be of emptying it?
> 
> I think I'm confused about how this object is going to be used. Could you please provide a bit of example consumer code, showing how you'd expect to use this?

Alright so say for example you wanted to shard a collection, and whenever `getSnapshot` or `getSnapshotBulk` is called, you want to include the shard key in queries. 

According to the [MongoDB docs ](https://docs.mongodb.com/manual/core/distributed-queries/#read-operations-to-sharded-clusters):

>Read operations on sharded clusters are most efficient when directed to a specific shard. Queries to sharded collections should include the collection’s shard key. When a query includes a shard key, the mongos can use cluster metadata from the config database to route the queries to shards.

Our code only [looks up based on ID](https://github.com/share/sharedb-mongo/blob/master/index.js#L312).

We know some values of shard keys would always be the same per session (for example the ID of the account that the user belongs to)

So with this change, on connect, we would set something like:

```
connection.agent.dbMiddlewareOptions.shardKeyValue = 'foo'
```

Then, when a snapshot lookup is triggered, it is passed through the middleware. The middleware can then do something like:

```
db.use ('beforeSnapshotLookup', (request, next) => {
  request.query = {shardKey: request.options.shardKeyValue};
  next();
});
```

Now the lookups will always have the shard key in their queries.

alecgibson:
Aha I see so you do just want connection-time information. For some reason I had it in my head that you wanted to manipulate `custom` in a `sharedb` middleware before control is handed off to `sharedb-mongo` (eg add bonus shard context when an op is submitted), but I think I was conflating your two solutions.

So for clarity, you have two different suggestions:

 1. Add a new `sharedb` middleware like `beforeSnapshotRead` or something, where you can manipulate `snapshotOptions` before control is passed off into `sharedb-mongo` middleware
 2. No new middleware hook, and just add `agent.custom` into `snapshotOptions`

Maybe it is fine to just pass `custom` all the way through, then. I'm not sure it's worth the effort of maintaining a whole separate object just for adapter middleware, but equally separation of concerns is always nice. I don't feel too strongly either way. I think my main issue was when I thought you were mutating `custom` to achieve something, but I misunderstood.

pypmannetjies:
> Maybe it is fine to just pass `custom` all the way through, then. I'm not sure it's worth the effort of maintaining a whole separate object just for adapter middleware, but equally separation of concerns is always nice. I don't feel too strongly either way.

Yeah, to me it seems the most straightforward, as it might be overcomplicating things a little bit to add another property that serves essentially the same purpose as `custom`.  

>  I think my main issue was when I thought you were mutating `custom` to achieve something

Yep in this case we do not need to do anything to the property we just want access to it in `sharedb-mongo` somehow :) 

I _could_ add another middleware but we thought that if we are running into this issue, future consumers might as well, and it seems over engineered to require two middlewares just to read a property on `custom`.

lib/backend.js
```diff
@@ -391,7 +391,8 @@ Backend.prototype.fetch = function(agent, index, id, options, callback) {
     collection: collection,
     id: id
   };
-  var snapshotOptions = options && options.snapshotOptions;
+  var snapshotOptions = (options && options.snapshotOptions) || {};
+  if (agent) snapshotOptions.custom = agent.custom;
```
ericyhwang, line 395: How about `snapshotOptions.agentCustom` to make it clear where it's coming from? `options.custom` is very general really give a hint about where it's coming from.

lib/backend.js
```diff
@@ -391,7 +391,8 @@ Backend.prototype.fetch = function(agent, index, id, options, callback) {
     collection: collection,
     id: id
   };
-  var snapshotOptions = options && options.snapshotOptions;
+  var snapshotOptions = (options && options.snapshotOptions) || {};
+  if (agent) snapshotOptions.agentCustom = agent.custom;
```
alecgibson, line 395: Out of interest, are we technically supporting `agent` being falsy? Feels a bit weird, given that it's the first argument (which you'd assume is never meant to be omitted).

alecgibson (commented):
Can you also update the docs, please!

lib/backend.js
pypmannetjies: I suppose we are, but that is partially a side effect from how the existing tests were written, where `null` is passed for agent. An option could be to throw an error and change the old tests, if we're saying agent must always be defined 🤔

pypmannetjies:
> Can you also update the docs, please!

I added what I could here: https://github.com/share/sharedb/pull/446/commits/30cb1afd913482b9a36534230598be76982f040b

I guess technically it should be added to `sharedb-mongo` itself as well, but I will hold off on that until I have completed the follow up work to this PR

alecgibson (commented):
Looks good in principle to me. I think I'd like to discuss the whole `if (agent) {}` check in the meeting. I suspect it would be part of a bigger change (raised an issue here: https://github.com/share/sharedb/issues/454)

pypmannetjies:
> Looks good in principle to me. I think I'd like to discuss the whole `if (agent) {}` check in the meeting. I suspect it would be part of a bigger change (raised an issue here: #454)

Are you happy for this to be merged in the meantime?

test/client/query-subscribe.js
```diff
@@ -441,6 +442,48 @@ module.exports = function(options) {
...
+        };
+      });
+
+      it('sends agentCustom to the db\'s getSnapshot call', function(done) {
+        var query = connection.createSubscribeQuery('dogs', this.matchAllDbQuery);
+        var getSnapshotSpy = sinon.spy(this.backend.db, 'getSnapshot');
+
+        query.on('insert', function() {
+          expect(getSnapshotSpy.callCount).to.equal(2);
+          expect(getSnapshotSpy.getCall(1).args[3]).to.deep.equal(expectedArg);
```
ericyhwang, line 463: Maybe we can also test the first (index 0) call from the `create`, which should also have the `agentCustom`. And/or add a comment explaining which call comes from which

alecgibson (approved):

lib/backend.js
pypmannetjies: Discussion can continue here: https://github.com/share/sharedb/issues/454

tiejunhu:
Pass this agent custom value to db adapter's getOps function will be as helpful.

PR #449, opened 2021-03-22
📝 Escape pipe

alecgibson:
Markdown renders a pipe as a table divider, which wasn't what was
intended here, so let's escape it.


pypmannetjies (approved):
LGTM! 👍

PR #450, opened 2021-03-22
📝 Fix links

alecgibson:
According to the [Jekyll docs][1]:

> Since Jekyll 4.0 , you don’t need to prepend `link` and `post_url`
> tags with `site.baseurl`

However, Github Pages is [stuck on v3][2], so we need to prepend all of
our links with `site.baseurl`.

Note that we could potentially just move to using relative paths
(rather than using the `link` tag), but the `link` tag offers some solid
benefits:

 - fails build if the linked page doesn't exist
 - keeps paths relative to the root, which makes updating links much
   easier (eg find-and-replacing a moved file)

[1]: https://jekyllrb.com/docs/liquid/tags/#links
[2]: https://github.com/github/pages-gem/issues/651

### DIFF
--- a/docs/adapters/database.md
+++ b/docs/adapters/database.md
@@ -36,7 +36,7 @@ ShareDB ships with an in-memory, non-persistent database. This is useful for tes
 
 ## Usage
 
-An instance of a database adapter should be provided to the [`Backend()` constructor]({% link api/backend.md %}#backend-constructor)'s `db` option:
+An instance of a database adapter should be provided to the [`Backend()` constructor]({{ site.baseurl }}{% link api/backend.md %}#backend-constructor)'s `db` option:
 
 ```js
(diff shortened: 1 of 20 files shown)


alecgibson:
NB: I'd originally wanted to solve this through a [plugin](https://gist.github.com/thedod/c7b2fd9c2844a604cf07), but according to the [docs](https://docs.github.com/en/github/working-with-github-pages/about-github-pages-and-jekyll#plugins):

> GitHub Pages cannot build sites using unsupported plugins. If you want to use unsupported plugins, generate your site locally and then push your site's static files to GitHub.

I'm not sure it's worth setting up our own build just for this, hence just using the manual method for now.

pypmannetjies (approved):
Sad but necessary :(

PR #452, opened 2021-03-23
Update database.md

Utwo:
I think the original sharedb-postgres is no longer maintained.


alecgibson:
Sounds sensible. Just waiting for confirmation in https://github.com/share/sharedb-postgres/issues/11

Utwo:
https://github.com/share/sharedb-postgres/pull/4#issuecomment-809757140

PR #458, opened 2021-04-07
🥅 Remove agent check from QueryEmitter

alecgibson:
Removes another agent check missed in https://github.com/share/sharedb/pull/457


pypmannetjies (approved):
❤️💚💙💛💜

PR #482, opened 2021-06-14
👷 Run build against latest Node.js and MongoDB

alecgibson:
This change adds a build run for Node.js v16, which was
[released in April][1].

We also run MongoDB tests against the latest version of MongoDB, since
v3.6 was [end-of-lifed in April][2].

[1]: https://nodejs.org/en/about/releases/
[2]: https://www.mongodb.com/support-policy


alecgibson:
NB: I've raised an issue to remove v10 support as a breaking change: https://github.com/share/sharedb/issues/483

.github/workflows/test.yml
```diff
@@ -18,9 +18,10 @@ jobs:
         - 10
         - 12
         - 14
+        - 16
     services:
       mongodb:
-        image: mongo:3.6
+        image: mongo:4.4
```
pypmannetjies, line 24: Just so I understand, this is not a breaking change, because sharedb is "db agnostic", correct?

alecgibson: Yeah, this is specifically just updating the version of MongoDB we use in the [build step](https://github.com/share/sharedb/blob/a837b0090ef4731ca4605d067a40a61e6ef0410c/.github/workflows/test.yml#L45) where we sense-check against `sharedb-mongo` (which should support all current MongoDB versions anyway)

ericyhwang (approved):

PR #486, opened 2021-06-15
💥 Rename `Doc` op events

alecgibson:
The `Doc` op events are currently a little confusing: a `json0` "op" can
be shattered into multiple "component ops".

In the current naming, the "op" emits a `'op batch'` event, and the
"op component" emits an `'op'` event.

However, calling the op an "op batch" is a little bit misleading,
because the "batch" is always an op representing a single version number
increase, and potentially considered a single conceptual change.

For example, a remote client might submit a single op:

```js
[
  {p: ['a'], oi: 'foo'},
  {p: ['b'], oi: 'bar'},
]
```

Under the **current** naming scheme, this emits the following events:

 1 `'before op batch'`: `[{p: ['a'], oi: 'foo'}, {p: ['b'], oi: 'bar'}]`
 2 `'before op'`: `[{p: ['a'], oi: 'foo'}]`
 3 `'op'`: `[{p: ['a'], oi: 'foo'}]`
 4 `'before op'`: `[{p: ['b'], oi: 'bar'}]`
 5 `'op'`: `[{p: ['b'], oi: 'bar'}]`
 6 `'op batch'`: `[{p: ['a'], oi: 'foo'}, {p: ['b'], oi: 'bar'}]`

This can be considered a little surprising, and you may expect the
`'op'` event to be emitted after the whole op (not its shattered
components) have been applied.

Under the **new** naming scheme, the following events are emitted:

 1 `'beforeOp'`: `[{p: ['a'], oi: 'foo'}, {p: ['b'], oi: 'bar'}]`
 2 `'beforeOpComponent'`: `[{p: ['a'], oi: 'foo'}]`
 3 `'opComponent'`: `[{p: ['a'], oi: 'foo'}]`
 4 `'beforeOpComponent'`: `[{p: ['b'], oi: 'bar'}]`
 5 `'opComponent'`: `[{p: ['b'], oi: 'bar'}]`
 6 `'op'`: `[{p: ['a'], oi: 'foo'}, {p: ['b'], oi: 'bar'}]`

 This way, you get the `'op'` event after the whole op has been applied.
 It also makes it more explicit that you're actively listening out for
 the shattered op components where applicable.

 Note that we also move the events to camelCase to be consistent with
 the Backend middleware actions.


alecgibson:
As discussed with @ericyhwang :

- camelCase is fine (consistent with our own Backend, and with Node.js)
- Would be nice to not reuse the `'op'` event, so that we can provide a migration pathway from v1
- Could be nice to unwrap the single-item arrays that we currently pass to the "op component" event
- Roll up 2nd arg into an object
- Naming ideas:
  - `opComponent` & `opComponentsList`
  - `change` & `subChange` / `update` & `partialUpdate`
  - `opApply` & `opComponentApply`

ericyhwang:
Another idea from today:
* `[before]OpApply` and `[before]Json0Component[Apply]`

ericyhwang:
(Was seeing if we could delete the v2 branch, which this PR points to, but that causes the diff to get wonky. Best leave it alone for now, probably.)

PR #469, opened 2021-04-28
🔧 Limit the files distributed in the package

alecgibson:
Fixes https://github.com/share/sharedb/issues/459

This change adds the [`files`][1] property to our `package.json`, which
lets us define which files to publish as part of the distributed package
and reduce the size of our published package from ~2MB to ~100kB

# Before

```
npm notice 📦  sharedb@1.8.1
npm notice === Tarball Contents ===
npm notice 368B    test/.jshintrc
npm notice 1.2kB   docs/Gemfile
npm notice 1.1kB   LICENSE
npm notice 4.4kB   examples/leaderboard/static/leaderboard.css
npm notice 328B    examples/rich-text-presence/static/style.css
npm notice 444.9kB examples/counter-json1/demo.gif
npm notice 444.9kB examples/counter/demo.gif
npm notice 1.1MB   examples/leaderboard/demo.gif
npm notice 398B    docs/404.html
npm notice 343B    examples/counter-json1/static/index.html
npm notice 328B    examples/counter/static/index.html
npm notice 200B    examples/leaderboard/static/index.html
npm notice 562B    examples/rich-text-presence/static/index.html
npm notice 287B    examples/rich-text/static/index.html
npm notice 694B    examples/textarea/static/index.html
npm notice 1.7kB   .eslintrc.js
npm notice 28.3kB  lib/agent.js
npm notice 29.8kB  lib/backend.js
npm notice 5.6kB   test/backend.js
npm notice 2.9kB   test/BasicQueryableMemoryDB.js
npm notice 1.2kB   examples/counter-json1/client.js
npm notice 1.1kB   examples/counter/client.js
npm notice 2.8kB   examples/rich-text-presence/client.js
npm notice 1.1kB   examples/rich-text/client.js
npm notice 1.2kB   examples/textarea/client.js
npm notice 320B    examples/leaderboard/client/connection.js
npm notice 24.8kB  lib/client/connection.js
npm notice 10.3kB  test/client/connection.js
npm notice 2.0kB   test/db-memory.js
npm notice 31.7kB  test/db.js
npm notice 4.2kB   test/client/deserialized-type.js
npm notice 739B    lib/client/presence/doc-presence.js
npm notice 31.7kB  test/client/presence/doc-presence.js
npm notice 35.2kB  lib/client/doc.js
npm notice 17.3kB  test/client/doc.js
npm notice 255B    lib/emitter.js
npm notice 3.8kB   lib/error.js
npm notice 787B    test/error.js
[...]

PR #489, opened 2021-06-30
🔧 Re-include `test/` folder in package

alecgibson:
We recently updated `package.json` to only include `lib/` in our
published package in https://github.com/share/sharedb/pull/469

However, the driver libraries run tests from `test/`, which are no
longer included in the package, which has broken downstream builds.

This change re-includes the `test/` directory to fix those builds.

This approximately doubles our package size. If we're concerned about
this, in future we should consider breaking out a separate testing
library (which may also make our library inter-dependency, and testing
strategy a bit more transparent).

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "sinon": "^7.5.0"
   },
   "files": [
-    "lib/"
+    "lib/",
+    "test/"
   ],
   "scripts": {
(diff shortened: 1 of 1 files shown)


pypmannetjies (approved):
Hmmm... That's unfortunate, but I think it should be a priority to get our other library's tests working again!

PR #492, opened 2021-07-08
♻️ Refactor `ot.applyOps()`

alecgibson:
At the moment, the `ot.applyOps()` duplicates a lot of logic from
`ot.apply()`.

This change refactors `applyOps()` into a simple loop wrapper, and
completely reuses all apply logic from `ot.apply()`.

### DIFF
--- a/lib/ot.js
+++ b/lib/ot.js
@@ -111,7 +111,7 @@ function applyOpEdit(snapshot, edit) {
   try {
     snapshot.data = type.apply(snapshot.data, edit);
   } catch (err) {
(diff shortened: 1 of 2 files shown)


alecgibson:
(I'm pretty sure the coverage drop is just because I've deleted lines)

curran:
Nice change!

ericyhwang (approved):
Did a sanity check, and they look equivalent in functionality, with the benefit of `exports.apply` having some more error checking than `applyOps` did.

Issue #254, opened 2018-11-04
Detailed documentation/test missing for ingestSnapshot


curran:
Currently there are no tests that invoke `ingestSnapshot` directly.

Such tests would be useful in documenting exactly how one could go about using `ingestSnapshot`, which is IMO not clear from the current documentation. For example, after reading "This data must include a version, snapshot and type.", I'm still not sure what to pass in exactly. I'd guess an object with properties `version`, `snapshot` and `type`, but still not sure what kinds of values would be expected.

curran:
In particular, I had difficulty finding out how to construct a `snapshot` from a fetched ShareDB.Doc. I eventually did get it to work, and here's the solution:

```js
const doc = connection.get('examples', 'textarea');
const snapshot = {
  v: doc.version,
  data: doc.data
};
```

This may be a good example to put in the README.

ericyhwang:
Yup, the documentation isn't as precise as it could be. The JSDoc for `ingestSnapshot` is clearer - it says it requires `v`, `data`, and `type` fields on the `snapshot` param.
https://github.com/share/sharedb/blob/v1.0.0-beta.14/lib/client/doc.js#L150

I've updated the README in https://github.com/share/sharedb/commit/a67ad42c384809eb377c9f70c7a45a7552a4be89 to clarify the format for the param.

Any additional clarifications on the documentation that you'd like?

Re the tests, `ingestSnapshot` gets covered indirectly by doc fetch/subscribe. As you say, it could be useful to have a simple test as "enforced" documentation. If you want to add one, that'd be great!

curran:
Thanks a ton @ericyhwang for making that update!

curran:
Closing as the update is IMO suffucient.

PR #493, opened 2021-07-10
Add detail for ingestSnapshot docs

curran:
Greetings,

I was attempting to use `ingestSnapshot` by generating the snapshot on the server and sending it to the client, and I realized that the best documentation for how to create that actual `snapshot` object were found in the comments thread of this issue: https://github.com/share/sharedb/issues/254

I tried to find in the codebase where the `snapshot` object is constructed by looking into the `doc.fetch` internals, but could not find it. So I figured I would contribute this small documentation change.

### DIFF
--- a/docs/api/doc.md
+++ b/docs/api/doc.md
@@ -141,7 +141,7 @@ doc.ingestSnapshot(snapshot [, callback])
 ```
 
 {: .warn }
-This method is generally called internally as a result of [`fetch()`](#fetch) or [`subscribe()`](#subscribe), and not directly from consumer code. Consumers *may* want to use this method to ingest data that was transferred to the client externally to the client's ShareDB connection.
+This method is generally called internally as a result of [`fetch()`](#fetch) or [`subscribe()`](#subscribe), and not directly from consumer code. Consumers *may* want to use this method to ingest data that was transferred to the client externally to the client's ShareDB connection. To generate a snapshot from a fetched `Doc`, you can do the following: `const snapshot = { v: doc.version, data: doc.data, type: doc.type }`.
 
 
 `snapshot` -- [Snapshot]({{ site.baseurl }}{% link api/snapshot.md %})


docs/api/doc.md
```diff
@@ -141,7 +141,7 @@ doc.ingestSnapshot(snapshot [, callback])
 ```
 
 {: .warn }
-This method is generally called internally as a result of [`fetch()`](#fetch) or [`subscribe()`](#subscribe), and not directly from consumer code. Consumers *may* want to use this method to ingest data that was transferred to the client externally to the client's ShareDB connection.
+This method is generally called internally as a result of [`fetch()`](#fetch) or [`subscribe()`](#subscribe), and not directly from consumer code. Consumers *may* want to use this method to ingest data that was transferred to the client externally to the client's ShareDB connection. To generate a snapshot from a fetched `Doc`, you can do the following: `const snapshot = { v: doc.version, data: doc.data, type: doc.type }`.
```
alecgibson, line 144: I think this isn't quite correct:

 - `doc.type` is an `object`
 - `snapshot.type` is a `string` (`doc.type.uri`)

alecgibson (commented):
I'm curious; what's the use-case here? Should we add a utility method to `Doc`?

curran:
Thanks for reviewing!

>I'm curious; what's the use-case here?

The use case:
 * Server-side code runs a ShareDB fetch operation.
 * That data gets baked into the server-rendered HTML.
 * That data gets parsed in the client and passed into `ingestSnapshot`
 * _so that_ the data is not fetched again during a subscribe.

In general, the need is to derive a well formed snapshot from an already fetched Doc (on the server), so that it can be passed into `ingestSnapshot` (on the client). For a given page load, this would reduce the number of database queries for the same document from 2 (server side fetch, client side subscribe without `ingestSnapshot`) to 1.

>Should we add a utility method to Doc?

Yes please! That would be amazing.

curran:
Oh hey! I have some public code I can reference as a concrete use case:

 * [getShareDBSnapshot](https://github.com/vizhub-open-core/vizhub-open-core/blob/1dcc2c33896e088c545023a8bd7940e535dc6235/packages/vizhub-core/src/plugins/vizPagePlugin/getShareDBSnapshot.js)
 * That's used in a [server side Express route](https://github.com/vizhub-open-core/vizhub-open-core/blob/1dcc2c33896e088c545023a8bd7940e535dc6235/packages/vizhub-core/src/plugins/vizPagePlugin/server.js#L22).
 * That route [encodes the snapshot in the HTML](https://github.com/vizhub-open-core/vizhub-open-core/blob/1dcc2c33896e088c545023a8bd7940e535dc6235/packages/vizhub-core/src/plugins/vizPagePlugin/indexHTML.js#L59)
 * That snapshot is then [decoded in the client](https://github.com/vizhub-open-core/vizhub-open-core/blob/1dcc2c33896e088c545023a8bd7940e535dc6235/packages/vizhub-core/src/client/index.js#L9) 
 * It can be then [passed into ingestSnapshot](https://github.com/vizhub-open-core/vizhub-open-core/blob/1dcc2c33896e088c545023a8bd7940e535dc6235/packages/vizhub-core/src/plugins/vizPagePlugin/VizContext.js#L12) to avoid fetching the same data again when we subscribe to the Doc.

This is all very WIP so bear with me. I hope what I'm trying to do makes sense. Thanks!

docs/api/doc.md
alecgibson: The other thing you could do is just cross-link to the [`Snapshot`](https://share.github.io/sharedb/api/snapshot) docs

curran:
Closing in favor of #508.

PR #496, opened 2021-07-14
🥅 Allow fetching historic snapshots with bad `json0` paths

alecgibson:
`json0` recently merged a [breaking change][1] which enforces some
stricter type checking.

Apart from this stricter type checking, no other changes were made, so
any client that only ever submits well-formed ops should be able to
upgrade directly without any trouble.

However, if clients submitted some bad ops (that are no longer allowed
under the stricter checking), then `fetchSnapshot()` will fail when
trying to apply these ops to rebuild the snapshot.

This failure can be surprising if the only "bad" thing about the ops
was that they had [bad path types][2], because at the time when they
were submitted, the snapshot would have been correctly updated.

This change rescues from this particular failure by coercing paths into
the correct type: `number` for arrays, and `string` for objects. Note
that we use the exact same check for arrays and objects as `json0` to
ensure consistency.

Also note that we don't attempt to rescue new ops being submitted,
because these should correctly be rejected.

[1]: https://github.com/ottypes/json0/pull/40
[2]: https://github.com/ottypes/json0/issues/37

### DIFF
--- a/lib/ot.js
+++ b/lib/ot.js
@@ -208,12 +208,12 @@ exports.transformPresence = function(presence, op, isOwnOp) {
  * have this stricter validation. This method fixes up the op paths to
  * pass the stricter validation
  */
-function normalizeJson0Paths(snapshot, json0op) {
+function normalizeJson0Paths(snapshot, json0Op) {
   if (snapshot.type !== types.defaultType.uri) return;
-  var ops = json0op.op;
-  if (!ops) return;
-  for (var i = 0; i < ops.length; i++) {
-    var path = ops[i].p;
+  var components = json0Op.op;
+  if (!components) return;
+  for (var i = 0; i < components.length; i++) {
+    var path = components[i].p;
     var element = snapshot.data;
     for (var j = 0; j < path.length; j++) {
       var key = path[j];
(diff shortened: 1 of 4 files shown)


lib/ot.js
```diff
@@ -197,3 +200,28 @@ exports.transformPresence = function(presence, op, isOwnOp) {
...
 };
+
+/**
+ * json0 had a breaking change in https://github.com/ottypes/json0/pull/40
+ * The change added stricter type checking, which breaks fetchSnapshot()
+ * when trying to rebuild a snapshot from old, committed ops that didn't
+ * have this stricter validation. This method fixes up the op paths to
+ * pass the stricter validation
+ */
+function normalizeJson0Paths(snapshot, json0op) {
```
ericyhwang, line 211: `json0Op` for correct lowerCamelCase

lib/ot.js
```diff
@@ -197,3 +200,28 @@ exports.transformPresence = function(presence, op, isOwnOp) {
...
+  if (snapshot.type !== types.defaultType.uri) return;
+  var ops = json0op.op;
```
ericyhwang, line 213: Change to either `opComponents` or `components` to distinguish between this and "top-level" ops

ericyhwang (approved):

Issue #503, opened 2021-07-29
Flaky Client Projections Test


dwinrick-lever:
I have seen this test fail, and then pass when ran again with CI (working on sharedb-mongo).

`'client projections' -> 'subscribe query' -> 'query subscribe on projection will update based on fields not in projection'`

![Screen Shot 2021-07-29 at 11 24 47 AM](https://user-images.githubusercontent.com/83318093/127549791-469181f1-9473-48b6-85ad-cec6b352b9c6.png)

PR #504, opened 2021-07-30
✅ Fix flaky test

alecgibson:
Fixes https://github.com/share/sharedb/issues/503

This change fixes a flaky test, which relies on a `setTimeout` with a
10ms wait.

This change replaces the use of a `setTimeout` with an event listener,
which is not prone to flakiness, and is more semantic.

### DIFF
--- a/test/client/projections.js
+++ b/test/client/projections.js
@@ -220,17 +220,16 @@ module.exports = function(options) {
             if (err) return done(err);
             expect(results).to.have.length(0);
 
+            query.once('changed', function() {
+              expect(query.results).to.have.length(1);
+              expect(query.results[0].id).to.equal('fido');
+              expect(query.results[0].data).to.eql({age: 3, owner: {name: 'jim'}});
+              done();
+            });
+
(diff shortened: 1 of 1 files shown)


dwinrick-lever (approved):

test/client/projections.js
```diff
@@ -220,17 +220,16 @@ module.exports = function(options) {
...
-            fido.submitOp({p: ['color'], od: 'gold', oi: 'black'}, null, function(err) {
-              if (err) return done(err);
-              setTimeout(function() {
-                expect(query.results).to.have.length(1);
-                expect(query.results[0].id).to.equal('fido');
-                expect(query.results[0].data).to.eql({age: 3, owner: {name: 'jim'}});
-                done();
-              }, 10);
-            });
+            fido.submitOp({p: ['color'], od: 'gold', oi: 'black'}, null, util.errorHandler(done));
```
ericyhwang, line 232: I didn't realize we had `util.errorHandler`!

ericyhwang (approved):

Issue #502, opened 2021-07-29
Flaky Query Subscribe Test


dwinrick-lever:
I've found this test to periodically fail:

- File: `test/client/query-subscribe.js`
- Test: `subscribed query gets simultaneous insert and remove after reconnecting`

This has happened multiple times when contributing to https://github.com/share/sharedb-mongo, to the point where the PR build have to be re-ran 3, 4 times.

```
  1) db
       client query subscribe
         subscribed query gets simultaneous insert and remove after reconnecting:
     Error: Already closed (and Mocha's done() called multiple times)
      at wrapErrorData (node_modules/sharedb/lib/client/connection.js:263:13)
      at Connection.handleMessage (node_modules/sharedb/lib/client/connection.js:191:11)
      at StreamSocket.socket.onmessage (node_modules/sharedb/lib/client/connection.js:142:18)
      at /home/runner/work/sharedb-mongo/sharedb-mongo/node_modules/sharedb/lib/stream-socket.js:60:12
      at processTicksAndRejections (node:internal/process/task_queues:78:11)
```

I haven't gotten it to fail locally. :D :D

PR #505, opened 2021-07-30
✅ Try to fix flaky query subscribe test

alecgibson:
This is a speculative fix for https://github.com/share/sharedb/issues/502

It seems impossible to get this test to fail locally, so this change
makes some generic quality-of-life improvements on the test code in the
hopes that we improve its stability:

 - wait for the insert and removes to be acked on `connection2` before
   reconnection `connection`
 - catch `del()` errors
 - change the query listeners to `.once()` to avoid multiple calls

### DIFF
--- a/test/client/query-subscribe.js
+++ b/test/client/query-subscribe.js
@@ -216,19 +216,28 @@ module.exports = function(options) {
         }
       ], function(err) {
         if (err) return done(err);
-        var query = connection.createSubscribeQuery('dogs', matchAllDbQuery, null, function(err) {
-          if (err) return done(err);
+        var query = connection.createSubscribeQuery('dogs', matchAllDbQuery);
+        query.on('error', done);
+
+        query.once('ready', function() {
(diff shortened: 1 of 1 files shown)


dwinrick-lever (approved):

test/client/query-subscribe.js
```diff
@@ -216,19 +216,28 @@ module.exports = function(options) {
         }
       ], function(err) {
         if (err) return done(err);
-        var query = connection.createSubscribeQuery('dogs', matchAllDbQuery, null, function(err) {
-          if (err) return done(err);
+        var query = connection.createSubscribeQuery('dogs', matchAllDbQuery);
+        query.on('error', done);
```
ericyhwang, line 220: If we only care about the first error from this test case, it may make sense to use `.once('error', done)` here as well

ericyhwang (approved):

dwinrick-lever:
🥳

PR #508, opened 2021-08-22
Add Doc.toSnapshot

### LINKED ISSUE #500: Doc utility method for creating a snapshot
Tracking this idea out of a discussion in https://github.com/share/sharedb/pull/493

As a user of ShareDB in the context of an isomorphic JavaScript app, I want to be able to use `ingestSnapshot` to ingest data that was transferred to the client externally to the client's ShareDB connection, so that the number of times a given document is queried for in the database and transferred over the wire is reduced from 2 (server side fetch, client side subscribe without `ingestSnapshot`) to 1 (server side fetch only).

Proposed API: `Doc.createSnapshot()`. This makes sense to invoke only after a successful fetch or subscribe.

curran:
Fix #500

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -920,18 +920,19 @@ Doc.prototype.resume = function() {
 };
 
 // Create a snapshot that can be serialized, deserialized, and passed into `Doc.ingestSnapshot`.
-Doc.prototype.toSnapshot = function() {
+// By default, this implementation defensively clones the document data
+// to guard against mutations. You can opt out of cloning by passing `true`
+// as the optional `doNotClone` argument.
+//
+// @param doNotClone (optional) If `true`, the document data is not cloned. This can be used
+//        as a performance optimization, but be sure that your code does not mutate the data.
+Doc.prototype.toSnapshot = function(doNotClone) {
   return {
     v: this.version,
-    data: clone(this.data),
+    data: doNotClone ? this.data : util.clone(this.data),
     type: this.type.uri
   };
 };
-
-// TODO refactor this so there's 1 copy in the codebase, not 3.
-function clone(obj) {
-  return (obj === undefined) ? undefined : JSON.parse(JSON.stringify(obj));
-}
 
 // *** Receiving operations
 
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -921,16 +921,10 @@ Doc.prototype.resume = function() {
 };
 
 // Create a snapshot that can be serialized, deserialized, and passed into `Doc.ingestSnapshot`.
-// By default, this implementation defensively clones the document data
-// to guard against mutations. You can opt out of cloning by passing `true`
-// as the optional `doNotClone` argument.
-//
-// @param doNotClone (optional) If `true`, the document data is not cloned. This can be used
-//        as a performance optimization, but be sure that your code does not mutate the data.
-Doc.prototype.toSnapshot = function(doNotClone) {
+Doc.prototype.toSnapshot = function() {
   return {
     v: this.version,
-    data: doNotClone ? this.data : clone(this.data),
+    data: clone(this.data),
     type: this.type.uri
   };
 };
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -923,10 +923,15 @@ Doc.prototype.resume = function() {
 Doc.prototype.toSnapshot = function() {
   return {
     v: this.version,
-    data: this.data,
+    data: clone(this.data),
     type: this.type.uri
   };
 };
+
+// TODO refactor this so there's 1 copy in the codebase, not 3.
+function clone(obj) {
+  return (obj === undefined) ? undefined : JSON.parse(JSON.stringify(obj));
+}
 
 // *** Receiving operations
 
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -920,13 +920,11 @@ Doc.prototype.resume = function() {
 };
 
 // Create a snapshot that can be serialized, deserialized, and passed into `Doc.ingestSnapshot`.
-//
-// Note that types.defaultType is used inside ingestSnapshot when snapshot.type is omitted.
-// If you need to use a different type, populate `snapshot.type` with the OT type you are using.
-Doc.prototype.createSnapshot = function() {
+Doc.prototype.toSnapshot = function() {
   return {
     v: this.version,
     data: this.data,
+    type: this.type.uri
   };
 };
 
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -3,6 +3,7 @@ var logger = require('../logger');
 var ShareDBError = require('../error');
 var types = require('../types');
 var util = require('../util');
+var clone = util.clone;
 var deepEqual = require('fast-deep-equal');
 
 var ERROR_CODE = ShareDBError.CODES;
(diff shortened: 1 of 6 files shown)


curran:
CI is failing for what looks like a silly reason unrelated to the changes here.

image

Any ideas on how to turn CI green?

lib/client/doc.js
```diff
@@ -919,6 +919,17 @@ Doc.prototype.resume = function() {
...
 };
 
+// Create a snapshot that can be serialized, deserialized, and passed into `Doc.ingestSnapshot`.
+//
+// Note that types.defaultType is used inside ingestSnapshot when snapshot.type is omitted.
+// If you need to use a different type, populate `snapshot.type` with the OT type you are using.
+Doc.prototype.createSnapshot = function() {
+  return {
+    v: this.version,
+    data: this.data,
```
alecgibson, line 929: We should probably clone `data`, because if an op comes in, it'll actually mutate the snapshot's data (since they share a reference), which should never change.

Can we also add a test case for this, so we don't accidentally do this in future?

test/client/doc.js
```diff
@@ -475,4 +475,19 @@ describe('Doc', function() {
...
   });
+
+  describe('snapshot generation', function() {
+    var doc;
+    beforeEach(function(done) {
+      doc = this.connection.get('dogs', 'scooby');
+      doc.create({name: 'Scooby'}, function(error) {
+        if (error) return done(error);
+        doc.whenNothingPending(done);
+      });
```
alecgibson, line 486:
```suggestion
      doc.create({name: 'Scooby'}, done);
```

We can probably simplify this setup — just less stuff to think about in future when looking at this test.

lib/client/doc.js
```diff
@@ -919,6 +919,17 @@ Doc.prototype.resume = function() {
   this.flush();
 };
 
+// Create a snapshot that can be serialized, deserialized, and passed into `Doc.ingestSnapshot`.
+//
+// Note that types.defaultType is used inside ingestSnapshot when snapshot.type is omitted.
+// If you need to use a different type, populate `snapshot.type` with the OT type you are using.
```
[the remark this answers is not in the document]
curran: It's because it's not serializable.

lib/client/doc.js
```diff
@@ -919,6 +919,17 @@ Doc.prototype.resume = function() {
...
+Doc.prototype.createSnapshot = function() {
```
ericyhwang, line 926: Since `createSnapshot` sounds like it might have a side effect, how about `asSnapshot` or `toSnapshot`?

lib/client/doc.js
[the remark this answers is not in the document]
curran: Note from call: We can use the URI of the type, so it's serializable.

The docs could be more specific on this - @alecgibson  is on it.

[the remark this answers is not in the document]
curran: Added type.

lib/client/doc.js
curran: Renamed.

lib/client/doc.js
curran: Added a failing test.

curran:
What is the suggested approach for doing the cloning? I see there are two existing implementations of `clone` in the codebase:

https://github.com/share/sharedb/blob/3e636d0100921c26b4bb5108ef0afd1b27f838af/lib/db/memory.js#L176

```js
function clone(obj) {
  return (obj === undefined) ? undefined : JSON.parse(JSON.stringify(obj));
}
```
This is not exported.

The same implementation is found in https://github.com/share/sharedb/blob/master/test/projections.js#L389

My instinct is to refactor `clone` into its own module, and import that into both of these existing places where it's used, and also import it into Doc where I can use it to clone for the snapshot generation.

Alternatively, I could copy-paste the implementation into Doc.js. That would make 3 copies of the same implementation.

Thoughts?

curran:
CI is failing due to something seemingly unrelated to these changes.

image

https://github.com/share/sharedb/pull/508/checks?check_run_id=3475145150

curran:
Actually, I'm not sold on the `clone` operation. I would favor calling out in the docs "Do not mutate the returned value", as cloning documents like this, every time a snapshot is created, is a huge waste of CPU resources.

curran:
Notes from call:
 * Viable option: add an optional flag to opt out of cloning step
 * We can put clone in the utils module

curran:
Requesting re-review @alecgibson @ericyhwang . Thanks!

CI is failing on ShareDB Mongo, not sure why (and not sure why this would not be the case in `master`).

lib/client/doc.js
```diff
@@ -919,6 +920,21 @@ Doc.prototype.resume = function() {
   this.flush();
 };
 
+// Create a snapshot that can be serialized, deserialized, and passed into `Doc.ingestSnapshot`.
+// By default, this implementation defensively clones the document data
+// to guard against mutations. You can opt out of cloning by passing `true`
+// as the optional `doNotClone` argument.
+//
+// @param doNotClone (optional) If `true`, the document data is not cloned. This can be used
```
alecgibson, line 928: I'm not sure we should let people do this. I know cloning is a performance overhead, but the failure mode is bad if consumers accidentally mutate the snapshot (because they'll mutate `doc.data`, which shouldn't happen outside of OT).

lib/util.js
```diff
@@ -96,3 +96,8 @@ exports.nextTick = function(callback) {
     callback.apply(null, args);
   });
 };
+
+exports.clone = function(obj) {
+  return (obj === undefined) ? undefined : JSON.parse(JSON.stringify(obj));
+};
```
alecgibson, line 102: This code has only hitherto been used in tests, and I wonder if we should use something more performant in production code?

If we're okay with adding a dependency, we've been using [`rfdc`](https://www.npmjs.com/package/rfdc) and it's worked well.

On the other hand, the performance gain is (relatively) small (~3x speed-up), so I don't mind just sticking with this.

@ericyhwang thoughts?

lib/client/doc.js
curran: The main use case for this is to serialize the snapshot and send it over the network. I'm of the opinion that consumers should be allowed to opt out of cloning if they know for sure that their code will never mutate the data.

lib/util.js
curran: >This code has only hitherto been used in tests

It was also used in `lib/db/memory.js`.

alecgibson: The `MemoryDB` [shouldn't be used in production](https://github.com/share/sharedb/blob/3e636d0100921c26b4bb5108ef0afd1b27f838af/lib/db/memory.js#L7), so this clone method still hasn't touched production code before now.

lib/client/doc.js
alecgibson: I still personally think that the risks of weird side-effects massively outweigh the relatively small performance impact of cloning. Cloning is a perfectly normal, defensive thing to do. `json0` itself uses clone in many places through its lifecycle.

(As a side-note, `json0` is just using a [`JSON` approach to cloning](https://github.com/ottypes/json0/blob/90a3ae26364c4fa3b19b6df34dad46707a704421/lib/json0.js#L42-L44), although I believe the comment about it being fast is out-of-date).

alecgibson:
@curran we've dropped Node.js v10 support now. You should be able to rebase and greenify the build.

lib/client/doc.js
curran: Got it. That simplifies things. Will do!

curran:
Addressed feedback and merged latest from `master`. Requesting re-review please @alecgibson @ericyhwang . Thanks!

alecgibson (approved):
Nice one!

curran:
Hooray! Thank you.

Issue #509, opened 2021-08-22
Make CI Sane


curran:
CI currently runs `npm run lint`, which is failing on `master`.

This causes all PRs to fail CI, such as [this one](https://github.com/share/sharedb/pull/508/checks?check_run_id=3393270552#step:6:24).

curran:
Still not quite sane.

![image](https://user-images.githubusercontent.com/68416/131561481-3200cb46-7eea-4c7c-a57a-d2c6efbb3971.png)

https://github.com/share/sharedb/pull/508/checks?check_run_id=3475145150

alecgibson:
Yeah we need to drop Node.js 10 support in a breaking change

curran:
Got it. Thanks!

Related https://github.com/share/sharedb/issues/483

PR #510, opened 2021-08-23
🚨 Ignore docs when linting

alecgibson:
Fixes https://github.com/share/sharedb/issues/509

For some reason, the build has started linting our `/docs/` folder. It's
possible that this was an `eslint` bug, and that it should have always
been linted, since we just use our `.gitignore` file (and the JS in our
docs folder is ignored by its own, nested `.gitignore` file).

This change bumps our version of `eslint`, and adds the `/docs`/ folder
separately to be ignored by `eslint`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,12 +13,12 @@
   "devDependencies": {
     "chai": "^4.2.0",
     "coveralls": "^3.0.7",
-    "eslint": "^6.5.1",
+    "eslint": "^7.32.0",
     "eslint-config-google": "^0.14.0",
-    "ot-json0-v2": "ottypes/json0",
     "lolex": "^5.1.1",
     "mocha": "^8.2.1",
     "nyc": "^14.1.1",
+    "ot-json0-v2": "ottypes/json0",
     "ot-json1": "^0.3.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,12 +13,12 @@
...
     "chai": "^4.2.0",
     "coveralls": "^3.0.7",
-    "eslint": "^6.5.1",
+    "eslint": "^7.32.0",
     "eslint-config-google": "^0.14.0",
-    "ot-json0-v2": "ottypes/json0",
     "lolex": "^5.1.1",
     "mocha": "^8.2.1",
     "nyc": "^14.1.1",
+    "ot-json0-v2": "ottypes/json0",
```
pypmannetjies, line 21: Interesting, why did we need to add this?

alecgibson: Oh it just got moved when I ran `npm install` — I'd added this by hand before, and I guess I don't alphabet so good 🤦🏼‍♂️

pypmannetjies: Oh right and I don't read diffs good

pypmannetjies (approved):
LGTM

Issue #512, opened 2021-08-24
Pass user information with the op to remote clients.


maxfortun:
Use case: Visually track change attribution across remote clients. 
  
[#224] allows to store `metadata`, which contains `userId`, but falls short of pushing `metadata` out to remote clients.

### On a server side:  

Retain backend constructor options in order to pass around `options.opsOptions` to configure metadata behavior.  

Include `message.m` in agent's `_sendOp` based on `backend.options.opsOptions.metadata` configuration.  

Include `op.m` in SubmitRequest  based on `backend.options.opsOptions.metadata` configuration.  

### On a client side:  

Emit `op.m` with every `op` related event.
``` 
this.emit('op', componentOp.op, source, op.src, op.m);
```

alecgibson:
I think it might be possible to get this to work without any modifications. You're allowed to set an arbitrary ID on your `Connection`:

```js
connection = nw Connection(socket);
// Concat with a `uuid` in case the same user has multiple sessions, since
// the Connection ID should be unique per-connection
connection.id = [userId, uuid()].join(':');
```

Then in remote clients, on the `'op'` event, you can check the `src`:

```js
doc.on('op', (op, source, src) => {
  const [userId] = src.split(':');
  // Do something with userId...
});
```

maxfortun:
@alecgibson , thank you for a quick response! I will experiment with your suggestion and it may be good enough for now. I was using `userId` as a simplified example and was ultimately hoping to pass around a more sophisticated metadata object with multiple fields. Code suggests that that's what `op.m` is intended for, so I researched that vector. Using a connection id is a cool idea.

alecgibson:
Yeah that's fair; I've long thought we should be able to facilitate getting op metadata to clients!

PR #513, opened 2021-08-24
Pass user information with the op to remote clients.

maxfortun:
Adds feature [#512] 
---
Use case: Visually track change attribution across remote clients. 
  
[#224] allows to store `metadata`, which contains `userId`, but falls short of pushing `metadata` out to remote clients.

### On a server side:  

Retain backend constructor options in order to pass around `options.opsOptions` to configure metadata behavior.  

Include `message.m` in agent's `_sendOp` based on `backend.options.opsOptions.metadata` configuration.  

Include `op.m` in SubmitRequest  based on `backend.options.opsOptions.metadata` configuration.  

### On a client side:  

Emit `op.m` with every `op` related event.
``` 
this.emit('op', componentOp.op, source, op.src, op.m);
```

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -232,7 +232,7 @@ GEM
       jekyll-seo-tag (~> 2.1)
     minitest (5.15.0)
     multipart-post (2.1.1)
-    nokogiri (1.13.7)
+    nokogiri (1.13.9)
       mini_portile2 (~> 2.8.0)
       racc (~> 1.4)
     octokit (4.22.0)


lib/agent.js
```diff
@@ -260,7 +260,9 @@ Agent.prototype._sendOp = function(collection, id, op) {
   if ('op' in op) message.op = op.op;
   if (op.create) message.create = op.create;
   if (op.del) message.del = true;
-
+  if(op.m && this.backend.options && this.backend.options.opsOptions && this.backend.options.opsOptions.metadata) {
```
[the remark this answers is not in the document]
maxfortun: One question, is there a way to reach `context` from `Agent.prototype._sendOp`? If not, we'd need to set `context.agent.opMetadataProjection`.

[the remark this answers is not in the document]
alecgibson: Why? I'm not sure I'm clear on the scope here: are we trying to get metadata to the client for newly-submitted ops, or for _all_ ops?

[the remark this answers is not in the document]
maxfortun: The confusion may be due to my lack of familiarity with the code. Please forgive me. It looks like in order to be able to generate an outbound projection inside `Agent.prototype._sendOp`, it needs to have access to the projection configuration set in `context.opMetadataProjection`. Are you suggesting to always `message.m = Object.assign({}, op.m);` inside  `Agent.prototype._sendOp` and handle the actual projection inside commit/submit ?

lib/client/doc.js
```diff
@@ -620,28 +621,28 @@ Doc.prototype._otApply = function(op, source) {
           if (transformErr) return this._hardRollback(transformErr);
         }
         // Apply the individual op component
-        this.emit('before op', componentOp.op, source, op.src);
+        this.emit('before op', componentOp.op, source, op.src, op.m);
```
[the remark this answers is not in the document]
maxfortun: Would you be open to this signature, where the 3rd param is op.src, and 4th param is the whole op? This way whoever needs whatever from op have access to that whatever?
```javascript
this.emit('before op', componentOp.op, source, op.src, op);
```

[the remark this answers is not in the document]
alecgibson: I think @ericyhwang had suggested something similar to this in the past. I'm not necessarily against it, although I'd potentially still bury it in a wrapper object in case we need more arguments

lib/submit-request.js
```diff
@@ -171,7 +171,6 @@ SubmitRequest.prototype.commit = function(callback) {
           var op = request.op;
           op.c = request.collection;
           op.d = request.id;
-          op.m = undefined;
```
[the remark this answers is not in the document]
maxfortun: Hi @alecgibson , did you have a weekly PR meeting on Tuesday?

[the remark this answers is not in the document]
maxfortun: Thank you.

lib/backend.js
```diff
@@ -197,7 +197,7 @@ Backend.prototype.trigger = function(action, agent, request, callback) {
     if (err) return callback(err);
     var fn = fns.shift();
     if (!fn) return callback();
-    fn(request, next);
+    fn(request, callback);
```
alecgibson, line 200: Can we move this change to a separate PR — I haven't got the headspace to think about it right now (hopefully a bit more time next week), but I think:

- it's not conceptually part of this change
- it might need further discussion, and I don't want the two changes to hold one another up

maxfortun: I am at conferences this and next week, so will not have time to refactor:( Will go through this right after.

maxfortun: Hi @alecgibson , I am a bit puzzled as I am not sure how to create 2 separate PRs from a single fork. Any ideas?

alecgibson: You should be able to just create a second branch on your fork, and raise a PR from there. You don't need to raise PRs from `master`.

maxfortun: @alecgibson , thank you. Moved into a separate PR [#517].

lib/submit-request.js
```diff
@@ -185,6 +185,38 @@ SubmitRequest.prototype.commit = function(callback) {
...
+    return undefined;
+  }
+
+  // Granular projection
+  if (typeof request.opMetadataProjection === 'object') {
+    return request._granularMetadataProjection();
+  }
+
+  // Full projection
+  return request.op.m;
```
ericyhwang, line 202: Did you specifically need the ability to get all metadata?

If it's not something you need at the moment, I'd leave this out for now, returning `undefined` as the fallback case.

Requesting just what's needed is better, so if we leave out a "get-everything" option, that removes the temptation to request all the metadata.

ericyhwang (commented):
Overall approach sounds fine to me.

Please add some tests for the projection - you can run the test suite locally with `npm test`.

We recommend adding your tests inside test/client/submit.js, which tests the codepath you're modifying.

maxfortun:
@ericyhwang I added a few tests to test/client/submit.js. Hopefully these are acceptable.

test/client/submit.js
```diff
@@ -1210,5 +1212,180 @@ module.exports = function() {
...
+        });
+
+        var connection = this.backend.connect(undefined, metadata);
+        var doc = null;
+        connection.on('connected', function() {
+          expect(connection.agent.custom).eql(metadata);
+          doc = connection.get('dogs', 'fido');
+          doc.create({name: 'fido'}, function() {
+            doc.on('op', function(op, source, src, context) {
+              expect(context.op.m).equal(undefined);
```
alecgibson, line 1278: I'm assuming this is acceptable for your use case?

test/client/submit.js
```diff
@@ -1210,5 +1212,180 @@ module.exports = function() {
...
+              done();
+            });
+            doc.submitOp([{p: ['tricks'], oi: ['fetch']}], {source: 'trainer'}, errorHandler(done));
+          });
+        });
+      });
+
+      it('concurrent changes', function(done) {
```
alecgibson, line 1286: I'm not sure I understand what we're trying to test here?

We also seem to be missing a simple test that remote clients receive the projected metadata (the whole point of this PR!).

Issue #483, opened 2021-06-14
Drop Node v10 support


alecgibson:
Node v10 has been [end-of-lifed](https://nodejs.org/en/about/releases/), so we should drop it from our build matrix, but this would be a breaking change.

PR #518, opened 2021-09-21
💥 Drop Node v10 support

alecgibson:
Fixes https://github.com/share/sharedb/issues/483

This is a **BREAKING CHANGE** which drops official Node.js v10 support,
which has been [end-of-lifed][1].

[1]: https://nodejs.org/en/about/releases/


ericyhwang (approved):
👍

Issue #451, opened 2021-03-22
Default to hard rollback


alecgibson:
In general, the `invert()` method is not very reliable as an API. Neither `json0` and `json1` guarantee that any given op is invertible. This can lead to surprising data states when using a "soft" rollback. We should probably default to hard rollback in all cases, and force consumers to opt-in if they want this behaviour (which means they should also hopefully know what they're doing).

Further discussion:
 - https://github.com/share/sharedb/issues/442
 - https://github.com/share/sharedb/pull/443

alecgibson:
Discussed with @ericyhwang — sounds like a sensible thing to do.

Flag-wise, we can add one at the global level to help consumers ease transition, and also set per-doc for more granular control. We'd need to have "ternary" logic at the doc level: `null` means inherit from the global flag, or if we've actively set it to `true` or `false`, then we should use the doc flag.

curran:
This seems like a good direction to me.

hsource:
I did a more targeted and backwards-compatible fix in #520 - Fall back to doing a _hardRollback when op.type.invert fails. Let me know if you have any feedback on that! We're running that in a patched production build already

PR #520, opened 2021-10-04
Fall back to doing a _hardRollback when `op.type.invert` fails

### LINKED ISSUE #442: Do not `invert` automatically in `Doc.prototype._rollback`

hsource:
## Motivation

When we had rollbacks due to conflicts on our server, the rollback would fail when the rejected ops were `rich-text` ops, which aren't invertible due to the error in #442.

This fix is a less breaking change than what's proposed in #442, #443, and #451, and can potentially close those issues. Unlike those changes, this change:

1. Doesn't require any changes in libraries to support it
2. Only changes the behaviour in cases that are totally broken today

Closes #442.

## Fix

When the ops fail to invert, just fall back to a hard rollback (refetching the data from the server)

## Testing

Added automated test for this case that failed before, but passes now.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -948,7 +948,13 @@ Doc.prototype._rollback = function(err) {
   var op = this.inflightOp;
 
   if ('op' in op && op.type.invert) {
-    op.op = op.type.invert(op.op);
+    try {
+      op.op = op.type.invert(op.op);
+    } catch (error) {
+      // If the op doesn't support `.invert()`, we just reload the doc
+      // instead of trying to locally revert it.
+      return this._hardRollback(err);
+    }
 
     // Transform the undo operation by any pending ops.
     for (var i = 0; i < this.pendingOps.length; i++) {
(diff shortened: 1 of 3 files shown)


alecgibson:
I think wrapping this in a try/catch is sensible, but I also think this is a bug in `json0`? I'm right in thinking this doesn't happen if you use a "pure" (non-sub-typed) `rich-text` document, right? Which is because `json0` [doesn't check](https://github.com/ottypes/json0/blob/90a3ae26364c4fa3b19b6df34dad46707a704421/lib/json0.js#L71-L75) if its subtypes support `invert`

PR #522, opened 2021-11-19
✨ Add `receivePresence` middleware hook

alecgibson:
At the moment, we have a [`sendPresence`][1] middleware hook, which is
triggered just before a presence update is about to be sent to a client.

This existing hook is useful for determining if a client should receive
an update or not on the server.

However, it doesn't let us *stop* clients from broadcasting to other
clients, which we may want to do for authentication reasons, for
example.

This change adds a new `receivePresence` middleware hook, which is
triggered when the server receives a presence message from a client,
before the presence has been transformed against any ops.

Passing an error to this middleware hook will return an error message to
the broadcasting client (as opposed to `sendPresence`, which will send
an error down to any subscribed clients).

[1]: https://github.com/share/sharedb/blob/07f8bd98e5cdae37e925d4c0c709a1c3cbbca5e0/lib/agent.js#L816

### DIFF
--- a/docs/middleware/actions.md
+++ b/docs/middleware/actions.md
@@ -99,6 +99,20 @@ This action has these additional `context` properties:
 
 > The reply about to be sent
 
+## `'receivePresence'`
+
+Presence information has just been received from a client. The presence has not yet been transformed against any ops it has missed.
+
+This action has these additional `context` properties:
+
+`collection` -- string
+
+> The collection the presence is associated with
+
+`presence` -- Object
+
+> The presence object that was received. Its shape depends on its [type]({{ site.baseurl }}{% link types/index.md %})
+
 ## `'sendPresence'`
 
 Presence information is about to be sent to a client.
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -736,17 +736,25 @@ Agent.prototype._src = function() {
 
 Agent.prototype._broadcastPresence = function(presence, callback) {
   var agent = this;
+  var backend = this.backend;
   var requests = this.presenceRequests[presence.ch] || (this.presenceRequests[presence.ch] = {});
   var previousRequest = requests[presence.id];
   if (!previousRequest || previousRequest.pv < presence.pv) {
     this.presenceRequests[presence.ch][presence.id] = presence;
   }
-  this.backend.transformPresenceToLatestVersion(this, presence, function(error, presence) {
+  var context = {
+    presence: presence,
+    collection: presence.c
+  };
+  backend.trigger(backend.MIDDLEWARE_ACTIONS.receivePresence, this, context, function(error) {
     if (error) return callback(error);
-    var channel = agent._getPresenceChannel(presence.ch);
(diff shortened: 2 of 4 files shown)


docs/middleware/actions.md
```diff
@@ -99,6 +99,16 @@ This action has these additional `context` properties:
 
 > The reply about to be sent
 
+## `'receivePresence'`
```
alecgibson, line 102: I'm not 100% sure on the naming of this hook — "receive" does sound a bit weird, because it makes me think that it's just before a client is about to receive a message, but from the server's point of view it does make sense.

The other thing that's slightly odd about the naming is that all of our other hook names are named in the convention that the action is about to happen (eg `apply` means an op is about to be applied to the snapshot), whereas here the receive has already happened. Could potentially rename to `afterReceivePresence`?

alecgibson: Oh although actually this naming is consistent with the existing [`receive`](https://github.com/share/sharedb/blob/07f8bd98e5cdae37e925d4c0c709a1c3cbbca5e0/lib/agent.js#L341) hook, so maybe it's fine after all.

ericyhwang:
I have no objections to the concept of adding a `receivePresence` middleware!

That said, on a related note, do you happen to remember if there's a specific reason for sending the `sendPresence` hook sends the message down the client even if the middleware has an error?

From my perhaps now-fuzzy understanding, it goes:
- Client A sends its presence to the server
- (Agent A would trigger the new proposed `receivePresence` middleware)
- Presence update is broadcast over pub/sub
- Agent X, Y, Z get the presence update from pub/sub, and each of them trigger `sendPresence` middleware
- Agent X, Y, Z send the presence update or error(s) from middleware down to Client X, Y, Z respectively

Since the `sendPresence` middleware is more of a "is it ok to pass this other client's presence along to my client?", then the receiving client (X, Y, Z) isn't necessarily waiting on a reply, and it may be OK to not send if there's an error from middleware?

alecgibson:
@ericyhwang yes I was actually thinking about the error handling of `sendPresence` — I honestly don't think I'd thought about it too hard at the time, and I suspect I just copied the behaviour of [`reply`](https://github.com/share/sharedb/blob/07f8bd98e5cdae37e925d4c0c709a1c3cbbca5e0/lib/agent.js#L298-L302).

But now I've thought about it more, I agree with you — `reply` is different, because the client is actively awaiting a response, whereas this is just a broadcast.

In fact, for security purposes, you probably _actively_ don't want to send an error down to clients, because malicious subscribers may be subscribed just to see if activity is happening (even if they don't know exactly what it is).

I'd definitely propose not broadcasting to subscribers if the `sendPresence` middleware errors (and probably just triggering an event on the server instead), but I'd probably make that as part of another change, which I guess would technically be breaking...?

alecgibson:
@ericyhwang I've deprecated the current `sendPresence` middleware error behaviour in https://github.com/share/sharedb/pull/524

ericyhwang (approved):

PR #524, opened 2021-11-23
🔒 Prevent sending `sendPresence` middleware errors to subscribers

alecgibson:
At the moment, if an error is passed in the `sendPresence` middleware,
this error is broadcast to any subscribed listeners.

Consumers may wish to use the `sendPresence` middleware to filter
messages being sent to particular clients (eg based on authentication or
other criteria). In these cases, no information at all should be
broadcast to subscribers, since this might be something of a security
risk - a malicious client can tell there's activity on a channel, even
if they don't know exactly what that activity is.

This change deprecates the current error behaviour, and adds a new flag
to opt-in to the new behaviour of emitting errors on `Backend` instead
of sending them down to clients. If consumers haven't actively opted in
for this behaviour, they will get a deprecation warning.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -823,7 +823,9 @@ Agent.prototype._handlePresenceData = function(presence) {
   var agent = this;
   backend.trigger(backend.MIDDLEWARE_ACTIONS.sendPresence, this, context, function(error) {
     if (error) {
-      return agent.send({a: 'p', ch: presence.ch, id: presence.id, error: getReplyErrorObject(error)});
+      if (backend.doNotForwardSendPresenceErrorsToClient) backend.errorHandler(error, {agent: agent});
+      else agent.send({a: 'p', ch: presence.ch, id: presence.id, error: getReplyErrorObject(error)});
+      return;
     }
     agent.send(presence);
   });
--- a/test/backend.js
+++ b/test/backend.js
@@ -1,29 +1,51 @@
 var Backend = require('../lib/backend');
 var expect = require('chai').expect;
 var sinon = require('sinon');
+var logger = require('../lib/logger');
 
 describe('Backend', function() {
   var backend;
-  var agent = {
-    custom: {
-      foo: 'bar'
-    }
-  };
-  var fetchOptions = {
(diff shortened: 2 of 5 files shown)


lib/agent.js
```diff
@@ -815,7 +815,9 @@ Agent.prototype._handlePresenceData = function(presence) {
   var agent = this;
   backend.trigger(backend.MIDDLEWARE_ACTIONS.sendPresence, this, context, function(error) {
     if (error) {
-      return agent.send({a: 'p', ch: presence.ch, id: presence.id, error: getReplyErrorObject(error)});
+      if (backend.sendPresenceMiddlewareServerErrors) backend.emit('error', error);
```
ericyhwang, line 818: From verbal discussion today:
1) Rename flag to something like `dontForwardSendPresenceErrorToClient`
2) New non-fatal error handling mechanism - Pass in an error handler to Backend. If it's not passed in, we have a default that just logs to console.
3) Use that error handling for this case, later we can convert more of the console logs in the backend code to use the new error handler.

test/backend.js
```diff
@@ -1,29 +1,51 @@
...
-    }
-  };
-  var fetchOptions = {
-    snapshotOptions: {
-      fizz: 'buzz'
-    }
-  };
-
-  beforeEach(function() {
-    backend = new Backend();
```
ericyhwang, line 19: It feels a bit odd to remove the backend setup from `beforeEach` but keep the tear-down in `afterEach` 🤔 

I suppose it's fair to expect that every test case here will construct a Backend somehow, since this is a test for backend.js. And this way tests won't forget to close their backend.

ericyhwang (approved):
LGTM

In the release notes, be sure to mention that the `errorHandler` for now is only getting the `sendPresence` errors, but that our intent is to later move more into it.

test/backend.js
alecgibson: I think it should be fine; if tests don't set up a `backend`, then the test will throw when trying to tear it down. We can rejig the tests again in that case.

PR #525, opened 2021-11-29
Only display sendPresence error deprecation if presence is enabled

ericyhwang:
https://github.com/share/sharedb/pull/524 added a deprecation notice about presence error handling for all constructed backends without a new opt-in flag:

> Broadcasting "sendPresence" middleware errors to clients is deprecated and will be removed in a future release. Disable this behaviour with:
> new Backend({doNotForwardSendPresenceErrorsToClient: true})

Since the warning is irrelevant for backends with the default of no presence, this PR updates the notice to only display for backends that were constructed with the option `presence: true`.


curran:
Nice.

pypmannetjies (approved):
LGTM

alecgibson (approved):

PR #528, opened 2021-12-16
🐛 Prevent local presence transformation when presence created late

alecgibson:
This change fixes a presence edge case:

 1. Register a `doc.on('op')` listener which submits presence
 2. Create the `LocalPresence` _after_ registering this listener
 3. Submit an op
 4. The presence being submitted is transformed against the submitted op
    when it shouldn't be

This happens because the `LocalPresence` handles its transformation
logic in the `'op'` event. If the `LocalPresence` is created after
another handler that submits presence, then its listener is bound late,
and is only triggered after the handler that submits the op.

The result is that the `LocalPresence` will try to transform the
presence, even though the presence is already consistent with the doc's
state.

We could in theory try to prepend the `LocalPresence` listener, but we'd
still be vulnerable to the same bug if consumers _also_ prepend
listeners, and we just move the problem to another place.

Instead, this change fixes the issue by tracking the state of `doc.data`
when the presence is submitted.

However, since `doc.version` is only incremented when an op is
[acknowledged][1], we can't use this to track our document state. To
address this, this change adds a new `_dataStateVersion` internal
variable, which is incremented every time `doc.data` is changed. This
allows us to compare the state of our local `doc.data` over time and
we can use this to check if our presence needs transformation or not.

[1]: https://github.com/share/sharedb/blob/a1a33852531432baeb3cb948ca36500bdbc60fa4/lib/client/doc.js#L947-L948

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -114,6 +114,10 @@ function Doc(connection, collection, id) {
   // ops are still received. Should be toggled through the pause() and
   // resume() methods to correctly flush on resume.
   this.paused = false;
+
+  // Internal counter that gets incremented every time doc.data is updated.
+  // Used as a cheap way to check if doc.data has changed.
+  this._dataStateVersion = 0;
 }
 emitter.mixin(Doc);
 
@@ -156,11 +160,16 @@ Doc.prototype._setType = function(newType) {
   } else if (newType === null) {
     this.type = newType;
     // If we removed the type from the object, also remove its data
-    this.data = undefined;
+    this._setData(undefined);
   } else {
     var err = new ShareDBError(ERROR_CODE.ERR_DOC_TYPE_NOT_RECOGNIZED, 'Missing type ' + newType);
     return this.emit('error', err);
   }
+};
+
+Doc.prototype._setData = function(data) {
+  this.data = data;
+  this._dataStateVersion++;
 };
 
 // Ingest snapshot data. This data must include a version, snapshot and type.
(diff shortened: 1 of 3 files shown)


curran:
Wow nice catch! Very subtle.

lib/client/doc.js
```diff
@@ -648,11 +659,13 @@ Doc.prototype._otApply = function(op, source) {
...
-        this.type.createDeserialized(op.create.data) :
-        this.type.deserialize(this.type.create(op.create.data)) :
-      this.type.create(op.create.data);
+    this._setData(
+      (this.type.deserialize) ?
+        (this.type.createDeserialized) ?
+          this.type.createDeserialized(op.create.data) :
+          this.type.deserialize(this.type.create(op.create.data)) :
+        this.type.create(op.create.data)
+    );
```
dawidreedsy, line 668: Can we change this to ifs as it looks more readable 

```suggestion
let data;

if (this.type.deserialize) {
     if (this.type.createDeserialized) {
       data = this.type.createDeserialized(op.create.data)
     } else {
       data = this.type.deserialize(this.type.create(op.create.data))
     }
} else {
     data = this.type.create(op.create.data)
}
this._setData(data);
```

lib/client/presence/local-doc-presence.js
```diff
@@ -76,9 +81,14 @@ LocalDocPresence.prototype._registerWithDoc = function() {
 
 LocalDocPresence.prototype._transformAgainstOp = function(op, source) {
   var presence = this;
+  var docDataVersion = this._doc._dataStateVersion;
+
   this._pendingMessages.forEach(function(message) {
+    var messageDocDataVersion = presence._docDataVersionByPresenceVersion[message.pv];
```
ericyhwang, line 87: Would be nice to add a brief comment about what case this is trying to handle.

ericyhwang (commented):
We also discussed adding an internal "_preOp" event, and having presence hook into that. It would be easier to reason about and passes the new test cases, but we're worried about:
- Future internal code may use the "_preOp" event and break the presence transform edge cases
- If we want to extract presence into a plugin in the future (once sharedb supports plugins), then depending on an internal event is a no-go.

ericyhwang (approved):

PR #535, opened 2022-01-13
🐛 Fix issue with null timestamp in fetchSnapshotByTimestamp

dawidreedsy:
When calling
```
backend.fetchSnapshotByTimestamp('books', 'mocking-bird', null, next)
```

shareDb was loading the latest milestone snapshot and was not applying
the ops that happen after the milestone.

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -772,6 +772,16 @@ Backend.prototype._fetchSnapshot = function(collection, id, version, callback) {
   var db = this.db;
   var backend = this;
 
+  var shouldGetLatestSnapshot = version === null;
+  if (shouldGetLatestSnapshot) {
+    return backend.db.getSnapshot(collection, id, null, null, function(error, snapshot) {
+      if (error) return callback(error);
+
+      callback(null, snapshot);
+    });
+  }
+
+
   this.milestoneDb.getMilestoneSnapshot(collection, id, version, function(error, milestoneSnapshot) {
     if (error) return callback(error);
 
@@ -833,6 +843,15 @@ Backend.prototype._fetchSnapshotByTimestamp = function(collection, id, timestamp
   var from = 0;
   var to = null;
 
+  var shouldGetLatestSnapshot = timestamp === null;
+  if (shouldGetLatestSnapshot) {
+    return backend.db.getSnapshot(collection, id, null, null, function(error, snapshot) {
+      if (error) return callback(error);
+
+      callback(null, snapshot);
+    });
+  }
+
   milestoneDb.getMilestoneSnapshotAtOrBeforeTime(collection, id, timestamp, function(error, snapshot) {
     if (error) return callback(error);
     milestoneSnapshot = snapshot;
@@ -883,10 +902,6 @@ function pluckIds(snapshots) {
 }
 
 function filterOpsInPlaceBeforeTimestamp(ops, timestamp) {
-  if (timestamp === null) {
-    return;
-  }
-
   for (var i = 0; i < ops.length; i++) {
     var op = ops[i];
     var opTimestamp = op.m && op.m.ts;
--- a/test/client/snapshot-timestamp-request.js
+++ b/test/client/snapshot-timestamp-request.js
@@ -490,6 +490,54 @@ describe('SnapshotTimestampRequest', function() {
             done();
           });
       });
+
+      describe('when timestamp is null', function() {
+        it('fetches latest snapshot', function(done) {
+          sinon.spy(milestoneDb, 'getMilestoneSnapshotAtOrBeforeTime');
+          sinon.spy(milestoneDb, 'getMilestoneSnapshotAtOrAfterTime');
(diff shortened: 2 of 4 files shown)


lib/backend.js
```diff
@@ -810,6 +810,7 @@ Backend.prototype.fetchSnapshotByTimestamp = function(agent, index, id, timestam
     id: id,
     timestamp: timestamp
```
dawidreedsy, line 811: No sure though if the timestamp in the request should be null or `Date.now()`

lib/backend.js
```diff
@@ -810,6 +810,7 @@ Backend.prototype.fetchSnapshotByTimestamp = function(agent, index, id, timestam
...
   };
+  timestamp = timestamp ? timestamp : Date.now();
```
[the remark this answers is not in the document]
alecgibson: Also, while this is a nice quick fix, I do wonder if we should address the deeper issues here (ie that `null` should be able to be passed through each step of this machinery).

Namely, if I pass `null` to `getMilestoneSnapshotAtOrAfterTime()` should it actually return the live snapshot (rather than the most recent milestone?).

Alternatively we could insert some short-circuit logic here for a performance gain: maybe now is finally the time to just check if we're asking for a `null` timestamp or version, and just directly call `db.getSnapshot()`? That would potentially save us a couple of DB round-trips, which feels like a nice win, especially if consumers are just using `fetchSnapshot()` as a convenience method for fetching the latest snapshot (which we are!).

Let's discuss with @ericyhwang next week.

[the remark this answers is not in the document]
ericyhwang: > Namely, if I pass null to `getMilestoneSnapshotAtOrAfterTime()` should it actually return the live snapshot (rather than the most recent milestone?).

As discussed, the milestone snapshot adapter is separate from the normal db adapter - the Mongo version uses [a different collection](https://github.com/share/sharedb-milestone-mongo/blob/6cf429ed4a1ccb5f02a016b304d9a928a381fed5/lib/mongo-milestone-db.js#L210-L212) - so we can't return the latest actual snapshot in this case.

> Alternatively we could insert some short-circuit logic here for a performance gain: maybe now is finally the time to just check if we're asking for a null timestamp or version, and just directly call `db.getSnapshot()`?

Yeah, makes sense, for both fetchSnapshot by version and timestamp

lib/backend.js
```diff
@@ -771,6 +771,17 @@ Backend.prototype.fetchSnapshot = function(agent, index, id, version, callback)
 Backend.prototype._fetchSnapshot = function(collection, id, version, callback) {
   var db = this.db;
   var backend = this;
+  var shouldGetLatestSnapshot = version === null;
+
+  if (shouldGetLatestSnapshot) {
+    var options = {metadata: true};
```
alecgibson, line 777: Why are we fetching metadata?

lib/backend.js
```diff
@@ -833,6 +844,16 @@ Backend.prototype._fetchSnapshotByTimestamp = function(collection, id, timestamp
   var from = 0;
   var to = null;
 
+  var shouldGetLatestSnapshot = timestamp === null;
+  if (shouldGetLatestSnapshot) {
+    var options = {metadata: true};
```
alecgibson, line 849: Same here, why are we fetching metadata?

lib/backend.js
```diff
@@ -771,6 +771,17 @@ Backend.prototype.fetchSnapshot = function(agent, index, id, version, callback)
...
   var db = this.db;
   var backend = this;
+  var shouldGetLatestSnapshot = version === null;
+
+  if (shouldGetLatestSnapshot) {
+    var options = {metadata: true};
+    return backend.db.getSnapshot(collection, id, null, options, function(error, snapshot) {
+      if (error) return callback(error);
+
+      callback(error, snapshot);
```
alecgibson, line 781: We've just checked, and we don't have an error, so let's explicitly pass `null`, otherwise it looks a little confusing (hooray...callbacks...)

```suggestion
      callback(null, snapshot);
```

lib/backend.js
```diff
@@ -833,6 +844,16 @@ Backend.prototype._fetchSnapshotByTimestamp = function(collection, id, timestamp
   var from = 0;
   var to = null;
 
+  var shouldGetLatestSnapshot = timestamp === null;
+  if (shouldGetLatestSnapshot) {
+    var options = {metadata: true};
+    return backend.db.getSnapshot(collection, id, null, options, function(error, snapshot) {
+      if (error) return callback(error);
+
+      callback(error, snapshot);
```
alecgibson, line 853:
```suggestion
      callback(null, snapshot);
```

lib/backend.js
dawidreedsy: Same as above

lib/backend.js
dawidreedsy: Changed

alecgibson (approved):
Nice one

ericyhwang (approved):

Issue #501, opened 2021-07-24
Server Rendering for Queries?


curran:
Greetings,

Is it possible to execute a ShareDB Query server-side, then pass the results data into the client via server rendering and use that to data hydrate a client-side ShareDB query?

Use case: a page that lists a number of ShareDB documents, where server rendering is in play.

Desired behavior: the query runs agains the database _once_ during server rendering, then in the client side the same query can be instantiated with results from the first execution, then subscribed for updates _without_ triggering the query to be executed again.

I notice there is an internal method on Queries called `_ingestSnapshots`. This seems useful for the task at hand.

Is it possible to leverage that method? Might it be possible to expose that method as part of the public API?

curran:
I have a feeling [Racer](https://github.com/derbyjs/racer) might support this.

curran:
I notice that [createSubscribeQuery](https://share.github.io/sharedb/api/connection#createsubscribequery) has `options.results`, described as "Prior query results, if available, such as from server rendering." However it's not clear how to generate these results and what type of objects they are expected to be.

Should `results` be an array of `Doc` instances? If so, how would one go about populating them? Could one call `doc.ingestSnapshot` for each of those? For that, we'd need an array of Snapshots corresponding to query results. For that, it would indeed be useful to have a [utility method on Doc that generates a Snapshot](https://github.com/share/sharedb/issues/500). If we had that, we could iterate over `results` from a fetched query and serialize that during server rendering.

curran:
Found it in Racer thanks to @ericyhwang

https://github.com/derbyjs/racer/blob/master/lib/Model/bundle.js

https://github.com/derbyjs/racer/blob/master/lib/Model/unbundle.js

curran:
I'm still trying to figure out how to do this. The Racer logic does not appear to be directly usable.

It seems that `connection.createSubscribeQuery()` from the client side (as seen in [Leaderboard](https://github.com/share/sharedb/blob/master/examples/leaderboard/client/Leaderboard.jsx#L20)) is the most common approach, but has anyone experimented with server rendering query results? Are there any examples out there where the query is executed in the server, then the results are deserialized and instantiated client side?

I understand that `connection.createSubscribeQuery(collection, query [, options [, callback]])` can be invoked where `options.results` is defined, but how exactly to define it is not clear.

The [documentation](https://share.github.io/sharedb/api/connection#createsubscribequery) states:
 * options.results – Object[]
 * Prior query results, if available, such as from server rendering

[...]

PR #546, opened 2022-03-01
[Docs] Indicate create*Query options.results is a Doc[]

ericyhwang:
The `Connection`'s `createFetchQuery` and `createSubscribeQuery` take an optional existing result array as `options.results`.

It should be an array of Doc instances:
https://github.com/share/sharedb/blob/1cca122c63329665ebfdeceb1984921badb0cf4d/lib/client/connection.js#L563-L567

This PR updates the docs to indicate the specific type, instead of `Object[]`.

### DIFF
--- a/docs/api/connection.md
+++ b/docs/api/connection.md
@@ -70,7 +70,7 @@ Optional
 
 > `options.results` -- [`Doc`]({{ site.baseurl }}{% link api/doc.md %})[]
 
-> > Prior query results, if available, such as from server rendering
+> > Prior query results, if available, such as from server rendering. This should be an array of Doc instances, as obtained from  `connection.get(collection, id)`. If the docs' data is already available, invoke [`ingestSnapshot`]({{ site.baseurl }}{% link api/doc.md %}#ingestsnapshot) for each Doc instance beforehand to avoid re-transferring the data from the server.
 
 > `options.*` -- any
 
@@ -107,7 +107,7 @@ Optional
 
 > `options.results` -- [`Doc`]({{ site.baseurl }}{% link api/doc.md %})[]
 
-> > Prior query results, if available, such as from server rendering
+> > Prior query results, if available, such as from server rendering. This should be an array of Doc instances, as obtained from `connection.get(collection, id)`. If the docs' data is already available, invoke [`ingestSnapshot`]({{ site.baseurl }}{% link api/doc.md %}#ingestsnapshot) for each Doc instance beforehand to avoid re-transferring the data from the server.
 
 > `options.*` -- any
 
--- a/docs/api/connection.md
+++ b/docs/api/connection.md
@@ -68,7 +68,7 @@ Optional
 
 > Default: `{}`
 
-> `options.results` -- Object[]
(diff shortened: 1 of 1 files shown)


docs/api/connection.md
```diff
@@ -105,7 +105,7 @@ Optional
 
 > Default: `{}`
 
-> `options.results` -- Object[]
+> `options.results` -- [`Doc`]({{ site.baseurl }}{% link api/doc.md %})[]
```
curran, line 108: Very nice! It might help to explicitly state in the docs what is expected with regards to subscription status.

curran (approved):
LGTM!

docs/api/connection.md
alecgibson: I agree — might be nice to expand on how to actually get this `Doc` array (I have to admit I've never done this, so when I wrote the documentation, I didn't add any details)

curran: Cool! I'm (hopefully) going to be experimenting with this API soon, so I can test out what happens in the various cases and report back / update the docs when I do:
 * Doc not fetched or subscribed
 * Doc fetched but not subscribed
 * Doc fetched and subscribed

curran:
Closes  #501

(just commenting so we can easily find this PR from the issue)

docs/api/connection.md
curran: I have verified the behavior for the various cases.

Short answer: any combination of fetch and subscribe states appears to work, as long as each doc in `results` is initialized using `connection.get`.

Long answer:
 * If the docs in `results` are initialized with `connection.get` followed by `ingestSnapshot` (without subscribing), then the query implementation will subscribe to all of them (note that you still need to manually add the listeners for `doc.on('load')` and `doc.on('op')` - the subscribeQuery does not fire events when the individual docs are updated, only when the query result list is changed). In this case, we get a message that looks like this over WebSockets `a: [["id1", 10], ["id2", 3],...`. This does not contain the full documents, only their ids and versions, which is the expected behavior here. Awesome!
 * If the docs in `results` are initialized with only `connection.get` and nothing else, then a message appears overWebSockets that contains the full document contents for each individual doc.

docs/api/connection.md
```diff
@@ -68,7 +68,7 @@ Optional
...
 
 > > Prior query results, if available, such as from server rendering
```
curran, line 73:
```suggestion
> > Prior query results, if available, such as from server rendering. This should be an array of Doc instances, as obtained from `connection.get(collection, id)`. For best results, invoke `ingestSnapshot` for each Doc instance to avoid transferring data for each document.
```

docs/api/connection.md
```diff
@@ -105,7 +105,7 @@ Optional
...
```
curran, line 110:
```suggestion
> > Prior query results, if available, such as from server rendering. This should be an array of Doc instances, as obtained from `connection.get(collection, id)`. For best results, invoke `ingestSnapshot` for each Doc instance to avoid transferring data for each document. There is no need to subscribe to the documents, they will be subscribed automatically.
```

curran (changes requested):
Added suggestions that add more detail around expected state of the Doc instances.

alecgibson (approved):

PR #555, opened 2022-03-17
Fix typos in query docs

curran:
I was going through the docs and noticed that "connection" was used in the place where I would have expected "collection". Perhaps it was a typo?


ericyhwang:
Yup, looks like a typo, thanks for the fix.

We just merged a fix for the `Docs / Jekyll` build. Can you update your branch with latest upstream?

curran:
Done. Thanks!

PR #570, opened 2022-08-09
Refactor action code strings into an enum-like object

ericyhwang:
Action codes used in `request.a` are currently string literals in the source code, e.g. `qf` (query fetch) or `s` (subscribe). That makes the code harder to navigate and learn.

This PR refactors them into an enum-like object, such that they can be used like `Action.QueryFetch`.

As a bonus, it'll help if/when we find the time to switch to TypeScript, too!

Things to discuss and look at:
1. Location of the object. I created a new `shared/` directory for it, since it's code shared between client and server.
2. Naming and casing of the object and its keys. For casing, I decided to go with the UpperCamelCase used by TS-style enums, but I'm open to arguments for other casing or names.
3. Completeness and correctness. I did a search for the strings and and looked at each non-test usage one-by-one, aiming to only update the strings that were actually action codes. One example of something I didn't change was the calls to Connection#sendBulk, where the strings aren't action codes directly, since they're later prefixed with "b": https://github.com/share/sharedb/blob/e828fbb50dc6126cc873737b9cfd877334bcce12/lib/client/connection.js#L381-L383
4. Speaking of tests, I decided to not update the action code string literals in tests, so the tests can continue to serve as a guard against someone accidentally changing the action codes.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -1,8 +1,9 @@
 var hat = require('hat');
+var ShareDBError = require('./error');
+var logger = require('./logger');
+var Action = require('./shared/message-actions').Action;
 var types = require('./types');
 var util = require('./util');
-var logger = require('./logger');
-var ShareDBError = require('./error');
 
 var ERROR_CODE = ShareDBError.CODES;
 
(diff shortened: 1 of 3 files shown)


ericyhwang:
- Move from `shared/` to top level, since other shared things are at the top level.
- Switch to `ACTIONS.queryFetch` casing

alecgibson (approved):

PR #576, opened 2022-09-27
✨ Add ping/pong

alecgibson:
This change adds the ability for a client to send a "ping" to the server
and receive a "pong" response.

The motivation for this is that sometimes the socket underlying the
`Connection` may not reliably report its connection state (for example
if using a wrapper around a "vanilla" websocket that handles
reconnection).

The most bullet-proof way for a client to determine its connection state
is to actually make a request to the server and assert that it receives
a timely response.

The implementation of ping/pong is arguably a websocket concern,
especially since it's already [defined by the spec][1]. However, the
browser JavaScript [`WebSocket`][2] does not expose this functionality,
so we have to add our own ping/pong layer on top anyway.

It's also worth noting that consumers of this library can't easily
send their own ping messages down the socket, since ShareDB will
[error][3] for unknown messages.

Note that this change only adds the ability for a client to ping the
server, and not the other way around. This is because:

 - the `Agent` is not an `Emitter`, and emitting a `pong` on the
   `Backend` is pretty meaningless
 - server-side implementations of WebSockets, such as `ws`, _do_
   [expose a `ping` API][4]

[1]: https://www.rfc-editor.org/rfc/rfc6455#section-5.5.2
[2]: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket
[3]: https://github.com/share/sharedb/blob/8b531bedf19860ffe0b37a3e1c8da7a32b5005bd/lib/agent.js#L451
[4]: https://github.com/websockets/ws/blob/966f9d47cd0ff5aa9db0b2aa262f9819d3f4d414/lib/websocket.js#L351


alecgibson:
NB This would be a perfect candidate for a plugin: https://github.com/share/sharedb/issues/337

lib/client/connection.js
```diff
@@ -476,6 +478,12 @@ Connection.prototype.send = function(message) {
   this.socket.send(JSON.stringify(message));
 };
 
+Connection.prototype.ping = function() {
```
ericyhwang, line 481: From discussion today - let's throw an error if the connection state is not connected.

This is to guard against a bunch of ping requests queueing up at the socket layer if the socket is currently disconnected or connecting, and having them suddenly flood the socket when eventually connected.

ericyhwang (approved):

curran:
Very nice!

PR #579, opened 2022-10-18
🐛 Allow getting `Presence` while it's being destroyed

alecgibson:
There's currently an edge case involving trying to get presence in the middle of being destroyed. For example, consider these synchronous calls:

```js
var presence = connection.getPresence('channel')
presence.destroy()
var newPresence = connection.getPresence('channel')
```

Since `Presence` objects are per-channel singletons, the above code is effectively the same as:

```js
var presence = connection.getPresence('channel')
presence.destroy()
var newPresence = presence
```

In other words, our "new" presence is actually already being destroyed. What's worse, is that if we start trying to create local presence before the `destroy()` callback is triggered, that gets "cleaned up" by the initial call to `destroy()`:

```js
var presence = connection.getPresence('channel')
presence.destroy(() => {
  // by the time we reach here, my newly-created localPresence
  // will have been destroyed
})
var newPresence = connection.getPresence('channel')
var localPresence = newPresence.create('me')
```

This change attempts to address this edge case by adding a private `_wantsDestroy` to our `Presence` instances, which is:

  - `false` on construction
  - set to `true` when calling `destroy()`
  - reset to `false` if calling `connection.getPresence()`

We then check this flag when performing the `destroy()` tidy-up, and avoid tidying up if we have since reset the flag.

Note that there's a corner case within this edge case, surrounding `LocalPresence` instances created while a `destroy()` is in progress. Since we might create `LocalPresence` instances at any point during the asynchronous `destroy()`, we define the following behaviour:

  - any `LocalPresence` instances that exist at the time of invoking `destroy()` will be destroyed, **even if** we subsequently cancel the destroy (by invoking `connection.getPresence()`)
  - any `LocalPresence` instances that are created *after* invoking `destroy()` will be kept if the destroy is cancelled
  - creating a new `LocalPresence` while `_wantDestroy === true` will throw, because clients shouldn't put themselves in this situation and the behaviour when tidying this up is undefined

### DIFF
--- a/lib/client/presence/presence.js
+++ b/lib/client/presence/presence.js
@@ -24,6 +24,7 @@ function Presence(connection, channel) {
 
   this._remotePresenceInstances = {};
   this._subscriptionCallbacksBySeq = {};
+  this._wantsDestroy = false;
 }
 emitter.mixin(Presence);
 
@@ -36,6 +37,9 @@ Presence.prototype.unsubscribe = function(callback) {
 };
 
 Presence.prototype.create = function(id) {
+  if (this._wantsDestroy) {
+    throw new Error('Presence is being destroyed');
+  }
   id = id || hat();
   var localPresence = this._createLocalPresence(id);
   this.localPresences[id] = localPresence;
@@ -43,10 +47,15 @@ Presence.prototype.create = function(id) {
 };
 
 Presence.prototype.destroy = function(callback) {
+  this._wantsDestroy = true;
   var presence = this;
+  // Store these at the time of destruction: any LocalPresence on this
+  // instance at this time will be destroyed, but if the destroy is
+  // cancelled, any future LocalPresence objects will be kept.
+  // See: https://github.com/share/sharedb/pull/579
+  var localIds = Object.keys(presence.localPresences);
   this.unsubscribe(function(error) {
     if (error) return presence._callbackOrEmit(error, callback);
-    var localIds = Object.keys(presence.localPresences);
     var remoteIds = Object.keys(presence._remotePresenceInstances);
     async.parallel(
       [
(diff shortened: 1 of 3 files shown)


lib/client/presence/presence.js
```diff
@@ -56,13 +61,14 @@ Presence.prototype.destroy = function(callback) {
...
+          if (!presence._wantsDestroy) return next();
           async.each(remoteIds, function(presenceId, next) {
             presence._remotePresenceInstances[presenceId].destroy(next);
           }, next);
         }
       ],
       function(error) {
-        delete presence.connection._presences[presence.channel];
+        if (presence._wantsDestroy) delete presence.connection._presences[presence.channel];
         presence._callbackOrEmit(error, callback);
```
alecgibson, line 72: Note that this will still invoke the `destroy()` callback, even if the destroy is cancelled. It does this because I feel like all callbacks should eventually be invoked (eg if you tried to promisify this method, it would never resolve or reject).

However, if clients care about this, we could either:
- expose `wantsDestroy` publically so they can check this in their callback; or
- pass `wantsDestroy` into the callback as an argument

alecgibson:
Also see: https://github.com/share/sharedb/pull/580

lib/client/presence/presence.js
```diff
@@ -36,17 +37,21 @@ Presence.prototype.unsubscribe = function(callback) {
...
   id = id || hat();
   var localPresence = this._createLocalPresence(id);
   this.localPresences[id] = localPresence;
   return localPresence;
 };
 
 Presence.prototype.destroy = function(callback) {
+  this._wantsDestroy = true;
   var presence = this;
+  var localIds = Object.keys(presence.localPresences);
```
ericyhwang, line 52: For future readers, it'd be worth a comment here to explain why this is gotten synchronously whereas the `remoteIds` is gotten inside the callback.

ericyhwang (approved):

PR #581, opened 2022-11-22
✨ Allow middleware to mutate submitted ops

alecgibson:
This change adds a new `request.$fixup()` method to the `SubmitRequest` class, which allows middleware to mutate ops in the `apply` middleware.

Motivation
----------
There are times when we want the server to make changes to ops submitted by a client. There may be many reasons for wanting to do this, including distrusting a client (eg setting timestamp, or auth information); keeping sensitive business logic out of client-side code; etc.

The only ways of doing this currently are:

 1. Committing the client's op, and then submitting an amendment op from the server; or
 2. Rejecting the client's op, and submitting the full amended op from the server

Option 1 works, but possibly commits badly-formed, or untrustworthy data to the database, and broadcasts it to remote clients. It also makes it possible to fetch historic snapshots of this untrustworthy in-between state, which undermines the trustworthiness of our data.

Option 2 also works, and might work with invertible types through a "soft" rollback, but can lose pending ops if we're not careful, or if the type isn't invertible.

Approach
--------
This change adds a `$fixup()` helper function to the `SubmitRequest` object, named after the `git` action for performing similar actions.

The idea of the `$fixup()` function is to act as a middleware-friendly `submitRequest()`.

Any ops submitted through `$fixup()` will be:

 - composed with the `request.op`, so downstream middleware can work on the fixed-up op
 - sent back to the original client, and applied
 - help to avoid infinite middleware loops, since we don't trigger any nested submission calls

Note that in order to handle the case of a client disconnecting before getting their fixup, we now also store the fixup ops on the op `m` metadata field, and fetch this when getting ops to the current snapshot.

Limitations
-----------
This feature will only work:

 - on types which implement the optional `.compose()` function
 - in the `apply` middleware, where the snapshot's `type` is available, but before the op has actually been applied to the snapshot
 - on creation or edit ops (not deletion)


lib/client/doc.js
```diff
@@ -960,6 +960,21 @@ Doc.prototype._opAcknowledged = function(message) {
...
+    } catch (error) {
+      return this._hardRollback(error);
+    }
+
+    for (var i = 0; i < this.pendingOps.length; i++) {
+      var transformErr = transformX(this.pendingOps[i], message);
+      if (transformErr) return this._hardRollback(transformErr);
+    }
+
+    return;
```
alecgibson, line 975: I don't know if there's any weirdness here about applying the fixup without incrementing the version? I guess before its ack, the doc is always "ahead" of its version?

lib/submit-request.js
```diff
@@ -40,9 +41,51 @@ function SubmitRequest(backend, agent, index, id, op, options) {
   this.snapshot = null;
   this.ops = [];
   this.channels = null;
+  this._fixupOps = [];
 }
 module.exports = SubmitRequest;
 
+SubmitRequest.prototype.$fixup = function(op) {
+  if (this.action !== this.backend.MIDDLEWARE_ACTIONS.apply) {
```
alecgibson, line 49: I'm not sure if this restriction is strictly necessary — we could theoretically allow consumers to fixup in the `commit` hook (say), and just `apply()` on the snapshot, but it felt like keeping the scope of this change narrow to start was better, because I'm not 100% sure I understand the ramifications of allowing that sort of behaviour.

lib/submit-request.js
```diff
@@ -40,9 +41,51 @@ function SubmitRequest(backend, agent, index, id, op, options) {
   this.snapshot = null;
   this.ops = [];
   this.channels = null;
+  this._fixupOps = [];
 }
 module.exports = SubmitRequest;
 
+SubmitRequest.prototype.$fixup = function(op) {
```
alecgibson, line 48: I don't know if adding a function to this is the best interface? I don't know what else to suggest: consumers could just push ops onto an array, but it wouldn't perform all the checks I'm performing here, or immediately update `request.op`, etc.

As far as naming goes, I know `$` normally marks internals, but we already have non-prefixed, internal-only methods on this, so I wanted something to set this apart. It feels not crazy; frameworks like `vue` use the `$` symbol for component utility methods.

lib/submit-request.js
```diff
@@ -103,7 +146,7 @@ SubmitRequest.prototype.submit = function(callback) {
 
     // Transform the op up to the current snapshot version, then apply
     var from = op.v;
-    backend.db.getOpsToSnapshot(collection, id, from, snapshot, null, function(err, ops) {
+    backend.db.getOpsToSnapshot(collection, id, from, snapshot, {metadata: true}, function(err, ops) {
```
alecgibson, line 149: I don't know if asking for metadata here has any wider-reaching consequences? We may also need to check support of this flag to determine whether we allow fixups? (I think it'll potentially break in a weird way if it's not supported)

lib/client/doc.js
alecgibson: Fixed this by returning the fixup in the actual ack message

test/middleware.js
```diff
@@ -417,4 +420,255 @@ describe('middleware', function() {
...
+          expect(remoteDoc.data.tricks).to.eql(['fetch', 'stay']);
+          expect(doc.version).to.equal(remoteDoc.version);
+          done();
+        });
+
+        doc.submitOp([{p: ['tricks'], oi: ['fetch']}], errorHandler(done));
+      });
+    });
+
+    it('transforms pending ops by the fixup', function(done) {
```
alecgibson, line 502: We should also add a test case that locally we transform correctly. The following case requires us to reorder our apply and transform:

```js
    it('transforms pending ops by the fixup', function(done) {
      var applied = false;
      backend.use('apply', function(request, next) {
        if (applied) return next();
        applied = true;
        next(request.$fixup([{p: ['tricks', 1], li: 'stay'}]));
      });

      var remoteConnection = backend.connect();
      var remoteDoc = remoteConnection.get('dogs', 'fido');

      remoteDoc.subscribe(function(error) {
        if (error) return done(error);

        expect(remoteDoc.data).to.eql({name: 'fido'});

        remoteDoc.on('op batch', function() {
          if (remoteDoc.version !== 3) return;
          expect(remoteDoc.data.tricks).to.eql(['stay']);
          expect(remoteDoc.data).to.eql(doc.data);
          done();
        });

        doc.preventCompose = true;
        doc.submitOp([{p: ['tricks'], oi: ['fetch']}], errorHandler(done));
        doc.submitOp([{p: ['tricks', 0], ld: 'fetch'}], errorHandler(done));
      });
    });
```

alecgibson: If I add logs here:

```js
  if (message.fu) {
    console.log(this.data);
    console.log(message.fu[0].op);
    for (var i = 0; i < message.fu.length; i++) {
      var fixupOp = message.fu[i];
      try {
        this._otApply(fixupOp, false);
      } catch (error) {
        return this._hardRollback(error);
      }

      for (var j = 0; j < this.pendingOps.length; j++) {
        var transformErr = transformX(this.pendingOps[i], fixupOp);
        if (transformErr) return this._hardRollback(transformErr);
      }
    }
  }
```

This is what I get:

```js
{ name: 'fido', tricks: [] }
[ { p: [ 'tricks', 1 ], li: 'stay' } ]
```

lib/client/doc.js
```diff
@@ -960,6 +960,22 @@ Doc.prototype._opAcknowledged = function(message) {
...
+      try {
+        this._otApply(fixupOp, false);
+      } catch (error) {
+        return this._hardRollback(error);
+      }
+
+      for (var j = 0; j < this.pendingOps.length; j++) {
+        var transformErr = transformX(this.pendingOps[i], fixupOp);
+        if (transformErr) return this._hardRollback(transformErr);
+      }
```
ericyhwang, line 975: The order here of apply -> transform pending is the opposite of what happens in `_handleOp` for a remotely issued op:
https://github.com/share/sharedb/blob/c5b90c964a2e5ff10a61ab2fd4dcf3987f3147c6/lib/client/doc.js#L369-L384

We had some discussion around whether the order matters, Alec then tried to write a test case where having the "original" order from `_handleOp` might matter, but it's not quite failing with this order here due to json0 being a bit lenient?
https://github.com/share/sharedb/pull/581/files#r1035068235

Conclusion seems to be that we should swap this to match the `_handleOp` order

test/middleware.js
alecgibson: If I then update to:

```js
  if (message.fu) {
    for (var i = 0; i < message.fu.length; i++) {
      var fixupOp = message.fu[i];
      for (var j = 0; j < this.pendingOps.length; j++) {
        var transformErr = transformX(this.pendingOps[i], fixupOp);
        if (transformErr) return this._hardRollback(transformErr);
      }

      try {
        console.log(this.data);
        console.log(fixupOp.op);
        this._otApply(fixupOp, false);
      } catch (error) {
        return this._hardRollback(error);
      }
    }
  }
```

Then I get as you'd expect: insert in an empty array at index `0`:

```js
{ name: 'fido', tricks: [] }
[ { p: [ 'tricks', 0 ], li: 'stay' } ]
```

lib/submit-request.js
alecgibson: We should scrub this metadata before sending back to the client, which can happen eg if an unsubscribed doc is behind the server and submits an op.

ericyhwang: Alec to verify whether catch-up ops' metadata is sent down or not if a client that's behind in versions sends a new op

lib/submit-request.js
```diff
@@ -40,9 +41,51 @@ function SubmitRequest(backend, agent, index, id, op, options) {
...
+      ERROR_CODE.ERR_TYPE_DOES_NOT_SUPPORT_COMPOSE,
+      typeId + ' does not support compose'
+    );
+  }
+
+  try {
+    if (this.op.create) this.op.create.data = type.apply(this.op.create.data, op);
+    else this.op.op = type.compose(this.op.op, op);
+  } catch (error) {
+    return error;
```
alecgibson, line 76: Let's throw instead of return errors, because errors will be more visible and force consumers to handle them.

ericyhwang:
One thing that we discussed yesterday was whether to use the shorthand message property `fu` or the slightly longer `fixup`.

I'm fine with either since there's not a big difference in length. Though we should switch to using the new-ish file for the message action property values, so that even if we use shorthand values, the code is still more easily understandable:
https://github.com/share/sharedb/blob/master/lib/message-actions.js

alecgibson:
@ericyhwang I've gone for the longer `fixup` because:

 - it's more descriptive
 - searching for `fu` matches every `function` in the codebase 🙈 
 - it's less sweary 😅 

I've moved to using the `ACTIONS` constant, but with **one exception**: the metadata property is still explicitly keyed by the literal `fixup`, because I guess this should never really change?

ericyhwang (approved):

PR #586, opened 2023-01-18
Feature/op metadata support

maxfortun:
Reworked the PR(#513) for issue(#512).
Hoping it will make it into the main branch this time around.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -602,7 +602,7 @@ Doc.prototype._otApply = function(op, source) {
 
     // NB: If we need to add another argument to this event, we should consider
     // the fact that the 'op' event has op.src as its 3rd argument
-    this.emit('before op batch', op.op, source);
+    this.emit('before op batch', op.op, source, op.src, {op: op});
 
     // Iteratively apply multi-component remote operations and rollback ops
     // (source === false) for the default JSON0 OT type. It could use
@@ -637,7 +637,7 @@ Doc.prototype._otApply = function(op, source) {
         this._setData(this.type.apply(this.data, componentOp.op));
         this.emit('op', componentOp.op, source, op.src);
       }
-      this.emit('op batch', op.op, source);
+      this.emit('op batch', op.op, source, op.src, {op: op});
       // Pop whatever was submitted since we started applying this op
       this._popApplyStack(stackLength);
       return;
@@ -645,7 +645,7 @@ Doc.prototype._otApply = function(op, source) {
 
     // The 'before op' event enables clients to pull any necessary data out of
     // the snapshot before it gets changed
-    this.emit('before op', op.op, source, op.src);
+    this.emit('before op', op.op, source, op.src, {op: op});
     // Apply the operation to the local data, mutating it in place
     this._setData(this.type.apply(this.data, op.op));
     // Emit an 'op' event once the local data includes the changes from the
@@ -653,8 +653,8 @@ Doc.prototype._otApply = function(op, source) {
     // submission and before the server or other clients have received the op.
     // For ops from other clients, this will be after the op has been
     // committed to the database and published
-    this.emit('op', op.op, source, op.src);
-    this.emit('op batch', op.op, source);
+    this.emit('op', op.op, source, op.src, {op: op});
+    this.emit('op batch', op.op, source, op.src, {op: op});
     return;
   }
 
--- a/lib/submit-request.js
+++ b/lib/submit-request.js
@@ -171,7 +171,7 @@ SubmitRequest.prototype.commit = function(callback) {
           var op = request.op;
           op.c = request.collection;
           op.d = request.id;
-          op.m = undefined;
+          op.m = request._metadataProjection();
           // Needed for agent to detect if it can ignore sending the op back to
           // the client that submitted it in subscriptions
           if (request.collection !== request.index) op.i = request.index;
@@ -184,6 +184,39 @@ SubmitRequest.prototype.commit = function(callback) {
       });
   });
 };
+
+SubmitRequest.prototype._metadataProjection = function() {
+  var request = this;
+
+  // Default behavior
+  if (!request.opMetadataProjection) {
+    return undefined;
+  }
+
+  // Granular projection
+  if (typeof request.opMetadataProjection === 'object') {
+    return request._granularMetadataProjection();
+  }
+
+  // Full projection
+  return request.op.m;
+};
+
+SubmitRequest.prototype._granularMetadataProjection = function() {
+  var request = this;
+  var metadataProjection = {};
+  for (var key in request.opMetadataProjection) {
+    var doProject = request.opMetadataProjection[key];
+    if (! doProject ) {
+      continue;
+    }
+
+    metadataProjection[key] = request.op.m[key];
+  }
+
+  return metadataProjection;
+};
+
 
 SubmitRequest.prototype.retry = function(callback) {
   this.retries++;
--- a/test/client/submit.js
+++ b/test/client/submit.js
@@ -1382,7 +1382,7 @@ module.exports = function() {
             thisConnection.doc = docs[thisConnection.__test_id] = thisConnection.get('dogs', 'fido');
 
             thisConnection.doc.on('op', function(op, source, src, context) {
-              if (!src) { // If I am the source there is no metadata to check
+              if (!src || !context) { // If I am the source there is no metadata to check
                 return doneAfter();
               }
               var id = op[0].p[0].split(' ')[1];
--- a/test/client/submit.js
+++ b/test/client/submit.js
@@ -7,6 +7,8 @@ var numberType = require('./number-type');
 types.register(deserializedType.type);
 types.register(deserializedType.type2);
 types.register(numberType.type);
+var util = require('../util');
+var errorHandler = util.errorHandler;
 
 module.exports = function() {
   describe('client submit', function() {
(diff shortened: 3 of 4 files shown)


alecgibson:
Hi @maxfortun just a quick note to say thanks for the PR and I promise I haven't forgotten about it! It's on my todo list, just very busy right now unfortunately.

lib/submit-request.js
```diff
@@ -185,6 +185,39 @@ SubmitRequest.prototype.commit = function(callback) {
...
+SubmitRequest.prototype._granularMetadataProjection = function() {
+  var request = this;
+  var metadataProjection = {};
+  for (var key in request.opMetadataProjection) {
+    var doProject = request.opMetadataProjection[key];
+    if (! doProject ) {
+      continue;
+    }
+
+    metadataProjection[key] = request.op.m[key];
```
alecgibson, line 214: You can make this a bit more readable by inverting the logic here:

```suggestion
    if (doProject) {
      metadataProjection[key] = request.op.m[key];
    }
```

lib/submit-request.js
```diff
@@ -185,6 +185,39 @@ SubmitRequest.prototype.commit = function(callback) {
...
+  }
+
+  // Full projection
+  return request.op.m;
+};
+
+SubmitRequest.prototype._granularMetadataProjection = function() {
+  var request = this;
+  var metadataProjection = {};
+  for (var key in request.opMetadataProjection) {
```
alecgibson, line 208: We should at least add a code comment here that we only support a single level of projection.

Would be extra nice if we updated the docs to include this flag and notes on its use.

lib/backend.js
```diff
@@ -789,7 +791,7 @@ Backend.prototype._fetchSnapshot = function(collection, id, version, callback) {
     // - we want to avoid the 'op' middleware, because we later use the 'readSnapshots' middleware in _sanitizeSnapshots
     // - we handle the projection in _sanitizeSnapshots
     var from = milestoneSnapshot ? milestoneSnapshot.v : 0;
-    db.getOps(collection, id, from, version, null, function(error, ops) {
+    db.getOps(collection, id, from, version, options, function(error, ops) {
```
alecgibson, line 794: I'm not sure we've put this in the right place? Shouldn't we be fetching op metadata in [`_getSanitizedOps()`](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/backend.js#L327) (and maybe its [bulk counterpart](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/backend.js#L336))?

maxfortun: It seems that if a snapshot has some saved metadata and we do not pass `fields` and `options` to  [getSnapshot](https://github.com/share/sharedb-mongo/blob/e72c6bb1a3e9896b29a3f0cbb8f120fe5d53bd63/index.js#L335) , the metadata will not be retrieved at all. Same with `_getSanitizedOps`, if we do not pass the `options` to [getOps](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/backend.js#L369) , the database layer will not retrieve the metadata and there will be nothing to sanitize. I have a very shallow understanding of the intent here, and maybe I am missing something fundamental. I could definitely use your help to understand this better. Thank you!

lib/submit-request.js
maxfortun: Added a comment, and updated [docs](https://github.com/share/sharedb/pull/586/files#diff-e185cf468e72547cbd5b8e2dd6f5604be3330207c759de532c540c9a2702cf5d) with an example.

lib/backend.js
alecgibson: I think we should basically not touch snapshots at all in this PR. As I mentioned in my [other comment](https://github.com/share/sharedb/pull/586#discussion_r1091765989), snapshot metadata is a very different beast, and I think we should consider it separately to ops.

I *think* this belongs in `_getSanitizedOps()` with the call to [`backend.db.getOps()`](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/backend.js#L327).

lib/submit-request.js
```diff
@@ -233,6 +265,8 @@ SubmitRequest.prototype._addSnapshotMeta = function() {
     meta.ctime = this.start;
   } else if (this.op.del) {
     this.op.m.data = this.snapshot.data;
+    // break circular dependency if snapshot data already contains metadata
+    delete this.op.m.data._m;
```
alecgibson, line 269: Since we're not dealing with snapshots in this PR, please remove this.

test/client/submit.js
```diff
@@ -1210,5 +1212,189 @@ module.exports = function() {
         });
       });
     });
+
+    describe('metadata projection', function() {
+      it('passed metadata to connect', function(done) {
```
alecgibson, line 1217: I'm not sure this test has anything to do with this PR?

test/client/submit.js
```diff
@@ -1210,5 +1212,189 @@ module.exports = function() {
...
+        });
+
+        var connection = this.backend.connect(undefined, metadata);
+        connection.on('connected', function() {
+          expect(connection.agent.custom).eql(metadata);
+          done();
+        });
+      });
+
+      it('passed metadata to submit', function(done) {
```
alecgibson, line 1232: Same here: I'm not sure this test has anything to do with this PR?

test/client/submit.js
```diff
@@ -1210,5 +1212,189 @@ module.exports = function() {
...
+
+          connection.on('connected', function() {
+            var thisConnection = this;
+
+            expect(thisConnection.agent.custom).eql(thisConnection.__test_metadata);
+
+            thisConnection.doc = docs[thisConnection.__test_id] = thisConnection.get('dogs', 'fido');
+
+            thisConnection.doc.on('op', function(op, source, src, context) {
+              if (!src || !context) { // If I am the source there is no metadata to check
```
alecgibson, line 1386: Why don't we send metadata back to the original client? The server might add context the sender wants (and which remote clients will get).

What's your use-case? Does it work without the original op submitter getting their metadata?

We could add the meta to the op acknowledgement.

test/client/submit.js
```diff
@@ -1210,5 +1212,189 @@ module.exports = function() {
...
+              }
+              expect(context.op.m).equal(undefined);
+              done();
+            });
+            doc.submitOp([{p: ['tricks'], oi: ['fetch']}], {source: 'trainer'}, errorHandler(function() {}));
+          });
+        });
+      });
+
+      it('concurrent changes', function(done) {
```
alecgibson, line 1289: Can we please simplify this test case to just have two clients? I'm not sure there's a vast amount of benefit in having so many clients; it just makes the test harder to read. Would be nice to get rid of the `for` loops.

test/client/submit.js
```diff
@@ -1210,5 +1212,189 @@ module.exports = function() {
...
+
+        this.backend.use('submit', function(request, next) {
+          expect(request.agent.custom).to.have.property('username');
+          Object.assign(request.op.m, request.agent.custom);
+          request.opMetadataProjection = {username: true};
+          next();
+        });
+
+        this.backend.use('apply', function(request, next) {
+          Object.assign(request.snapshot.m, request.op.m);
```
alecgibson, line 1304: Unneeded?

test/client/submit.js
```diff
@@ -1210,5 +1212,189 @@ module.exports = function() {
         });
       });
     });
+
+    describe('metadata projection', function() {
```
alecgibson, line 1216: We've flexed metadata on ops that were broadcast over pubsub, but ops can also be fetched when submitting an op from a stale document (that isn't subscribed, and is behind the server); can we test that the ops we get in that case also have their metadata correctly set?

This would need us to tweak [`SubmitRequest.submit()`](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/submit-request.js#L106).

**Also**, can we add a test case for an unsubscribed client calling fetch after another remote client has changed the doc: this is where you'll need the `backend._getSanitizedOps()` to handle metadata projection for us.

docs/middleware/op-submission.md
```diff
@@ -61,6 +67,10 @@ backend.use('apply', (context, next) => {
   if (userId !== ownerId) {
     return next(new Error('Unauthorized'))
   }
+
+  // Add op metadata to snapshot before snapshot is stored
+  Object.assign(context.snapshot.m, context.op.m);
```
alecgibson, line 72: Okay have discussed with @ericyhwang and we think:

- we should go back to having a feature flag (sorry about the U-turn!)
- if the feature flag is enabled, we send all metadata across pubsub
- if the feature flag is enabled, we also apply the metadata projection in the `op` middleware, which should handle the cases I asked for on your tests (reiterated here for clarity)

Cases we need to handle:
- Op submission (the case you're already testing): ops broadcast over pubsub to other subscribers
- Op fetching: have a stale document that calls `doc.fetch()` — this will fetch ops through [`backend._getSanitizedOps()`](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/agent.js#L608-L610)
- Ops sent back when submitting: have a stale document that calls `doc.submitOp()` — this retrieves ops through another mechanism in [`SubmitRequest.submit()`](https://github.com/share/sharedb/blob/cc0e3382bf3df4c38c7ccc6bc52da71aff8c9f74/lib/submit-request.js#L106)

maxfortun:
Ping?

alecgibson:
@maxfortun sorry you hadn't asked for re-review, so didn't know it was ready to look at again. I'll try to look today.

lib/client/doc.js
```diff
@@ -866,6 +864,12 @@ Doc.prototype.submitOp = function(component, options, callback) {
     callback = options;
     options = null;
   }
+
+  // If agent has metadata, append on client level so that both client and backend have access to it
+  if(this.connection.agent && this.connection.agent.custom && typeof this.connection.agent.custom.metadata === "object") {
```
alecgibson, line 869: I'm not sure I understand why we're doing this? The `custom` object is meant to be a generic data blob for clients to use. Setting a magical property here is surprising and may clash with clients' existing usage.

Also `this.connection` is only ever available in setups where the client is running on the same machine as the server (as opposed to, say, a client connected remotely over websocket).

lib/submit-request.js
```diff
@@ -171,7 +171,8 @@ SubmitRequest.prototype.commit = function(callback) {
           var op = request.op;
           op.c = request.collection;
           op.d = request.id;
-          op.m = undefined;
+          op.m = request.agent.custom.metadata ? request.op.m : undefined;
```
alecgibson, line 174: As discussed above, we shouldn't be using `agent.custom`.

alecgibson (changes requested):
As discussed inline, we shouldn't be using `agent.custom`. We've also not got a feature flag as we previously requested.

I think I've lost sight of what it is this PR was trying to achieve? You wish to attach metadata from a client and broadcast it to other clients? Or you just want to broadcast server-side metadata from the server to receiving clients?

PR #587, opened 2023-01-30
✨ Add multi channel query subscription.

dawidreedsy:
At the moment we can only specify one channel the specific query, would be able to listen to, as per docs:
```javascript
backend.use('query', (context, next) => {
  // Set our query to only listen for changes on our user-specific channel
  context.channel = userChannel(context)
  next()
})
```

However let's imagine the situation where the user wants to query all the posts, the where posted by him and all his friends. Now we would need new query every for friend separately to make sure the proper scalability is preserved and we do not receive all the changes to posts collection. This change allows to listen for multiple channels, so if we want to query all user friends posts. We can do it by this:
```javascript
backend.use('query', (context, next) => {
  // Set our query to only listen for changes on our user-specific channel
  context.channels = [userChannel(context), friendChannel(context))]
  next()
})
```

Now this query would only listen to all the changes that were made to the user posts and his friends.


lib/backend.js
```diff
@@ -661,24 +661,33 @@ Backend.prototype.querySubscribe = function(agent, index, query, options, callba
...
+          backend.emit('timing', 'querySubscribe.reconnect', Date.now() - start, request);
+          callback(null, queryEmitter);
+          return;
         }
-        var ids = pluckIds(snapshots);
-        var queryEmitter = new QueryEmitter(request, stream, ids, extra);
-        backend.emit('timing', 'querySubscribe.initial', Date.now() - start, request);
-        callback(null, queryEmitter, snapshots, extra);
+        // Issue query on db to get our initial results
+        backend._query(agent, request, function(err, snapshots, extra) {
```
alecgibson, line 681: Would be nice if we could avoid hitting the DB with the same query multiple times

lib/backend.js
```diff
@@ -661,24 +661,33 @@ Backend.prototype.querySubscribe = function(agent, index, query, options, callba
...
-          stream.destroy();
-          return callback(err);
+
+    var channels = [request.channel];
+
+    if (request.channels.length) {
+      channels = request.channels;
+    }
+
+    channels.forEach(function(channel) {
```
alecgibson, line 671: What happens if we have a query listening on multiple channels, and an op that broadcasts on multiple matching channels? Will we get duplicate events being emitted?

dawidreedsy: Yeah that's a good point i would guess yes, but not sure if we can prevent that. 🤔

ericyhwang (commented):
Comments from last week, forgot to publish 😅

lib/backend.js
```diff
@@ -661,25 +661,59 @@ Backend.prototype.querySubscribe = function(agent, index, query, options, callba
...
+
+    var channels = request.channels;
+
+    if (request.channel) {
+      logger.warn('[DEPRECATED] Do no use channel for query, use channels instead.');
+      channels = [request.channel];
+    }
+
+    if (!channels || !channels.length) {
+      return callback(new Error('Required minimum one query channel.'));
```
ericyhwang: (Alec) We should use a ShareDBError here, like above

lib/backend.js
```diff
@@ -661,25 +661,59 @@ Backend.prototype.querySubscribe = function(agent, index, query, options, callba
...
+    }
+
+    var streams = [];
+
+    function cleanStreams() {
```
ericyhwang: (Craig) Why not call it `destroyStreams`

lib/backend.js
```diff
@@ -661,25 +661,59 @@ Backend.prototype.querySubscribe = function(agent, index, query, options, callba
...
+      streams.forEach(function(stream) {
+        stream.destroy();
+      });
```
ericyhwang, line 681: We can consider clearing `streams` at the end, in case this function is called multiple times. Alec suggests setting `streams.length` to 0.

lib/query-emitter.js
```diff
@@ -41,23 +43,30 @@ QueryEmitter.prototype._open = function() {
...
 };
 
 QueryEmitter.prototype.destroy = function() {
   clearTimeout(this._pollDebounceId);
   clearTimeout(this._pollIntervalId);
-  this.stream.destroy();
+
+  this.streams.forEach(function(stream) {
+    stream.destroy();
+  });
```
ericyhwang: (Alec) This may cause recursive calls to the same emitter's destroy, where this triggers another stream's destroy, which causes this emitter's stream be destroyed again.

He suggest doing something like
```js
var stream;
while (stream = streams.pop()) {
  stream.destroy();
}
```

test/client/query-subscribe.js
```diff
@@ -592,5 +592,769 @@ module.exports = function(options) {
...
+          if (err) return done(err);
+          connection.createSubscribeQuery('dogs', matchAllDbQuery, null, function(err) {
+            if (!err) return done('Should throw Required minimum one query channel error');
+            expect(err.message).to.be.equal('Required minimum one query channel.');
+            done();
+          });
+        });
+      });
+
+      it('creating a document updates a subscribed query', function(done) {
```
ericyhwang, line 657: Since many of these tests are identical to above, let's either make a shared function with the shared tests invoked for each case, or use a for-each. That reduces test maintenance.

test/client/query-subscribe.js
```diff
@@ -592,5 +592,769 @@ module.exports = function(options) {
         });
       });
     });
+
+    describe('custom channels', function() {
```
ericyhwang, line 596: (Alec) Let's add a test around making sure things still work (no duplicate doc events) if multiple channels match, if debounce is disabled by the sharedb user.

lib/backend.js
```diff
@@ -661,25 +661,59 @@ Backend.prototype.querySubscribe = function(agent, index, query, options, callba
         'DB does not support subscribe'
       ));
     }
-    backend.pubsub.subscribe(request.channel, function(err, stream) {
-      if (err) return callback(err);
+
+    var channels = request.channels;
+
+    if (request.channel) {
+      logger.warn('[DEPRECATED] Do no use channel for query, use channels instead.');
```
ericyhwang: 1. Suggested update to be clearer, since this only shows the log line - `[DEPRECATED] "query" middleware's context.channel is deprecated, use context.channels instead.`
2. Also update query middleware documentation - https://share.github.io/sharedb/middleware/actions#query

test/client/query-subscribe.js
```diff
@@ -11,586 +11,579 @@ module.exports = function(options) {
...
-      async.parallel([
-        function(cb) {
-          connection.get('dogs', 'fido').on('error', done).create({age: 3}, cb);
-        },
-        function(cb) {
-          connection.get('dogs', 'spot').on('error', done).create({age: 3}, cb);
-        }
-      ], function(err) {
+
+  it('creating an additional document updates a subscribed query', function(done) {
```
ericyhwang, line 478: We can get rid of this second one, apparently we've had this exact duplicate copy for a long time.

test/client/query-subscribe.js
```diff
@@ -11,586 +11,579 @@ module.exports = function(options) {
...
+            {pollInterval: 0, pollDebounce: 0},
+            function(err) {
+              if (err) return done(err);
+              connection.get('dogs', count.toString()).on('error', done).create({});
+            }
+          );
+          query.on('error', done);
+          query.on('insert', function() {
+            count++;
+            if (count === 3) {
```
ericyhwang, line 60: Alec - let's add a `> 3` condition that calls done with error, so it will catch cases where duplicate causes it to go above 3

ericyhwang (commented):
Alec notes that we don't have a "golden path" test case on the expected use of this new featurre.

Something like:
- Notes query on `{ users: { $in: { ['user-1', 'user2'] } } }`, with those channels
- One op on solely channel 'user-1', another op solely on 'user-2' channel
- Make sure query receives both updates

lib/backend.js
dawidreedsy: Changed

test/client/query-subscribe.js
dawidreedsy: Removed

test/client/query-subscribe.js
dawidreedsy: Changed

dawidreedsy:
> Alec notes that we don't have a "golden path" test case on the expected use of this new featurre.
> 
> Something like:
> 
> * Notes query on `{ users: { $in: { ['user-1', 'user2'] } } }`, with those channels
> * One op on solely channel 'user-1', another op solely on 'user-2' channel
> * Make sure query receives both updates

Added new test https://github.com/share/sharedb/blob/98a00be0ffba0d9dd625ddd77dc5bfc3d35c73bb/test/client/query-subscribe.js#L17-L88

dawidreedsy:
@ericyhwang @alecgibson I manage to get duplicated results with multiple channels. https://github.com/share/sharedb/actions/runs/4184518977/jobs/7250274650
In this test https://github.com/share/sharedb/blob/98a00be0ffba0d9dd625ddd77dc5bfc3d35c73bb/test/client/query-subscribe.js#L120-L147.

Although it is quite difficult to get it in test, even if the test passes we cannot be 100% sure this multi channel connection works. Maybe just test like 100 messages to ensure non of them are duplicated. 🤷 

I wonder if we should force `pollDebounce`, but i guess it still can get duplicated if there is enough messages coming through :/

alecgibson:
@dawidreedsy I'm not sure that that's a duplicate message; just looks like the inserts were reported in a different order? I'm not sure if a query's sort order is well-defined...? Or does it depend on the query? Maybe we need to fix the order?

dawidreedsy:
@alecgibson  Yeah sorry I missed the `+` `-` and tought there is `2` duplicated. Yeah will change to ignore order.

lib/error.js
```diff
@@ -68,7 +68,8 @@ ShareDBError.CODES = {
   ERR_SUBMIT_TRANSFORM_OPS_NOT_FOUND: 'ERR_SUBMIT_TRANSFORM_OPS_NOT_FOUND',
   ERR_TYPE_CANNOT_BE_PROJECTED: 'ERR_TYPE_CANNOT_BE_PROJECTED',
   ERR_TYPE_DOES_NOT_SUPPORT_PRESENCE: 'ERR_TYPE_DOES_NOT_SUPPORT_PRESENCE',
-  ERR_UNKNOWN_ERROR: 'ERR_UNKNOWN_ERROR'
+  ERR_UNKNOWN_ERROR: 'ERR_UNKNOWN_ERROR',
+  ERR_MISSING_QUERY_CHANNEL: 'ERR_MISSING_QUERY_CHANNEL'
```
alecgibson, line 72: All of our other error codes generally have the format `ERR_<THING>_<DESCRIPTION>`, so let's try to stick to that:

```suggestion
  ERR_QUERY_CHANNEL_MISSING: 'ERR_MISSING_QUERY_CHANNEL'
```

Also can you please alphabetise

test/client/query-subscribe.js
```diff
@@ -11,596 +12,828 @@ module.exports = function(options) {
...
-          expect(docs.length).eql(1);
-          expect(from).a('number');
-          expect(to).a('number');
-          expect(util.pluck(query.results, 'id')).eql(['spot', 'fido']);
-          expect(util.pluck(query.results, 'data')).eql([{age: 2}, {age: 3}]);
-          done();
-        });
-      });
+  it('returns pubSub error if fails to subscribe to channel', function(done) {
+    sinon.stub(this.backend.pubsub, 'subscribe').callsFake(function(_channel, callback) {
```
ericyhwang, line 821: Let's add an `afterEach` with `sinon.restore` in test/setup.js just to be clean about the tear-down, though Alec points out we create a new backend each test case so it's fine in this case.

test/client/query-subscribe.js
```diff
@@ -11,596 +12,828 @@ module.exports = function(options) {
...
+    });
+    query.on('error', done);
+    query.on('insert', function(docs) {
+      expect(util.pluck(docs, 'id')).eql(['fido']);
+      done();
+    });
+  });
+
+  it('pollInterval captures additional unpublished creates', function(done) {
+    this.timeout(4000);
```
ericyhwang, line 615: Let's use sinon fake timers here, to avoid artificially lengthening the test case.

Will have to advance time at least 3 times, we think?

test/client/query-subscribe.js
dawidreedsy: Added

test/client/query-subscribe.js
dawidreedsy: Added fakeTimers

alecgibson:
@dawidreedsy I've released the `sharedb-mongo` fix, but looks like Node 14 is still having issues?

curran:
It may be worth considering dropping support for Node 14, as it's nearing EOL according to https://github.com/nodejs/release#release-schedule

image

alecgibson (approved):
🚀

PR #591, opened 2023-02-21
Add missing import in hello world example

jameshfisher:


ericyhwang (approved):

ericyhwang:
Thanks! The perils of writing code examples in docs instead of as code 😅

Issue #593, opened 2023-03-02
Example using Vite


curran:
I'm working on a product that uses Vite, and was considering creating a new example for the `/examples` directory that uses Vite, just to show others how it can be done and also to get feedback on the approach.

Would this be a welcome contribution?

Is anyone else using ShareDB with Vite?

Cheers!

curran:
Related issues:
 * https://github.com/share/sharedb/issues/255
 * https://github.com/share/sharedb/issues/499

curran:
FWIW This is the approach I'm using right now in this project https://github.com/vizhub-core/vzcode :

https://github.com/vizhub-core/vzcode/tree/c5ae9e8083ed972703d98e85c632a6ad75f50a7e

This currently includes a nasty dependency on [sharedb-client-browser](https://www.npmjs.com/package/sharedb-client-browser) that I'd like to ultimately get rid of, while also avoiding addition of a node polyfill build plugin.

curran:
Closed in https://github.com/share/sharedb/pull/598

PR #598, opened 2023-03-02
Vite example

curran:
Towards #593 by cloning the `counter-json1` example and attempting to adapt it to use Vite.

Desired workflows:

 * In dev, `npm start` to start the server, then in another terminal/tab `npm run dev` to load the front end with hot reloading enabled (still need to set up the WebSocket proxy for this, but there are blockers in the way of getting to this point)
 * In prod `npm run build` followed by `npm start`, like in the other examples.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,18 +34,13 @@ var SHAREDB_RULES = {
   // as-needed quote props are easier to write
   'quote-props': ['error', 'as-needed'],
   'require-jsdoc': 'off',
-  'valid-jsdoc': 'off',
-
-  // Required after upgrade to ecmaVersion: 6
-  'no-invalid-this': 'off'
+  'valid-jsdoc': 'off'
 };
 
 module.exports = {
   extends: 'google',
   parserOptions: {
-    // Support ES6 imports and exports
-    ecmaVersion: 6,
-    sourceType: 'module'
+    ecmaVersion: 3
   },
   rules: Object.assign(
     {},
@@ -54,5 +49,17 @@ module.exports = {
   ),
   ignorePatterns: [
     '/docs/'
+  ],
+  overrides: [
+    {
+      files: ['examples/counter-json1-vite/*.js'],
+      parserOptions: {
+        ecmaVersion: 6,
+        sourceType: 'module'
+      },
+      rules: {
+        quotes: ['error', 'single']
+      }
+    }
   ]
 };
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -34,13 +34,18 @@ var SHAREDB_RULES = {
   // as-needed quote props are easier to write
   'quote-props': ['error', 'as-needed'],
   'require-jsdoc': 'off',
-  'valid-jsdoc': 'off'
+  'valid-jsdoc': 'off',
+
+  // Required after upgrade to ecmaVersion: 6
+  'no-invalid-this': 'off'
 };
 
 module.exports = {
   extends: 'google',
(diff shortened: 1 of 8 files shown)


curran:
Current status:

`npm run dev` should in theory work (aside from the WebSocket connection, which is solvable).

Currently the build results in the following error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'prototype')
    at Object.mixin (emitter.js:7:32)
    at node_modules/sharedb/lib/client/doc.js (doc.js:124:9)
    at __require (chunk-RSJERJUL.js?v=7d3727ba:3:50)
    at node_modules/sharedb/lib/client/connection.js (connection.js:1:11)
    at __require (chunk-RSJERJUL.js?v=7d3727ba:3:50)
    at node_modules/sharedb/lib/client/index.js (index.js:1:22)
    at __require (chunk-RSJERJUL.js?v=7d3727ba:3:50)
    at index.js:6:18
```

Also, a warning is emitted 

```
browser-external:events:9 Module "events" has been externalized for browser compatibility. Cannot access "events.EventEmitter" in client code. See http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility for more details.
```

This page has some interesting information http://vitejs.dev/guide/troubleshooting.html#module-externalized-for-browser-compatibility, ultimately recommending that client-side libraries expose entry points that don't rely on Node polyfills.

curran:
I'm not sure what the best solution for this would be, but _ideally_ ShareDB would provide a build for the client entrypoint that is compatible with Vite and other build systems that do not have built-in support for Node polyfills.

# Proposal

Introduce a new asset published to NPM that is a pre-built bundle of the client-side only. For consumers of ShareDB, instead of saying

```
import sharedb from 'sharedb/lib/client'
```

we could say

```
import sharedb from 'sharedb/lib/client-browser.js'
```

To produce `client-browser.js`, the prepublish step of ShareDB itself could include a step that runs 'sharedb/lib/client' through `browserify`, which would include the appropriate Node polyfills.

Thoughts?

curran:
This is odd.. Node 15 breaks in CI, but the breakage seems unrelated to the new changes:

image

curran:
CI breakage is unrelated to this change.

See also:

 * https://github.com/share/sharedb/pull/599
 * https://github.com/share/sharedb-mongo/issues/147

curran:
This is working and ready for review.

.eslintrc.js
```diff
@@ -34,13 +34,18 @@ var SHAREDB_RULES = {
...
+  'no-invalid-this': 'off'
 };
 
 module.exports = {
   extends: 'google',
   parserOptions: {
-    ecmaVersion: 3
+    // Support ES6 imports and exports
+    ecmaVersion: 6,
+    sourceType: 'module'
```
ericyhwang, line 48: Since this repo in general uses ES3 syntax at the moment, I'd recommend doing a glob override to use this custom config just for the new example, instead of for the entire repo:
https://eslint.org/docs/latest/use/configure/configuration-files#configuration-based-on-glob-patterns

ericyhwang (commented):
I haven't used Vite before, but seems straightforward enough! Just one suggestion around the ESLint config.

Re-review should happen without much delay, I was on vacation then got sick, so I'm just now getting caught up.

curran:
Thanks for the review @ericyhwang ! Feedback addressed.

ericyhwang (approved):
LGTM, thanks for the new example!

PR #606, opened 2023-04-21
💥 Drop Node.js 14 and add Node.js 20

alecgibson:
According to the Node.js [release schedule][1]:

 - v14 will be end-of-lifed on 30 April
 - v20 has been released

This change drops v14 from our test matrix, and adds v20.

[1]: https://nodejs.dev/en/about/releases/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,9 +15,9 @@ jobs:
     strategy:
       matrix:
         node:
-        - 14
         - 16
         - 18
+        - 20
     services:
       mongodb:
         image: mongo:4.4


curran:
Nice!

ericyhwang (approved):
In the past, we've major versioned sharedb when dropping old Node versions.

However, I wonder whether the major version is strictly needed? If sharedb doesn't rely on anything in the newer version of Node, then the code's still fully backwards-compatible.

I ask mainly because, even just across the Share repos, doing updates for new major versions of `sharedb` is a bit of a pain.

The counter-argument in favor of major versioning is, if we do introduce a dependency on a language or Node feature only available in Node 16+, then it'd be a breaking change in Node 14. If we're not careful, we could miss it if our tests no longer run on Node 14.

Thoughts?

alecgibson:
I think this is a major version bump: this update is a statement that we no longer test (and therefore no longer guarantee support for) older Node.js versions. I haven't got any examples off the top of my head, but I'm pretty sure major versioning for a Node.js version drop is pretty standard practice.

PR #609, opened 2023-05-09
Fixed sending multiple fetch

Qquanwei:
In some cases, multiple fetches will be sent here, resulting in a very high pressure on the server. This PR fixes this problem, but there may be some situations that have not been considered. I hope that someone can provide some information for this PR to be improved.


alecgibson:
👋🏼  Hi @Qquanwei thanks for the pull request! This is a good suggestion. I think the issue is deeper, and the real solution is in changing how `doc.fetch()` itself acts. I've put together a change that should hopefully address your issue, as well as the more general case: https://github.com/share/sharedb/pull/610

Please let me know if that doesn't fix your problem.

Qquanwei:
Move to #610

PR #610, opened 2023-05-09
⚡️ Deduplicate `fetch()` network requests

alecgibson:
At the moment, if a client calls `doc.fetch()` with a fetch request already in progress, the client will send a second request, which probably isn't needed (or wanted).

This change prevents duplicate requests from being sent when there's already an inflight fetch request, and adds an internal `force` option, which can be used to force a duplicate request to be sent. This is currently used internally for `_hardRollback()`, which forcibly unsets the `doc.version`, and therefore requires a separate `fetch()` call, since fetch requests are [handled differently][1] depending on the presence of `version`.

[1]: https://github.com/share/sharedb/blob/21750e320f6865ee14620d6bf9610cb7ebc5f18e/lib/agent.js#L607

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -425,12 +425,17 @@ Doc.prototype._resubscribe = function() {
 
 // Request the current document snapshot or ops that bring us up to date
 Doc.prototype.fetch = function(callback) {
-  if (this.connection.canSend) {
-    var isDuplicate = this.connection.sendFetch(this);
-    pushActionCallback(this.inflightFetch, isDuplicate, callback);
-    return;
-  }
+  this._fetch({}, callback);
+};
+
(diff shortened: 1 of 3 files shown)


alecgibson:
@Qquanwei why wouldn't subsequent fetches succeed? We [clear inflight](https://github.com/share/sharedb/blob/21750e320f6865ee14620d6bf9610cb7ebc5f18e/lib/client/doc.js#L297) even if there's an error, so a second fetch will send another request. Do you have a failing test case?

Qquanwei:
Thank you for your reply. I re-checked the code and found that inflightFetch will be cleared when the fetch error occurs. If this logic can work normally (for example, it can work normally without receiving any fetch reply), then this pr Should solve my problem.

ericyhwang (approved):

PR #613, opened 2023-06-12
Fix "Cannot read properties of null (reading 'callback')" error occured in Doc.prototype._handleSubscribe

jiazheng:
There is a "Cannot read properties of null (reading 'callback')" error in our logging system when using this library in our project. The error occured in method: Doc.prototype._handleSubscribe. When this method called variable `request` would be null.
Add two checkings to fix it.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -311,7 +311,7 @@ Doc.prototype._handleSubscribe = function(error, snapshot) {
   this.inflightSubscribe = null;
   var callbacks = this.pendingFetch;
   this.pendingFetch = [];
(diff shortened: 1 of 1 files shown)


alecgibson:
Can you please identify the root cause of this?

If you're on the latest version of ShareDB you might have these warnings in your logs: https://github.com/share/sharedb/pull/607

alecgibson:
I don't think this is the right fix for this; we should never reach this state.

jiazheng:
Thanks for your response. I will check the code and find out why it happened.

PR #621, opened 2023-08-29
wip

alecgibson:


alecgibson:
As far as I can tell, deleting this test doesn't affect test coverage? Although I feel like the Coveralls reporting has changed (I swear it used to say stuff like "Coverage stayed the same/dropped/etc."?)

ericyhwang (approved):
Sounds fine to me.

The scenario is pretty much covered by the above "doc fetches ops to catch up if it receives a future op" case, and I can't think of a strong reason to separately test multiple subscription ops coming in, especially considering the flakiness rate.

PR #624, opened 2023-09-19
🔊 Add log when hard rollback fetch failes

dawidreedsy:
There is a possibility that during the hard rollback the fetch operation fails, however at the moment we do not do anything with this error. We do not even log it. This change at least logs the error to notify the client that something went horribly wrong.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -329,10 +329,6 @@ Doc.prototype._handleSubscribe = function(error, snapshot) {
 Doc.prototype._handleOp = function(err, message) {
   if (err) {
     if (this.inflightOp) {
-      // The server has rejected submission of the current operation. If we get
-      // an "Op submit rejected" error, this was done intentionally
-      // and we should roll back but not return an error to the user.
-      if (err.code === ERROR_CODE.ERR_OP_SUBMIT_REJECTED) err = null;
       return this._rollback(err);
     }
     return this.emit('error', err);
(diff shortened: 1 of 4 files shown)


lib/client/doc.js
```diff
@@ -1036,12 +1036,15 @@ Doc.prototype._hardRollback = function(err) {
...
   var doc = this;
-  this._fetch({force: true}, function() {
+  this._fetch({force: true}, function(fetchError) {
     // We want to check that no errors are swallowed, so we check that:
     // - there are callbacks to call, and
     // - that every single pending op called a callback
     // If there are no ops queued, or one of them didn't handle the error,
     // then we emit the error.
+
+    fetchError && logger.error('Hard rollback doc fetch failed.', this.version, fetchError);
```
alecgibson, line 1046: I generally find this a bit more readable:

```suggestion
    if (fetchError) logger.error('Hard rollback doc fetch failed.', this.version, fetchError);
```

Also why are we logging `this.version` here and not any other information about the document? (I'd have thought an anonymous document's version on its own is pretty useless information).

I also wonder if we should `emit('error', fetchError)`?

alecgibson (commented):
Can we get a test case, please?

dawidreedsy:
As per disscussion today:
1. Make sure the ahrd rollback always have an error to ensure that we do not accidentally resolve the promises
2. Add `doc.emit('error)` for the fetch error as this is fatal.
3. Make it as a breaking change as we clinet would need to handle now the `ERR_OP_SUBMIT_REJECTED` error

lib/client/doc.js
```diff
@@ -977,48 +973,49 @@ Doc.prototype._opAcknowledged = function(message) {
...
+  // by the server, the editor window should update to reflect the undo.
+  try {
+    this._otApply(op, false);
+  } catch (otApplyError) {
+    logger.info('Cannot apply op in rollback.', op, otApplyError, 'Original error', error);
+    return this._hardRollback(otApplyError);
   }
 
-  this._hardRollback(err);
+  this._clearInflightOp(error);
```
alecgibson, line 1015: We can probably keep the old `OP_SUBMIT_REJECTED` logic here, since at this point in the code we've successfully rolled back the inflight op with no further errors, so we can swallow the error as originally intended.

```suggestion
  // The server has rejected submission of the current operation. If we get
  // an "Op submit rejected" error, this was done intentionally
  // and we should roll back but not return an error to the user.
  if (error.code !== ERROR_CODE.ERR_OP_SUBMIT_REJECTED) {
    this._clearInflightOp(error);
  }
```

lib/client/doc.js
```diff
@@ -977,48 +973,49 @@ Doc.prototype._opAcknowledged = function(message) {
...
 };
 
-Doc.prototype._hardRollback = function(err) {
+Doc.prototype._hardRollback = function(error) {
```
alecgibson, line 1018: I think we should always make sure this is set: I don't think it makes sense to ever resolve one of the pending op callbacks from `_hardRollback()` (since they were never actually submitted to the server).

```suggestion
Doc.prototype._hardRollback = function(error) {
  if (!error) error = new ShareDBError(...);
```

test/client/doc.js
```diff
@@ -514,6 +514,42 @@ describe('Doc', function() {
...
+            done();
+          });
+
+          // The server error should get all the way back to our handler
+          doc.submitOp(nonInvertibleOp, function() {
+            next();
+          });
+        }
+      ]);
+    });
```
alecgibson, line 552: Tweaked the test to:

- simplify setup
- add an error handler to `async.series()`
- make clearer assertions on which error is expected where, and wait for both errors

```js
it('throws an error when hard rollback fetch failed', function(done) {
  var backend = this.backend;
  doc = this.connection.get('dogs', 'scrappy');
  types.register(richText);

  async.series([
    doc.create.bind(doc, {ops: [{insert: 'Scrappy'}]}, 'rich-text'),
    function(next) {
      backend.use('reply', function(replyContext, cb) {
        if (replyContext.request.a !== 'f') return cb();
        cb({code: 'FETCH_ERROR'});
      });
      backend.use('submit', function(_context, cb) {
        cb(new ShareDBError('SUBMIT_ERROR'));
      });
      var nonInvertibleOp = [{insert: 'e'}];

      var count = 0;
      function expectError(code) {
        count++;
        return function(error) {
          expect(error.code).to.equal(code);
          count--;
          if (!count) next();
        };
      }

      doc.on('error', expectError('FETCH_ERROR'));
      doc.submitOp(nonInvertibleOp, expectError('SUBMIT_ERROR'));
    }
  ], done);
});
```

test/client/presence/doc-presence.js
```diff
@@ -705,7 +705,13 @@ describe('DocPresence', function() {
           });
         });
 
-        doc1.submitOp({index: 5, value: 'ern'}, errorHandler(done));
+        doc1.submitOp({index: 5, value: 'ern'}, errorHandler(function(error) {
+          if (!error) {
+            return done('should throw an error');
+          }
```
alecgibson, line 711: Is this here because `mocha` swallows assertion errors inside callbacks?

If that's the case, can we write a little helper to tidy this pattern up? Something like:

```js
function asyncAssert(done, cb) {
  return function() {
    try {
      cb.apply(this, arguments);
    } catch (error) {
      done(error);
    }
  }
}

doc1.submitOp({...}, asyncAssert(done, function(error) {
  expect(error.code).to.be.equal('ERR_OP_SUBMIT_REJECTED');
});
```

ericyhwang: I believe Mocha does handle assertion errors inside callbacks properly, so this separate `if (!error)` check may not be needed. Try removing and see if it's ok?

dawidreedsy:
Fetch error changes moved to : 
- https://github.com/share/sharedb/pull/626

PR #626, opened 2023-10-03
♻  Throw error for pending ops on hard rollback

dawidreedsy:
Our current flow for handling the rollback swallows some errors and discards pending ops.
1. Whenever we do hard rollback and fetch for hard rollback fails we should throw the error so that user is informed that something really wrong happen.
2. Whenever we preform the hard rollback and we have some pending ops and the `ERR_OP_PENDING_OP_SUBMIT_REJECTED` error is thrown we should soft reject, as this error is meant for the user to reject some of the ops from submition to DB, however we should throw for all the pending ops s the pending ops would be otherwise just discarded. This way we at least allow user to handle the error
3. For the soft rollback we just resolve for this error and allow pending ops to be submitted to the server.

This PR is extracted from:
- https://github.com/share/sharedb/pull/624


ericyhwang:
From discussion between @dawidreedsy, @alecgibson, Edmond, and me:

For soft-rejecting an op:
- The main op's callback is resolved, both in the soft rollback and hard rollback cases
- If it was a soft rollback, then originally pending ops are re-applied onto the rolled-back version
- If it was a hard rollback, then originally pending ops get rejected with a special hard-rollback error

That behavior should apply to edit, create, and del.

test/client/submit.js
```diff
@@ -949,6 +970,29 @@ module.exports = function() {
...
+          if (request.op.create) {
+            next(request.rejectedError());
+          }
+        });
+
+        var connection = this.backend.connect();
+        var doc = connection.get('dogs', 'fido');
+        doc.preventCompose = true;
+
+        doc.create({age: 3});
```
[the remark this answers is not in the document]
dawidreedsy: Cannot really as we check create without callback

lib/client/doc.js
```diff
@@ -1035,17 +1037,46 @@ Doc.prototype._hardRollback = function(err) {
...
     // - there are callbacks to call, and
     // - that every single pending op called a callback
     // If there are no ops queued, or one of them didn't handle the error,
     // then we emit the error.
+
+    if (fetchError) {
+      // This is critical error as it means that our doc is not in usable state
+      // anymore, we should throw doc error.
+      logger.error('Hard rollback doc fetch failed.', fetchError, inflightOp);
+      doc.emit('error', fetchError);
```
ericyhwang: Let's create a new error and concat the `fetchError` message on, e.g. "Hard rollback doc fetch error: _____"

lib/client/doc.js
```diff
@@ -1035,17 +1037,46 @@ Doc.prototype._hardRollback = function(err) {
...
+      }
+
+      if (!pendingOps.length) return;
+      err = new ShareDBError(
+        ERROR_CODE.ERR_PENDING_OP_REMOVED_BY_OP_SUBMIT_REJECTED,
+        'Discarding pending op because of hard rollback during ERR_OP_SUBMIT_REJECTED'
+      );
+    }
+
+    if (inflightOp) pendingOps.push(inflightOp);
```
ericyhwang: (Alec) Switch to `unshift` to preserve order compared to today

alecgibson (approved):

Issue #646, opened 2024-03-27
Messages artificially delayed when running `Backend` in the browser


alecgibson:
We run `Backend` in the browser for some of our frontend tests, which lets us test against a full ShareDB stack.

This works fine, but I recently discovered that ShareDB slows our tests down non-negligibly, just because of the [`setTimeout()`](https://github.com/share/sharedb/blob/91cae28a0065faf68749243ef36aadf54bf0645d/lib/stream-socket.js#L58) in `StreamSocket._write()`.

Digging into this, I found that we're hitting an artificial minimum 4ms delay as described in the [HTML standard](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers):

> Timers can be nested; after five such nested timers, however, the interval is forced to be at least four milliseconds.

These message delays add up (even just waiting for a `Connection` handshake will take at least 8ms of idle time), and over the course of hundreds of browser-based tests, can cause unneeded slowdown on the order of minutes.

I'm not sure what I think the solution should be. In our tests, we're monkey-patching the method to wait for a `Promise` instead of a `setTimeout()`, which isn't subject to the same minimum delay. However, that's probably not a good thing to do in Production, because waiting for a microtask is not the same as waiting for a macrotask (and also Node.js doesn't appear to be subject to this minimum time).

I tried playing with `MessageChannel` very quickly, but that doesn't seem to work in the way I was expecting.

alecgibson:
We got a speed-up of ~40% just by removing the 4ms delay from `sharedb`'s `nextTick()`:

![Screenshot 2024-03-28 at 13 21 32](https://github.com/share/sharedb/assets/12036746/2029d32a-654f-4c9b-9688-b3e01933e66e)

PR #647, opened 2024-03-28
⚡️ Remove artificial 4ms `nextTick()` delay when running in a browser

alecgibson:
Fixes https://github.com/share/sharedb/issues/646

ShareDB makes heavy use of [`nextTick()`][1]. It notably calls it every time we [send a message over a `StreamSocket`][2].

If ShareDB is running both `Backend` and `Client` in a browser (eg in client tests), `nextTick()` will [fall back to `setTimeout()`][3].

However, according to the [HTML standard][4]:

> Timers can be nested; after five such nested timers, however, the
> interval is forced to be at least four milliseconds.

So using `setTimeout()` can incur a penalty of 4ms in the browser, just idling.

Over the course of a test suite, which makes a lot of fast ShareDB requests in series, these delays can add up to many seconds or even minutes of idle CPU time.

This change adds an alternative `nextTick()` implementation, using `MessageChannel`, which is present in both [Node.js][5] and [HTML][6] (with slightly different APIs, but close enough for our purposes). This class offers a way of waiting for a tick on the event loop, without the arbitrary 4ms delay.

`MessageChannel` is [supported back to even IE10][7], and since  Node.js v10.5.0, but if for some reason it's missing, we'll still fall back to `setTimeout()`.

[1]: https://github.com/share/sharedb/blob/5259d0e0b66c50ff9e745fe34a55c5fb16c57a8e/lib/util.js#L88
[2]: https://github.com/share/sharedb/blob/5259d0e0b66c50ff9e745fe34a55c5fb16c57a8e/lib/stream-socket.js#L58
[3]: https://github.com/share/sharedb/blob/5259d0e0b66c50ff9e745fe34a55c5fb16c57a8e/lib/util.js#L98
[4]: https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timers
[5]: https://nodejs.org/api/worker_threads.html#class-messagechannel
[6]: https://html.spec.whatwg.org/multipage/web-messaging.html#message-channels
[7]: https://caniuse.com/mdn-api_messagechannel_messagechannel

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -95,9 +95,20 @@ exports.nextTick = function(callback) {
     args[i - 1] = arguments[i];
   }
 
-  setTimeout(function() {
+  if (typeof MessageChannel === 'undefined') {
+    return setTimeout(triggerCallback);
+  }
+
+  var channel = new MessageChannel();
+  channel.port1.onmessage = function() {
+    triggerCallback();
+    channel.port1.close();
+  };
+  channel.port2.postMessage('');
+
+  function triggerCallback() {
     callback.apply(null, args);
-  });
+  }
 };
 
 exports.clone = function(obj) {
--- a/.mocharc.yml
+++ b/.mocharc.yml
@@ -2,3 +2,5 @@ reporter: spec
 check-leaks: true
 recursive: true
 file: test/setup.js
+globals:
+- MessageChannel # Set/unset to test nextTick()
--- a/test/util-test.js
+++ b/test/util-test.js
@@ -45,6 +45,39 @@ describe('util', function() {
         });
         expect(called).to.be.false;
       });
+
+      describe('without MessageChannel', function() {
+        var _MessageChannel;
+
+        before(function() {
+          _MessageChannel = global.MessageChannel;
+          delete global.MessageChannel;
+        });
+
+        after(function() {
+          global.MessageChannel = _MessageChannel;
+        });
+
+        it('uses a different ponyfill', function(done) {
+          expect(process.nextTick).to.be.undefined;
+
+          util.nextTick(function(arg1, arg2, arg3) {
+            expect(arg1).to.equal('foo');
+            expect(arg2).to.equal(123);
+            expect(arg3).to.be.undefined;
+            done();
+          }, 'foo', 123);
+        });
+
+        it('calls asynchronously', function(done) {
+          var called = false;
+          util.nextTick(function() {
+            called = true;
+            done();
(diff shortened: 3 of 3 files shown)


lib/util.js
alecgibson, line 1: To discuss: we *can* do this, but *should* we do it?

The 4ms is probably in the standard for some reason, but I don't really know what. Presumably the original goal was to prevent CPU starvation, but I don't know why you would decree an arbitrary 4ms, and not just say that the time can't be guaranteed (and then it should be up to the vendor to defer event loop tasks if it needs to do other browsery things like repaint, etc.).

I note that Safari isn't standards-compliant (shocker), and doesn't actually seem to enforce this minimum time, so clearly *someone* thinks it's not too important to wait for this time.

I acknowledge that my proposed use-case is for tests, and maybe shouldn't impact Production-grade code.

One potential option is to add a flag that can select a `nextTick()` method? Or we just close this and I'll stick to my monkeypatch:

```ts
// Patch nextTick with a version that doesn't use setTimeout() so that our
// messages don't suffer artificial lag
shareDbUtils.nextTick = async function(callback: Function, ...args: any[]): Promise<void> {
  await macrotask();
  callback(...args);
};

function macrotask(): Promise<void> {
  const channel = new MessageChannel();
  return new Promise((resolve) => {
    channel.port1.onmessage = () => {
      resolve();
      channel.port1.close();
    };
    channel.port2.postMessage('');
  });
}
```

alecgibson, line 1: I also just realised that this PR won't actually help our situation 🤦🏼‍♂️ We have to include `process` anyway (I think other Node.js polyfills might rely on it, like `stream`).

The only way this would help us is if we moved the `MessageChannel` approach to be the _first_ approach (above `process.nextTick()`), but I can understand that might not be palatable?

alecgibson: Although I've dug a bit and it seems like the `process` thing is [a bug](https://github.com/browserify/node-util/issues/43), and we can avoid needing `util` completely. Have separately raised https://github.com/share/sharedb/pull/648

ericyhwang (approved):
From discussion:
- There are a few places in browser code that call `util.nextTick`, but in normal usage none of them should nest
- Existing bundled browser code should be unaffected since we currently require a polyfill of `util`, which pulls in a `process` polyfill.
- Might be worth separately updating `util.nextTick` to check on module init and swap implementations, instead of checking each call, though JS engines may mostly optimize that difference away

PR #648, opened 2024-03-28
♻️ Try loading `'inherits'` library if `'util'` is missing

alecgibson:
When running ShareDB in a browser environment, we have to polyfill the `util` library.

However, we only use the `inherits` part of that library.

This change will attempt to fall back to the browser-friendly [`inherits` library][1] if `util` can't be found, allowing for lighter polyfilling.

[1]: https://github.com/isaacs/inherits

### DIFF
--- a/lib/op-stream.js
+++ b/lib/op-stream.js
@@ -1,4 +1,3 @@
-var inherits = require('util').inherits;
 var Readable = require('stream').Readable;
 var util = require('./util');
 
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "async": "^3.2.4",
     "fast-deep-equal": "^3.1.3",
     "hat": "0.0.3",
+    "inherits": "^2.0.4",
```
alecgibson, line 11: If we don't want to add an extra dependency, we could potentially add something like `inherits` itself does?

```js
try {
  var util = require('util');
  if (typeof util.inherits !== 'function') throw new Error();
  module.exports = util.inherits;
} catch (e) {
  module.exports = require('inherits');
}
```

alecgibson: I actually think this is better. Going to change to this.

ericyhwang (approved):

PR #649, opened 2024-04-02
♻️ Check `nextTick` polyfill once at module level

alecgibson:
At the moment we check for the presence of `process` every time we call `nextTick()`.

This change checks just once at the module level instead.


lib/util.js
```diff
@@ -85,31 +85,36 @@ exports.truthy = function(arg) {
...
-  if (typeof process !== 'undefined' && process.nextTick) {
-    return process.nextTick.apply(null, arguments);
-  }
-
-  var args = [];
-  for (var i = 1; i < arguments.length; i++) {
-    args[i - 1] = arguments[i];
-  }
+if (typeof process !== 'undefined' && typeof process.nextTick === 'function') {
+  exports.nextTick = process.nextTick.apply(null, arguments);
```
ericyhwang, line 89: I think this can just be:
```
  exports.nextTick = process.nextTick;
```

`nextTick.apply` immediately invokes nextTick at module load time, which isn't quite what we want. Plus `arguments` won't be defined at module level.

lib/util.js
```diff
@@ -85,31 +86,13 @@ exports.truthy = function(arg) {
...
-  };
-  channel.port2.postMessage('');
-
-  function triggerCallback() {
-    callback.apply(null, args);
-  }
-};
+if (typeof process !== 'undefined' && typeof process.nextTick === 'function') {
+  exports.nextTick = process.nextTick;
+} else if (typeof MessageChannel === 'undefined') {
```
ericyhwang, line 91:
```suggestion
} else if (typeof MessageChannel !== 'undefined') {
```

PR #651, opened 2024-05-14
🐛 Avoid mutating milestone snapshots when building from ops

alecgibson:
At the moment, we directly pass our milestone snapshot to `ot.applyOps()`, which [mutates the snapshot][1].

In most Production-grade implementations of the `MilestoneDb`, this is fine, since mutating the in-memory `Snapshot` will have no effect on the underlying database representation.

However, this does break the `MemoryMilestoneDb` (eg in tests).

In order to fix this, and generally be defensive (eg if db adapters decide to implement an in-memory cache), this change clones the snapshot before calling `ot.applyOps()` on it.

[1]: https://github.com/share/sharedb/blob/a5c19b986d65c93abbc9ed11f7a25c0dbaf0788c/lib/ot.js#L172

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -909,7 +909,7 @@ Backend.prototype._fetchSnapshotByTimestamp = function(collection, id, timestamp
 };
 
 Backend.prototype._buildSnapshotFromOps = function(id, startingSnapshot, ops, callback) {
-  var snapshot = startingSnapshot || new Snapshot(id, 0, null, undefined, null);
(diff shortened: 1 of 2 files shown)


ericyhwang (approved):
Looks reasonable. `_buildSnapshotFromOps` is only called for milestone retrieval, so overall extra performance overhead should be negligible.

Just had one suggestion for an extra comment to clarify the new test case, go ahead and merge afterwards.

PR #652, opened 2024-05-14
✅ Allow deleting ops in `MemoryDb`

alecgibson:
It can be useful to delete ops from `MemoryDb` to test how a system will act with TTLed ops.

This change:

 - adds a `MemoryDb.deleteOps()` helper method
 - adapts `MemoryDb.getOps()` to filter out deleted ops, and check that the correct number of ops are being returned, and error if not (similar to [`sharedb-mongo`][1])

[1]: https://github.com/share/sharedb-mongo/blob/0013d15df717ac8af32aa7847ed64bca96be7f66/index.js#L550

### DIFF
--- a/lib/db/memory.js
+++ b/lib/db/memory.js
@@ -84,16 +84,30 @@ MemoryDB.prototype.getOps = function(collection, id, from, to, options, callback
   if (typeof callback !== 'function') throw new Error('Callback required');
   util.nextTick(function() {
     var opLog = db._getOpLogSync(collection, id);
-    if (to == null) {
-      to = opLog.length;
+    if (!from) from = 0;
+    if (to == null) to = opLog.length;
+    var ops = clone(opLog.slice(from, to).filter(Boolean));
+    if (ops.length < to - from) {
+      return callback(new Error('Missing ops'));
     }
-    var ops = clone(opLog.slice(from, to));
     if (!includeMetadata) {
       for (var i = 0; i < ops.length; i++) {
         delete ops[i].m;
       }
     }
     callback(null, ops);
+  });
+};
+
+MemoryDB.prototype.deleteOps = function(collection, id, from, to, options, callback) {
+  if (typeof callback !== 'function') throw new Error('Callback required');
+  var db = this;
+  util.nextTick(function() {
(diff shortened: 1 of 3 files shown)


test/client/snapshot-version-request.js
```diff
@@ -144,7 +144,7 @@ describe('SnapshotVersionRequest', function() {
 
     it('errors if asking for a version that does not exist', function(done) {
       backend.connect().fetchSnapshot('books', 'don-quixote', 4, function(error, snapshot) {
-        expect(error.code).to.equal('ERR_OP_VERSION_NEWER_THAN_CURRENT_SNAPSHOT');
+        expect(error).to.be.ok;
```
alecgibson, line 147: I had to weaken this assertion, but note that if this was running against `sharedb-mongo` (for example), we also wouldn't get the expected error code, since `sharedb-mongo` would [error in `getOps()`](https://github.com/share/sharedb-mongo/blob/0013d15df717ac8af32aa7847ed64bca96be7f66/index.js#L550), which would [trigger an early return](https://github.com/share/sharedb/blob/a5c19b986d65c93abbc9ed11f7a25c0dbaf0788c/lib/backend.js#L830) before hitting the [asserted error](https://github.com/share/sharedb/blob/a5c19b986d65c93abbc9ed11f7a25c0dbaf0788c/lib/backend.js#L836)

alecgibson: Okay coverage fixed

lib/db/memory.js
```diff
@@ -84,10 +84,12 @@ MemoryDB.prototype.getOps = function(collection, id, from, to, options, callback
   if (typeof callback !== 'function') throw new Error('Callback required');
   util.nextTick(function() {
     var opLog = db._getOpLogSync(collection, id);
-    if (to == null) {
-      to = opLog.length;
+    if (!from) from = 0;
```
ericyhwang, line 87: ℹ️ This defaulting matches the behavior of sharedb-mongo
https://github.com/share/sharedb-mongo/blob/0013d15df717ac8af32aa7847ed64bca96be7f66/index.js#L632-L633

ericyhwang (approved):

PR #659, opened 2024-05-22
🗃 Add `_create` metadata to snapshots to avoid `getCommittedOpVersion()`

alecgibson:
Supersedes https://github.com/share/sharedb/pull/657

The `DB.getCommitedOpVersion()` function is only used [in one place][1]: in a corner case that differentiates between two "blind", versionless (created without first fetching) create op conflict cases:

 - a versionless create op that conflicts with another create op
 - a versionless create op that has been re-submitted because the connection was closed before the ack was received

The first of these cases should result in an error, and the second is non-fatal and the error can be swallowed.

At the moment, this differentiation is made using the special `DB.getCommittedOpVersion()` function, which - given an op with a `src` and `seq` combination - will return `null` if the op hasn't been committed, or its version number if it has.

If the op has a committed version number, we know that the submit is a duplicate, and we can safely ignore it.

This approach is problematic, because:

 - it [needs a whole op index][2] just to handle this niche corner case
 - the default behaviour [fetches **all** ops][3] unless the driver overrides this behaviour

This change proposes that we actually don't need this function in most cases, and implements an alternative approach to differentiate between the above cases.

When creating an snapshot, we store the `src`, `seq` and `v` of the op that was used to create it.

If a conflicting `create` is received, we can then compare directly with this metadata, without needing to fetch anything extra from the database.

This should work in the majority of cases, but won't work if the metadata is missing, which could happen if:

 - the snapshot is "old": it was created before this change
 - the driver doesn't support metadata (eg [Postgres][4])

In the case where metadata is unavailable, we fall back to the existing method, using `getCommittedOpVersion()`. However, if drivers support metadata, this should happen sufficiently infrequently that consumers could potentially remove the `src`/`seq`/`v` index with no noticeable performance penalty.

[1]: https://github.com/share/sharedb/blob/7b20313ded4c302b9416cc6c7821694a7fa491b8/lib/submit-request.js#L112
[2]: https://github.com/share/sharedb-mongo/issues/94
[3]: https://github.com/share/sharedb/blob/7b20313ded4c302b9416cc6c7821694a7fa491b8/lib/db/index.js#L69
[4]: https://github.com/share/sharedb-postgres/blob/499702acd478645bcc249fa50ba6fc066d257d04/index.js#L140

### DIFF
--- a/lib/submit-request.js
+++ b/lib/submit-request.js
@@ -109,7 +109,7 @@ SubmitRequest.prototype.submit = function(callback) {
         // case, we should return a non-fatal 'Op already submitted' error. We
         // must get the past ops and check their src and seq values to
         // differentiate.
-        backend.db.getCommittedOpVersion(collection, id, snapshot, op, null, function(err, version) {
+        request._fetchCreateOpVersion(function(error, version) {
           if (err) return callback(err);
           if (version == null) {
             callback(request.alreadyCreatedError());
@@ -194,6 +194,15 @@ SubmitRequest.prototype.commit = function(callback) {
   backend.trigger(backend.MIDDLEWARE_ACTIONS.commit, this.agent, this, function(err) {
     if (err) return callback(err);
     if (request._fixupOps.length) request.op.m.fixup = request._fixupOps;
+    if (request.op.create) {
+      // When we create the snapshot, we store a pointer to the op that created
+      // it. This allows us to return OP_ALREADY_SUBMITTED errors when appropriate.
+      request.snapshot.m._create = {
+        src: request.op.src,
+        seq: request.op.seq,
+        v: request.op.v
+      };
+    }
 
     // Try committing the operation and snapshot to the database atomically
     backend.db.commit(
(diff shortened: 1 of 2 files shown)


lib/submit-request.js
```diff
@@ -290,6 +299,20 @@ SubmitRequest.prototype._shouldSaveMilestoneSnapshot = function(snapshot) {
   return this.saveMilestoneSnapshot;
 };
 
+SubmitRequest.prototype._fetchCreateOpVersion = function(callback) {
+  var create = this.snapshot.m._create;
```
alecgibson, line 303: Note that this approach has one pitfall: it opens up a *new* corner case where
an op may be incorrectly rejected instead of acked:

 1. Client A submits Create 1
 2. Create 1 committed
 3. Client A disconnects before receiving the ack
 4. Client B submits and commits Delete, then Create 2
 5. Client A reconnects, and resubmits Create 1
 6. Create 1 doesn't match the current snapshot metadata for Create 2, so it's rejected with `DOC_ALREADY_CREATED` instead of the non-fatal `OP_ALREADY_SUBMITTED`

I don't know how much we care about this? It definitely is a bug that didn't used to exist, but I'm not sure it's important enough to keep leaning on `getCommittedOpVersion()` all the time? 

In my personal opinion, if two clients are creating/deleting/recreating the same `Doc` in such a close span of time, errors are bound to happen anyway, since creates/deletes don't have great transformation behaviour. And if this is an expected use case, consumers should already have robust error-handling in place for this sort of conflict.

test/client/submit.js
```diff
@@ -206,93 +207,122 @@ module.exports = function() {
...
+      describe('no snapshot metadata available', function() {
+        beforeEach(function() {
+          var getSnapshot = MemoryDB.prototype.getSnapshot;
+          sinon.stub(MemoryDB.prototype, 'getSnapshot')
+            .callsFake(function(collection, id, fields, snapshotOptions, callback) {
+              // The $submit field requests metadata, so setting to false will
+              // imitate a driver that doesn't support metadata; or a snapshot that's
+              // missing it
+              if (fields) fields.$submit = false;
+              getSnapshot.call(this, collection, id, fields, snapshotOptions, callback);
```
ericyhwang, line 225: I'm a bit worried `$submit` has or in the future will have other side effects, so it'd be safer to manually strip the metadata from the resulting snapshot

lib/submit-request.js
ericyhwang: Per discussion, this should be a vanishingly small edge case in practice, and Client A reconnecting may run into other errors while reconnecting anyways that would require it to re-initialize.

ericyhwang (approved):

PR #660, opened 2024-05-29
🐛 Fix error handling in `_fetchCreateOpVersion()`

alecgibson:
We're currently not returning errors from the database adapter if there are issues fetching the committed op, because of a code typo.

This change adds coverage for this case and fixes the bug.


alecgibson:
Flaky build should be fixed in https://github.com/share/sharedb/pull/662

PR #662, opened 2024-05-29
✅ Fix create resubmit test

alecgibson:
One of our tests currently checks resubmitting a create op after reconnecting during submission.

However, since reconnection happens so quickly, this test case can cause two commit attempts in parallel.

We currently only wait for the first request to resolve, which means that the second request can continue into the next test, causing unexpected test failures.

This change waits for both acks to be sent before finishing the test.

### DIFF
--- a/test/client/submit.js
+++ b/test/client/submit.js
@@ -289,11 +289,15 @@ module.exports = function() {
             next();
           });
 
-          var doc = connection.get('dogs', 'fido');
-          doc.create({age: 3}, function(error) {
-            expect(doc.version).to.equal(1);
-            done(error);
+          var count = 0;
+          backend.use('reply', function(message, next) {
+            next();
+            if (message.reply.a === 'op') count++;
+            if (count === 2) done();
           });
+
+          var doc = connection.get('dogs', 'fido');
+          doc.create({age: 10}, errorHandler(done));
         });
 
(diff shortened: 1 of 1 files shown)


alecgibson:
This is an example error of the failing test:

```
  1) db
       client submit
         create
           metadata enabled
             does not fail when resubmitting a create op on a doc that was deleted:
     Error: Op submit failed. DB missing ops needed to transform it up to the current snapshot version
      at wrapErrorData (/home/runner/work/sharedb/sharedb/lib/client/connection.js:268:13)
      at Connection.handleMessage (/home/runner/work/sharedb/sharedb/lib/client/connection.js:194:11)
      at socket.onmessage (/home/runner/work/sharedb/sharedb/lib/client/connection.js:145:18)
      at /home/runner/work/sharedb/sharedb/lib/stream-socket.js:59:12
      at process.processTicksAndRejections (node:internal/process/task_queues:77:11)
```

ericyhwang (approved):
Per discussion, we should separately consider doing a `db.close()` in sharedb-mongo afterEach for these test cases imported from core, as that may help with these kinds of issues in general.

https://github.com/share/sharedb-mongo/blob/d48f2c7858a8a3353b1ac0628d17109b41bfcd1d/test/test_mongo.js#L9-L21

PR #667, opened 2024-06-26
✨ Allow to preserve previous snapshot version that will be available in commit middleware

dawidreedsy:
Occasionally, it is necessary to reference the previous version of a snapshot to execute auxiliary operations within the commit middleware, for example:

Let's imagine we have this doc that describes user:
```typescript
{
  id: '123',
  emailConfirmed: false;
}
```
Now we would want to send notification to the admin whenever `emailConfirmed` changed from `false` to `true`

At the moment there is no simple way to check in the middlewares if something is changing, apart form trying to apply the ops manually in the `apply` middleware. Something like

```typescript
backend.use('apply', function(request, next) {
  const changedSnapshot = applyOps(clone(request.snapshot), request.op);
  if(
    request.snapshot.data.emailConfirmed === false &&
    request.snapshot.data.emailConfirmed === true
  ) {
    notifyAdmin();
  }
})
```

This approach has few drawbacks:
1. We need to apply ops twice once in this middleware and once in the sharedb itself.
2. It is quite difficult for app to make considerations and mimick the flow of `$fixOps`

Solution
---
Let's make it possible to opt-in for storing the pre-apply snapshot version in request by setting a `preservePreapplySnapshot` flag to `true` in any middleware that happens before apply. This will allow us to:
1. Have the old and new version of snapshot available in the `commit` middleware
2. We do not have much memory overhead as this is opt-in, so if user never needs it the snapshot would never be cloned.
3. We only need to apply ops once
4. The apply mechanism is using shareDb power house.

### DIFF
--- a/lib/submit-request.js
+++ b/lib/submit-request.js
@@ -2,6 +2,7 @@ var ot = require('./ot');
 var projections = require('./projections');
 var ShareDBError = require('./error');
 var types = require('./types');
+var util = require('./util');
 
 var ERROR_CODE = ShareDBError.CODES;
 
@@ -42,6 +43,8 @@ function SubmitRequest(backend, agent, index, id, op, options) {
   this.ops = [];
   this.channels = null;
   this._fixupOps = [];
+  this.preservePreapplySnapshot = false;
+  this.preapplySnapshot = null;
 }
 module.exports = SubmitRequest;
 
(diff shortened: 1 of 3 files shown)


lib/submit-request.js
```diff
@@ -180,6 +183,15 @@ SubmitRequest.prototype.apply = function(callback) {
   this.backend.trigger(this.backend.MIDDLEWARE_ACTIONS.apply, this.agent, this, function(err) {
     if (err) return callback(err);
 
+    /**
+     * Middleware is able to opt in to store old snapshot
+     * which could be useful in cases where we want to compare the old snapshot
+     * (snapshot without the ops applied) and the new snapshot (with the ops applied)
+     */
+    if (request.preservePreapplySnapshot) {
+      request.preapplySnapshot = util.clone(request.snapshot);
```
alecgibson, line 192: Now that I see this written down, is this any more advantageous than consumers just doing this themselves in middleware directly?

With this approach:

```js
backend.use('apply', (request, next) => {
  request.preservePreapplySnapshot = true;
});
```

Or we can just do this with no changes to ShareDB at all:

```js
backend.use('apply', (request, next) => {
  request.preapplySnapshot = clone(request.snapshot);
});
```

dawidreedsy: Yeah actaully you are right, but there is an argument to be made that it should be part of sharedb itself for simplicity sake, but I am fine to just close this PR and just add some example of this use case in the docs.

dawidreedsy: We decided to just add the documentation about this use case

PR #671, opened 2024-07-20
🐛 Fix inflight/pending callbacks execute in wrong order

benjismith:
I encountered a situation in ShareDB where my callbacks are often executed in the wrong order.

I tracked this down to incorrect behavior in the `doc.fetch()` method, which pushes the `inflightCallback` onto the back of the `callbacks` array. If there are already any `pendingCallbacks` in the array, then the `inflightCallback` will be executed AFTER all the `pendingCallbacks`.

This is obviously not the desired behavior. The `inflightCallback` should always be executed FIRST, before any of the `pendingCallbacks`.

I fixed the code and wrote a test demonstrating the fix. If you remove my bugfix and run the test, you'll see that it reliably fails.

### DIFF
--- a/lib/client/doc.js
+++ b/lib/client/doc.js
@@ -295,7 +295,7 @@ Doc.prototype._handleFetch = function(error, snapshot) {
   var callbacks = this.pendingFetch;
   this.pendingFetch = [];
   var callback = this.inflightFetch.shift();
-  if (callback) callbacks.push(callback);
+  if (callback) callbacks.unshift(callback);
   if (callbacks.length) {
     callback = function(error) {
       util.callEach(callbacks, error);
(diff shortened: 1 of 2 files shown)


alecgibson:
Thanks for the submission! I've looked into the build failure; looks like it's an (upstream Node.js issue)[https://github.com/npm/cli/issues/7657] 😬 

Have run the tests (and linting) locally and it works fine for me.

benjismith:
Any news?

alecgibson:
@benjismith sorry our PR meeting last week was cancelled. @ericyhwang you around to discuss today? Or happy to take a look async?

ericyhwang (approved):
Sorry for the delay, I was out last week. Makes sense to me.

alecgibson:
Released in https://github.com/share/sharedb/releases/tag/v5.0.3

benjismith:
Awesome, thank you!

PR #678, opened 2024-08-27
fix(docs): mistake ws usage

electroluxcode:
hi, it seem no work when i try run code as docs 
then i clone this respository，and found that it may be a mistake  about docs

usage


alecgibson (approved):
Thanks for the contribution!

PR #679, opened 2024-08-27
⚡️ Cache latest op version when broadcasting presence

alecgibson:
At the moment, when sending a presence update to other subscribers, we [call `transformPresenceToLatestVersion()`][1] for every presence update which internally [calls `getOps()`][2] for every presence update.

Calls to `getOps()` can be expensive, and rapid presence updates may cause undue load on the server, even when the `Doc` has not been updated.

This change tries to mitigate this by subscribing to a pubsub stream for any `Doc`s that an `Agent` tries to broadcast presence on. We keep an in-memory cache of the latest snapshot version sent over this stream, which lets us quickly check if a presence broadcast is already current without needing to query the database at all.

To avoid leaking streams, the `Agent` will internally handle its stream subscription state by:

 - subscribing whenever a non-`null` presence update is broadcast
 - unsubscribing whenever a `null` presence update is broadcast

This means that rapid changes in presence being `null` or not can still result in database calls, but even in this case they should be less bad than before, because we only perform a snapshot fetch instead of ops.

[1]: https://github.com/share/sharedb/blob/297ce5dc66563a5955311793a475768d73ac8b87/lib/agent.js#L804
[2]: https://github.com/share/sharedb/blob/297ce5dc66563a5955311793a475768d73ac8b87/lib/backend.js#L919

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -830,16 +830,7 @@ Agent.prototype._broadcastPresence = function(presence, callback) {
       var latestDocVersion = util.dig(agent.latestDocVersions, presence.c, presence.d);
       var presenceIsUpToDate = presence.v === latestDocVersion;
       if (!presenceIsUpToDate) {
-        // null presence can't be transformed, so skip the database call and just
-        // set the version to the latest known Doc version
-        if (presence.p === null) {
-          transformer = function(agent, presence, callback) {
-            presence.v = latestDocVersion;
-            callback(null, presence);
-          };
-        } else {
-          transformer = backend.transformPresenceToLatestVersion.bind(backend);
-        }
+        transformer = backend.transformPresenceToLatestVersion.bind(backend);
       }
 
       transformer(agent, presence, function(error, presence) {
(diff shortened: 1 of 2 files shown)


lib/agent.js
```diff
@@ -794,25 +808,83 @@ Agent.prototype._broadcastPresence = function(presence, callback) {
...
+      // op.v behind snapshot.v by 1
+      latestDocVersions[collection][id] = op.v + 1;
+    });
+
+    callback();
+  });
+};
+
+Agent.prototype._unsubscribeDocVersion = function(collection, id, callback) {
+  var stream = util.dig(this.latestDocVersionStreams, collection, id);
```
ericyhwang, line 883: This doesn't clean up the `this.latestDocVersions`, which means a long-lived client that goes through many docs could have an agent with tons of old entries in that map.

Per discussion, immediately cleaning up `this.latestDocVersions` may make the case of multiple null presences worse. Perhaps a delayed cleanup or something.

ericyhwang (approved):

PR #680, opened 2024-08-27
⚡️ Deduplicate `Backend.getOps()` calls

alecgibson:
Calls to `Backend.getOps()` can be quite expensive. In order to minimize the impact of these calls on the server, this change deduplicates concurrent calls with the same arguments, and makes only a single call, then invoking all the callbacks with the single result.


lib/util.js
```diff
@@ -98,6 +98,32 @@ exports.clone = function(obj) {
   return (obj === undefined) ? undefined : JSON.parse(JSON.stringify(obj));
 };
 
+exports.deduplicateRequests = function(fn) {
+  var callbacksByArgs = {};
+  return function() {
+    var args = Array.prototype.slice.call(arguments);
+    var callback = args.pop();
+    var argString = JSON.stringify(args);
```
ericyhwang, line 106: `agent.custom` could in theory contain non-JSON-serializable objects, and also in practice could result in calls from different clients not being deduped.

We'll have to think on this a bit more - Alec suggested possibly moving it to sharedb-mongo instead

PR #682, opened 2024-10-03
⚡️ Avoid committing no-ops to the database

alecgibson:
This change checks for no-ops just before committing, and avoids writing them to the database.

Motivation
----------
Sometimes we have situations where multiple clients try to make an identical change to the same document at the same time (if multiple clients are trying to keep two documents in sync, for example).

When this happens, this can result in a `O(n^2)` retry cascade, since the requests will all keep retrying, with only a single request succeeding at a time, even if those ops are all no-ops.

Solution
--------
This change adds a new special `ERR_NO_OP` error code. If the server detects a no-op, it will skip committing, and early-return this error code, which takes the request out of the retry queue.

The client then has some special handling for this case, where it will:

 1. Fetch from remote to get up-to-date with any potential conflicting ops
 2. Then acknowledge the in-flight op, **without** bumping the version (since the op wasn't committed)

Note that if `$fixup()` has been used, we **do not** early-return a no-op error, even if the fixup results in a no-op, since the client would be left in an inconsistent state without the fixup ops being returned and applied.

## Depends on:
 - [x] https://github.com/share/sharedb/pull/684


lib/submit-request.js
```diff
@@ -204,6 +204,13 @@ SubmitRequest.prototype.commit = function(callback) {
       };
     }
 
+    if (request.op.op && request.op.op.length === 0 && request._fixupOps.length === 0) {
+      // The op is a no-op, either because it was submitted as such, or - more
+      // likely - because it was transformed into one. Let's avoid committing it
+      // and tell the client.
+      return callback(request.noOpError());
```
alecgibson, line 211: This check has to happen at some point after the [op is transformed](https://github.com/share/sharedb/blob/945dc50ebc93d53e7306c522226bd804a802bc49/lib/submit-request.js#L152), and before the [op is committed](https://github.com/share/sharedb/blob/945dc50ebc93d53e7306c522226bd804a802bc49/lib/submit-request.js#L208).

I don't think I have a strong opinion on when we return this, so long as it's at either one of these extremes(?).

The benefit of doing it just before `commit()` is that it's a bit more directly linked to the fact that commit is about to be skipped. It does mean that all the middleware will be fired, though (which may or may not be a good/expected thing?).

lib/client/doc.js
alecgibson, line 1: Should this be an opt-in feature behind a flag? I don't know if this should be considered a breaking change?

alecgibson: Okay I wound up:
 - hiding this behind a feature flag
 - bumping the protocol to `1.2` and adding a protocol check, so that old clients won't get `ERR_NO_OP` errors that they don't know how to handle

lib/submit-request.js
ericyhwang: After some discussion, we decided to move just after transform, before the "apply" hook.

The "commit" hook right now means that sharedb core is going to call the DB adapter's `commit` method, and leaving the no-op early return here would have meant a violation of that behavior. By moving it earlier, that behavior is preserved.

ericyhwang (approved):

PR #684, opened 2024-10-07
✨ Store client protocol on `Agent`

alecgibson:
When developing new features, it will be useful to be able to compare server and client protocol versions.

For example, if a server is ahead of an old client, the server could gracefully degrade performance for clients that don't support the new functionality.

This change stores the client's protocol on the `Agent` so we can perform these checks.

We also add a `checkAtLeast()` protocol helper function, and store the current protocol in a constant to avoid magic numbers.

### DIFF
--- a/lib/client/connection.js
+++ b/lib/client/connection.js
@@ -11,6 +11,7 @@ var types = require('../types');
 var util = require('../util');
 var logger = require('../logger');
 var DocPresenceEmitter = require('./presence/doc-presence-emitter');
+var protocol = require('../protocol');
 
 var ERROR_CODE = ShareDBError.CODES;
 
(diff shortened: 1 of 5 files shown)


lib/protocol.js
```diff
@@ -0,0 +1,27 @@
...
+  major: 1,
+  minor: 1,
+  checkAtLeast: checkAtLeast
+};
+
+function checkAtLeast(toCheck, checkAgainst) {
+  toCheck = normalizedProtocol(toCheck);
+  checkAgainst = normalizedProtocol(checkAgainst);
+  return toCheck.major >= checkAgainst.major &&
+    toCheck.minor >= checkAgainst.minor;
```
ericyhwang: ⚠️ This check would fail 2.0 > 1.1

lib/client/connection.js
```diff
@@ -753,7 +759,7 @@ Connection.prototype._handlePingPong = function(error) {
 Connection.prototype._initialize = function(message) {
   if (this.state !== 'connecting') return;
 
-  if (message.protocol !== 1) {
+  if (message.protocol !== protocol.major) {
```
ericyhwang, line 762: Confirming that we intend for a client to only support servers on the same major protocol version, disallowing a newer client to connect to a server on an older major version?

ericyhwang: Confirmed that is intended after discussion. That situation would only potentially happen during a rolling update that includes a major protocol update; or a server version rollback under the same situation.

ericyhwang (approved):

PR #689, opened 2025-01-21
Transactions

alecgibson:


test/client/transaction.js
```diff
@@ -0,0 +1,219 @@
...
+          remoteDoc.fetch.bind(remoteDoc),
+          function(next) {
+            expect(remoteDoc.data).to.eql({tricks: ['sit', 'shake']});
+            expect(doc.data).to.eql(remoteDoc.data);
+            next();
+          }
+        ], done);
+      });
+
+      it('remote submits after but commits first', function(done) {
```
alecgibson, line 150: This is the test case that managed to force my [MongoDB driver update](https://github.com/share/sharedb-mongo/pull/163) into deadlock.

I added a lot of complexity into the `sharedb-mongo` update to get this test case to pass, but I wonder if we need to.

The main issue is that we do this:

1. Wait for `doc` to submit an op to a transaction
2. `remoteDoc` tries to update the same document outside of this transaction
3. We try to wait for `remoteDoc.submitOp` to resolve
4. However, it seems that MongoDB locks the snapshot document until `doc`'s transaction is committed, so `remoteDoc`'s submission stalls, introducing deadlock

We can trivially rewrite this test case to remove the deadlock by not having such a strict series of awaiting — if we run the two transactions in parallel with no awareness of one another (which is how all transactions will run in practice in the real world), then no deadlock happens, because `doc` will commit its transaction,  and `remoteDoc` will finally return with a failed write and retry, then eventually succeed.

The reason I didn't write it like this originally is that I was trying to force us into an edge case where both transactions attempt to touch the database, and running these operations in parallel removes a lot of the control I have over logical paths we take through ShareDB's recovery/retry machinery. However, is it worth enforcing this as part of the spec just for this? I'm not sure it is...?

lib/submit-request.js
```diff
@@ -225,30 +255,40 @@ SubmitRequest.prototype.commit = function(callback) {
       request.snapshot,
       request.options,
       function(err, succeeded) {
+        // TODO: Should this callback be called before committing a transaction?
```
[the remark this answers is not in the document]
ericyhwang: When I was thinking about this pre-holidays, I was thinking of:
- `startTransaction` creates a random session id locally, and causes subsequent `submitOp` with the transaction to queue locally
- `transaction.commit()` causes all the locally queued ops to be sent to the server. Then the driver does whatever it needs to do with the batch of ops, based on the database.
- Any errors are reported back in `transaction.commit()`

Per discussion today, we do think the approach of local op queueing does make the implementation, middleware, and testing easier, so we're going with that.

lib/db/memory.js
```diff
@@ -133,6 +150,47 @@ MemoryDB.prototype.query = function(collection, query, fields, options, callback
...
+  // let's just save the DB state before and restore if the transaction fails
+  var docs = JSON.stringify(this.docs);
+  var ops = JSON.stringify(this.ops);
+
+  for (var commit of transaction.commits) {
+    var error = commit();
+
+    if (error) {
+      this.docs = JSON.parse(docs);
+      this.ops = JSON.parse(ops);
```
ericyhwang, line 181: With potentially large sets of docs/ops in tests, using a structured clone would be better for performance/memory. You mentioned rfdc is working well for you in other situations, and I've looked at in the past and it seems good.

alecgibson:
As discussed in https://github.com/share/sharedb/pull/689#discussion_r1924151892 we're moving to a client-side buffer approach, which was basically a total rewrite: https://github.com/share/sharedb/pull/690

PR #692, opened 2025-03-03
🐛 Unhandled rejection when submitting ops during hard rollback.

dawidreedsy:
After doing the steps:
1. Create a doc with rich text type (or any other non irreversible type)
2. Make op submission fail
3. Now in the hard rollback we do `this._setType(null);`
4. If there is any op comming before the hard rollback `fetch` is finished, we get the error
```
Cannot submit op. Document has not been created.
```
as in the `_submit` we do:
```typescript
    if (!this.type) {
      var err = new ShareDBError(
        ERROR_CODE.ERR_DOC_DOES_NOT_EXIST,
        'Cannot submit op. Document has not been created. ' + this.collection + '.' + this.id
      );
      if (callback) return callback(err);
      return this.emit('error', err);
    }
```

We definitely do not handle this case properly. Possible solutions:
1. Just throw error whenever that happens, which is easy to implement and it is not really breaking. User would be then able to react on the error or just ignore it.
2. Store copy of cannonical snapshot in the doc itself, so that we do not have to do fetch for hard rollback. More difficult to implement and has a side effect of storing the doc twice in the memory.


lib/client/doc.js
```diff
@@ -748,6 +750,15 @@ Doc.prototype._submit = function(op, source, callback) {
   // The op contains either op, create, delete, or none of the above (a no-op).
   if ('op' in op) {
     if (!this.type) {
+      if (this._isInHardRollback) {
+        var err = new ShareDBError(
+          ERROR_CODE.ERR_DOC_IN_HARD_ROLLBACK,
+          'Cannot submit op. Document is performing hard rollback. ' + this.collection + '.' + this.id
+        );
```
alecgibson, line 757: You could just move this after the `var err = new ShareDBError(ERR_DOC_DOES_NOT_EXIST)` to override this "default" error and then avoid having to duplicate the `emit()` machinery

alecgibson (commented):
Test, please!

dawidreedsy:
Yes i know want to preare the test today, just dneed to fiddle a bit to make it work.

dawidreedsy:
To be honest i think there might be more to this than just my fix as before fetch we do 
```js
  this._setType(null);
  this.version = null;
  this.inflightOp = null;
  this.pendingOps = [];
```
and after fetch is finished
```js
    if (inflightOp) pendingOps.unshift(inflightOp);
    var allOpsHadCallbacks = !!pendingOps.length;
    for (var i = 0; i < pendingOps.length; i++) {
      allOpsHadCallbacks = util.callEach(pendingOps[i].callbacks, err) && allOpsHadCallbacks;
    }
    if (err && !allOpsHadCallbacks) doc.emit('error', err);
    doc._isInHardRollback = false;
```,

I think it was menat to reject all ops that were added after fetch started. I am not sure just throwing the above error wouldn't break anything. Even though all tests are passing.

File: lib/client/doc.js [hunk 2]
```diff
@@ -1077,7 +1088,10 @@ Doc.prototype._hardRollback = function(err) {
         inflightOp = null;
       }
 
-      if (!pendingOps.length) return;
+      if (!pendingOps.length) {
+        doc._isInHardRollback = false;
```
ericyhwang: If we want to avoid potential future issues with forgetting to reset `_isInHardRollback` on early returns, we could reset to false at the top of this `_fetch` callback

ericyhwang (approved):

PR #693, opened 2025-03-06
 🐛 Wrong error when submitting presence during hard rollback.

dawidreedsy:
Similar case to:
 - https://github.com/share/sharedb/pull/692

 however instead of submitting ops to doc is when we submit doc presence

### DIFF
--- a/lib/client/presence/local-doc-presence.js
+++ b/lib/client/presence/local-doc-presence.js
@@ -30,10 +30,20 @@ LocalDocPresence.prototype.submit = function(value, callback) {
     // be null. Let's early return, instead of error since this is a harmless
     // no-op
     if (value === null) return this._callbackOrEmit(null, callback);
-    var error = {
-      code: ERROR_CODE.ERR_DOC_DOES_NOT_EXIST,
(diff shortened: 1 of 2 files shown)


lib/client/presence/local-doc-presence.js
```diff
@@ -30,10 +30,20 @@ LocalDocPresence.prototype.submit = function(value, callback) {
     // be null. Let's early return, instead of error since this is a harmless
     // no-op
     if (value === null) return this._callbackOrEmit(null, callback);
-    var error = {
-      code: ERROR_CODE.ERR_DOC_DOES_NOT_EXIST,
-      message: 'Cannot submit presence. Document has not been created'
-    };
+
+    var error = null;
+    if(this._doc._isInHardRollback) {
```
dawidreedsy, line 35: I decided to not change `_isInHardRollback` to `isInHardRollback` as i think it should not be used out side of lib. Feels a bit code smell TBH. I am open for suggesstions.

```suggestion
    if(this._doc._isInHardRollback) {
```

alecgibson (approved):

PR #695, opened 2025-05-14
🐛 Fix hasOwn function implementation to ensure proper property checking

welkinwong:
In browsers that lack native `Object.hasOwn` support, assigning `Object.prototype.hasOwnProperty.call` to a variable (e.g., hasOwn = `Object.prototype.hasOwnProperty.call`) detaches the `call` method from its original function context.   This results in a runtime error: `hasOwn is not a function`.

image

image

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -9,7 +9,9 @@ exports.hasKeys = function(object) {
 };
 
 var hasOwn;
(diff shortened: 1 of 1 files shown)


ericyhwang (approved):
Thank you for the fix! We're unsure how the `.call` got through in the first place.

Coverage shows a drop, but that's fine since it's hard to test the fallback in any remotely recent version of Node.

PR #696, opened 2025-07-24
✨ Allow sorting middleware

alecgibson:
At the moment middleware is executed in the order that it is registered.

This is fine if middleware is only registered in a single place, but in distributed setups, there may be multiple places in code (possibly across repos) that consumers may want to register middleware.

In order to allow some distributed way of defining an execution order of middlewares, this is a non-breaking change that lets consumers define an `order` for their middleware when registering:

```js
backend.use('commit', fn, 10);
```

This `order` will be used to sort the middlewares, such that higher `order` execute after middleware with lower `order`. Ties are broken on registration order (the current behaviour), since JavaScript's `.sort()` [is stable][1].

The default value for `order` is `0`, so if consumers want to run middleware before other middleware that has not specified an `order`, they will need to set a negative `order` (which is legal):

```js
backend.use('commit', fn, -10);
```

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort#sort_stability

### DIFF
--- a/lib/backend.js
+++ b/lib/backend.js
@@ -183,15 +183,19 @@ Backend.prototype.addProjection = function(name, collection, fields) {
 /**
  * Add middleware to an action or array of actions
  */
-Backend.prototype.use = function(action, fn) {
+Backend.prototype.use = function(action, fn, order) {
   if (Array.isArray(action)) {
     for (var i = 0; i < action.length; i++) {
-      this.use(action[i], fn);
+      this.use(action[i], fn, order);
     }
     return this;
   }
+  fn.__order = order || 0;
   var fns = this.middleware[action] || (this.middleware[action] = []);
   fns.push(fn);
+  fns.sort(function(a, b) {
+    return a.__order - b.__order;
+  });
   return this;
 };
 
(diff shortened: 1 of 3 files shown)


lib/backend.js
```diff
@@ -183,15 +183,19 @@ Backend.prototype.addProjection = function(name, collection, fields) {
...
-Backend.prototype.use = function(action, fn) {
+Backend.prototype.use = function(action, fn, order) {
   if (Array.isArray(action)) {
     for (var i = 0; i < action.length; i++) {
-      this.use(action[i], fn);
+      this.use(action[i], fn, order);
     }
     return this;
   }
+  fn.__order = order || 0;
```
alecgibson, line 193: I don't know how hacky/icky it is to define a property on the middleware function itself.

Ideally I would have wrapped this into an object or something:

```js
{fn, order}
```

...but `this.middleware` isn't prefixed, so consumers may technically assume it as part of the public API.

If I had access to modern ES features, I'd maybe use a `WeakMap` to avoid mutating the function, but we still technically target ES3.

alecgibson: There's also a bug here where if someone registers the same middleware on 2 different servers (in the same thread) with different orders, it'll overwrite the first order with the second order

PR #706, opened 2026-04-11
Update dev dependencies, replace Node 18 with Node 24 in CI

ericyhwang:
1. **Update to latest dev dependencies** (eslint, mocha, chai, nyc, sinon)
    - ESLint now requires the newer "flat config". I used the eslint migration tool to produce eslint.config.js based on the existing .eslintrc.js.
    - [sinon@19](https://sinonjs.org/guides/migration-guide.html#sinon-19) started stubbing `nextTick` by default, which causes some tests using fake timers to fail.
        - At first, I tried using the new option `clock.setTickMode({mode: 'nextAsync'});`, but that caused the sharedb-mongo tests' Mocha process to hang after test suite completion. Went down a debugging rabbit-hole, and ended up filing https://github.com/sinonjs/fake-timers/issues/564
        - To avoid the issue, I've added a `useFakeTimers` wrapper in test/util.js that sets a default `toFake` list with just the ones that matter to us, and I've also switched the `clock.tickAsync()` calls (which weren't being awaited/then-ed) to `clock.tick()`.
    - Removed a very ancient `test/.jshintrc` file, we missed it when migrating to ESLint
2. **In CI, remove Node 18 and add Node 24**
    - Node 18's been end-of-life for almost a year, Node 24 is the latest LTS.
    - CI will now run with Node 20, 22, and 24.

Technically, we should major version for dropping Node 18, though the sharedb library itself should still run fine on Node 18 since it doesn't use anything in Node 20+. Note that Node 20's going EoL in less than a month, so we could just wait until then to major version.


alecgibson (approved):
Nice one!

PR #707, opened 2026-04-13
Initial TypeScript setup and conversion (CJS to ESM, ES5 class functions to ES6 classes)

ericyhwang:
## Summary

Set up TypeScript, do file renames from *.js to *.ts, then run codemods to convert CommonJS to ES modules and ES5 class functions to ES6 classes.

This is a redo of https://github.com/share/sharedb/pull/705 without the dev dependency upgrades, which I moved out into https://github.com/share/sharedb/pull/706 and are already merged into master.

Couple bigger codemod changes here on top of the older PR:
- Non-function prototype properties, e.g. `Backend.prototype.MIDDLEWARE_ACTIONS = ...`, are declared as fields, with the assignment moved into `static` blocks to reduce the diff size.
- Comments on prototype property assignments are ported over to the class field definitions. If such comments are one or more single-line `// ` comments, they are converted to JSDoc-style `/** */` comments.

## Goals

It targets the `typescript` branch, which we can use to stage the TypeScript changes until they're ready to merge into main. Making smaller intermediate PRs into that branch will make things easier to review.

For now, my goal with the `typescript` staging branch is to end up with TS source code that compiles into JS output compatible with the current JS source, so we could publish it a minor version. In a future major version, we can choose to update the compile target to drop ES5 support, especially if we want to start using things like async/await without the extra verboseness of downleveling.

We should probably major version soon anyways, since Node 18's been EoL for a year, and Node 20's about to reach EoL.

### Why target ES5 and not ES3 or ES6?

While the ESLint config currently specifies ES3, our JS code actually uses ES5-specific standard library functions like JSON.parse/stringify, Object.create(), .bind(). etc., so the tsconfig has to specify at least ES5. Plus, [TypeScript 5.5 removed support for ES3 output](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#disabling-features-deprecated-in-typescript-50).

Why not an ES6 compile target? That would mean the TS compiler outputs `class` syntax, `const`/`let` instead of converting to `var`, etc. That is techncially a breaking change from the current JS code's ES5 syntax, and even though all remotely modern browsers support ES6 now, there is a less immediately obvious breaking change - You can't use old-style prototypical inheritance to call a ES6 superclass. If you try to do `MyBaseClass.call(this)` in the subclass constructor function, you'll get an error `TypeError: Class constructor MyBaseClass cannot be invoked without 'new'`. A couple references:

https://stackoverflow.com/questions/51344127/proto-inheritance-from-es6-class
https://stackoverflow.com/questions/33369131/is-it-possible-for-an-old-style-class-to-inherit-from-an-ecmascript-6-class-in-j
In practice what that means, is to upgrade to ES6 classes at runtime, you have to upgrade "leaf nodes" in the class tree to ES6 classes first, then upgrade their direct parents, and so forth up the class chain. (An ES6 class can call super() to invoke a function/prototype base class just fine.)

For ShareDB itself, we'd want to upgrade the leaf-node libraries like sharedb-mongo to ES6 class output first, then sharedb core. We'd also want to indicate in the major version release notes that consumers must either switch to ES6 class output themselves, or use a transpiler like Babel on sharedb code.

## Reviewer tips and notes
- I recommend going through the commits one by one.
- Unfortunately, GitHub's new diff viewer doesn't respect the "Ignore whitespace" option when you have it show a file with a large diff, so for the codemod commit, it'd be better to pull this `typescript-initial` branch down and locally view the commit's diff, e.g. in your editor.
- If you would like to compare the original JS source with the tsc-produced JS output, I have that pushed up to the `typescript-cjs-class-codemods-js-outputs` branch, in commit a4afeec. Similarly, it's better to pull the branch and compare locally.

Let me know if you want to schedule time to go through this on a call!

## Changes

1. **Set up TypeScript**, rename all `lib/**/*.js` files to `src/**/*.ts` with no changes
    - Doing a commit with simple moves/renames guarantees that Git can detect the renames, making it easier to use `git blame` and other history spelunking tools on the "new" TS files.
2. Pre-codemod manual changes
    - Move some `module.exports` statements up. Some class JSDoc comments were attached to the `module.exports` lines. Moving them so the comments are attached to the class functions means the codemod (next step) can do a better job porting the comments over to the ES6 class.
    - Minor tweaks to next-tick.ts and util.ts to eliminate duplicate declaration issues post-codemod.
3. **Run jscodeshift** codemod-cjs-to-esm.ts and codemod-es5-classes.ts
    - Codemods run really fast, and a given codemod and input will consistently produce the same output, unlike LLM-based gen AI. That said, I did use gen AI to help write these codemods.
    - Background and details:
      - Many years ago, for Lever's TS conversions, I hand-wrote the codemods. LLMs were still in relative infancy and not as well-known back then.
      - Since I no longer have access to Lever code, I've made new ground-up codemods with gen-AI assistance, focused on converting ShareDB source code.
      - Specifically, I used Google's Antigravity IDE (their agent-focused editor) with a mix of Gemini 3.1 Pro and Gemini 3 Flash models. It did a pretty good job, though as expected I did have to go through many rounds of prompting and refining requirements to handle various cases.
      - At the end, I went through and re-read all the code in the codemods, doing light refactoring and adding comments as needed.
    - The codemods - they're gists for now until I get around to putting them in their own repo:
      - https://gist.github.com/ericyhwang/4ef3a6b02d0b86a4a7beda4d57520ccb
      - https://gist.github.com/ericyhwang/32046cc7bcf569c34fda1afefa4d4b0c

## Next steps

First thing after this PR is to go through the many TS compilation errors and fix them. Most fall into these categories:
- Declaring more class fields/methods, missing either due to assignments outside the constructor or due to inheritance from built-in classes not being able to use `extends`
- Marking optional params
- Typings for properties added onto object literals after initial creation

Then it'll be adding stronger types, likely based on our DefinitelyTyped definitions; doing more cleanup as needed; and converting test files to TS.


tsconfig.json
```diff
@@ -8,12 +8,20 @@
...
     ],
     "strict": true,
-    "esModuleInterop": true,
+    "esModuleInterop": false,
     "skipLibCheck": true,
     "forceConsistentCasingInFileNames": true,
     "ignoreDeprecations": "6.0",
     "noImplicitAny": false,
     "noImplicitThis": false,
+    "removeComments": false,
```
alecgibson: Why? JSDoc or something?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
+{
+  "compilerOptions": {
+    "rootDir": "./src",
+    "outDir": "./lib",
+    "target": "ES5",
```
alecgibson, line 5: How did you decide this?

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "module": "CommonJS",
```
alecgibson, line 6: (Switching to ESM at some point is going to be fun 🙄 )

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    "types": [
+      "node"
```
alecgibson, line 8: I'm not sure I like this. On a selfish note, I think this will pollute our own code's types, because we run `backend.js` even in the browser 😅 

But from a less selfish perspective, this shouldn't apply to `client/` which is designed for browser consumption. I think we possibly want 2 different `tsconfig.json` files? One for backend stuff, and one for client?

In an ideal world I think backend and client would be 2 separate packages in a monorepo, but that's obviously out of scope of this change.

tsconfig.json
```diff
@@ -0,0 +1,21 @@
...
+    ],
+    "strict": true,
+    "esModuleInterop": true,
+    "skipLibCheck": true,
```
alecgibson: 😢 Why?

tsconfig.json
ericyhwang: Since a goal in this first phase is to keep the JS output compatible with the original JS code (prototype-based classes, `var`, no for-of), ES5 was really the only choice. An ES6 compile target would mean the TS compiler outputs `class` syntax, `const`/`let` as in the source instead of converting to `var`, etc.

This can't be ES3 because our JS code uses ES5-specific standard library functions like `JSON.parse`/`stringify`, `Object.create()`, `.bind()`. etc. Plus, [TypeScript 5.5 removed support for ES3 output](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#disabling-features-deprecated-in-typescript-50).

----

So why not go with ES6 output since all remotely modern browsers support ES6 now?

There would be a less immediately obvious breaking change, where you can't use old-style prototypical inheritance to call a ES6 superclass. If you try to do `MyBaseClass.call(this)` in the subclass constructor function, you'll get an error `TypeError: Class constructor MyBaseClass cannot be invoked without 'new'`. A couple references:
* https://stackoverflow.com/questions/51344127/proto-inheritance-from-es6-class
* https://stackoverflow.com/questions/33369131/is-it-possible-for-an-old-style-class-to-inherit-from-an-ecmascript-6-class-in-j

In practice what that means, is to upgrade to ES6 classes at runtime, you have to upgrade "leaf nodes" in the class tree to ES6 classes first, then upgrade their direct parents, and so forth up the class chain. (An ES6 class can call `super()` to invoke a function/prototype base class just fine.)

For ShareDB itself, we'd want to upgrade the leaf-node libraries like sharedb-mongo to ES6 class output first, then sharedb core. We'd also want to indicate in the major version release notes that consumers must either switch to ES6 class output themselves, or use a transpiler like Babel on sharedb code.

tsconfig.json
ericyhwang: The `require(esm)` feature in Node 20+ makes it less of a pain than before, thankfully: 
https://joyeecheung.github.io/blog/2025/12/30/require-esm-in-node-js-from-experiment-to-stability/

tsconfig.json
ericyhwang: This shouldn't affect any consumer code, since we compile to JS (plus eventually d.ts) and publish that.

The TS compiler only cares about the main tsconfig you point it at. For any precompiled dependencies, it just reads the d.ts files.

----

ShareDB client code already assumes access to Node libraries, necessitating a bundler:
* Dynamic calls to `require()`, such as https://github.com/share/sharedb/blob/3552c286c687ae8879e98ddfc9415b2e9da7f157/lib/util.js#L114
* Type definitions for the Node built-in 'events' library, which is directly used in client code from https://github.com/share/sharedb/blob/3552c286c687ae8879e98ddfc9415b2e9da7f157/lib/client/presence/doc-presence-emitter.js#L2 and also indirectly used in more places via emitter.js

The `"types": ["node"]` just formally codifies that for the TS compiler.

If we really wanted to do an internal split of tsconfig files, I think we'd want three - client, server, and shared - but even then, we'd still need to reference the Node types in client/shared code due to the two points above. To eliminate that, we'd have to rewrite the shared/client code to avoid Node built-in libraries or functions.

tsconfig.json
ericyhwang: With this enabled, tsc will still type-check d.ts files that are directly or indirectly referenced by your project code, it just skips the default behavior of checking all d.ts files in node_modules.

https://www.typescriptlang.org/tsconfig/#skipLibCheck

I've just gotten used to using it in larger projects, where this can speed up compilation times by quite a bit, but we probably don't need it for ShareDB's relatively small dependency set. I'll remove it.

tsconfig.json
ericyhwang: Yeah, keeping comments means that the output will have JSDoc, which is especially useful once we start publishing d.ts files in ShareDB itself.

Just checked, and this defaults to `false`, so I can leave it off.

tsconfig.json
alecgibson: Cool sounds sensible, just wanted it documented. Thanks!

tsconfig.json
alecgibson: Okay this is fine for now, but I'd like to sense check when we eventually ship type definitions that `@types/node` isn't being dragged into client code.

I think the DefinitelyTyped types currently achieve it with things like this:

```ts
import { Duplex } from "stream";
```

Which I think has the "expected behaviour" of:

 - in an environment where `@types/node` is available, everything is fine
 - in a browser/bundler environment, this fails compilation unless you've provided a polyfill and/or defined your own type stub (or do something hideous like `skipLibCheck: true`)

alecgibson (approved):
👏🏼  Nice to finally get this started! 🙌🏼

PR #719, opened 2026-07-31
🐛 Restore the op before fixups when retrying a submit

### LINKED ISSUE #716: $fixup is not restored across the submit retry loop, so it is applied more than once
## Summary

`$fixup()` composes its op permanently into `this.op.op`, but the `apply` middleware re-runs on every submit retry. `apply()` resets the fixup *bookkeeping* on each attempt while leaving `op.op` carrying the previous attempt's fixup, so a submit that loses a commit race commits the fixup twice — and acknowledges only one of them to the client.

Reproduced on `master` (`lib/submit-request.js` there is byte-identical to the published 6.0.0).

## Repro

No dependencies beyond sharedb's own. Save as `repro-fixup-retry.js` in the repo root and `node repro-fixup-retry.js`:

```js
const Backend = require('./lib/backend');
const json0 = require('ot-json0').type;

const backend = new Backend();
const COLLECTION = 'c';
const ID = 'd';

// A trivial apply middleware that always applies the same fixup. Nothing about it is
// stateful or non-idempotent: it computes the same F every time it is called.
const F = [{p: ['n'], na: 1}];
let applyCalls = 0;
backend.use('apply', (request, next) => {
  if (!('op' in request.op)) return next();
  applyCalls++;
  console.log(`  [apply #${applyCalls}] op.op in  = ${JSON.stringify(request.op.op)}`);
  request.$fixup(F);
  console.log(`  [apply #${applyCalls}] op.op out = ${JSON.stringify(request.op.op)}`);
  next();
});

const submit = (op) => new Promise((resolve, reject) => {
  backend.submit({custom: {}, backend, stream: {}}, COLLECTION, ID, op, null,
    (error, ops, request) => error ? reject(error) : resolve(request));
});
const getSnapshot = () => new Promise((resolve, reject) => {
  backend.db.getSnapshot(COLLECTION, ID, null, null, (e, s) => e ? reject(e) : resolve(s));
});

async function main() {
  await submit({create: {type: json0.uri, data: {n: 0}}, v: 0});

  // Make the first commit attempt lose the race, exactly as a concurrent commit does.
  const commit = backend.db.commit.bind(backend.db);
  let lostRace = false;
  backend.db.commit = function(collection, id, op, snapshot, options, callback) {
    if (!lostRace) {
      lostRace = true;
      console.log('  [db.commit] losing the race once -> ShareDB retries');
      return process.nextTick(() => callback(null, false));
    }
    return commit(collection, id, op, snapshot, options, callback);
  };

  const request = await submit({op: [{p: ['s'], oi: 'x'}], v: 1, src: 'client', seq: 1});
  const snapshot = await getSnapshot();

  console.log(`\n  committed op.op = ${JSON.stringify(request.op.op)}`);
  console.log(`  snapshot.data   = ${JSON.stringify(snapshot.data)}`);
  console.log(`  n = ${snapshot.data.n}, expected 1 (the middleware's fixup applied once)`);

  console.log(`\n  ack fixup ops = ${JSON.stringify(request._fixupOps)}`);
}

main().then(() => process.exit(0), (error) => {
  console.error(error);
  process.exit(1);
});
```

### Actual

```
  [apply #1] op.op in  = [{"p":["s"],"oi":"x"}]
  [apply #1] op.op out = [{"p":["s"],"oi":"x"},{"p":["n"],"na":1}]
  [db.commit] losing the race once -> ShareDB retries
  [apply #2] op.op in  = [{"p":["s"],"oi":"x"},{"p":["n"],"na":1}]
  [apply #2] op.op out = [{"p":["s"],"oi":"x"},{"p":["n"],"na":2}]

  committed op.op = [{"p":["s"],"oi":"x"},{"p":["n"],"na":2}]
  snapshot.data   = {"n":2,"s":"x"}
  n = 2, expected 1 (the middleware's fixup applied once)

  ack fixup ops = [{"src":"client","seq":1,"v":1,"op":[{"p":["n"],"na":1}]}]
```

### Expected

`n = 1`, and the committed `op.op` to contain the fixup once.

## Two distinct problems

**1. `op.op` accumulates fixups across attempts.** `$fixup` mutates the op in place:

https://github.com/share/sharedb/blob/master/lib/submit-request.js#L73-L74

```js
if (this.op.create) this.op.create.data = type.apply(this.op.create.data, op);
else this.op.op = type.compose(this.op.op, op);
```

`retry()` calls `submit()` again, which reaches `apply()` and re-triggers the middleware, so attempt 2 composes its fixup onto an op that already contains attempt 1's. With N retries the fixup lands N+1 times.

**2. The ack under-reports what was committed.** `apply()` clears the bookkeeping on each attempt:

https://github.com/share/sharedb/blob/master/lib/submit-request.js#L189-L190

```js
this._fixupOps = [];
delete this.op.m.fixup;
```

`Agent.prototype._submit` then sends only the surviving attempt's fixups:

```js
var ack = {src: op.src, seq: op.seq, v: op.v};
if (request._fixupOps.length) ack[ACTIONS.fixup] = request._fixupOps;
```

So in the repro the server commits `na: 2` while the client is told `na: 1`. This is a client/server divergence produced entirely by ShareDB's own bookkeeping — it happens even for a middleware whose fixup is a pure function of the request, as above.

I'd argue (2) is what makes this a bug rather than an undocumented constraint on middleware authors: clearing `_fixupOps` per attempt is only coherent if `op.op` is also restored per attempt, and nothing does that.

## Suggested fix

Restoring at the top of `apply()` isn't enough, because `_transformOp` legitimately mutates `op.op` between attempts (transforming over whatever won the race) and we want the *original* op transformed forward, not the fixed-up one.

Stashing before the middleware trigger and restoring in `retry()` gets the ordering right:

- in `apply()`, before `this.backend.trigger(...apply...)`, stash `this._preFixupOp = clone(this.op.op)` (and `this.op.create.data` for the create branch)
- in `retry()`, restore `this.op.op` from the stash before calling `submit()`

Then attempt 2 transforms the original op forward, the middleware recomputes its fixup against the new snapshot, and `_fixupOps` matches what was committed. `delete this.op.m.fixup` becomes redundant but harmless.

Worth deciding whether the `op.create` path needs the same treatment — `$fixup` mutates `op.create.data` there, and a create can also be retried.

## Docs

`docs/middleware/op-submission.md` describes `apply` as firing "when the snapshot has been fetched, and the op is about to be applied", with no mention that a submission can be retried or that `apply` may therefore run more than once for one submission. Worth stating explicitly whichever way this is resolved, since middleware written against the current docs has no reason to expect re-entry.

## Related

Found while auditing a migration middleware that uses `$fixup` to carry a non-idempotent content change; that's what makes the double-composition visible rather than benign. Existing `$fixup` users tend to write idempotent metadata, which masks it.

alecgibson:
Fixes https://github.com/share/sharedb/issues/716

At the moment, `$fixup()` composes its op permanently into `op.op` (or `op.create.data`), but the `apply` middleware is re-triggered on every submit retry. `apply()` resets the fixup *bookkeeping* on each attempt, while leaving `op.op` carrying the abandoned attempt's fixup, and nothing restores it.

This causes two problems for a submit that loses a commit race. First, `op.op` accumulates fixups: with N retries the fixup lands N+1 times. Second, the ack under-reports what was committed, since `Agent` only sends the surviving attempt's `_fixupOps`. So we can commit `na: 2` while telling the client `na: 1` — a client/server divergence at the same version, produced entirely by our own bookkeeping. It happens even for a middleware whose fixup is a pure function of the request; existing `$fixup` users tend to write idempotent metadata, which masks it.

Restoring at the top of `apply()` isn't enough, because `_transformOp` legitimately mutates `op.op` between attempts, and we want the *original* op transformed forward, not the fixed-up one. So this change stashes the op as it was before the attempt's first fixup, and resets all of the fixup state — the bookkeeping, and the op itself — in `retry()` before re-submitting. The middleware then recomputes its fixup against the newer snapshot, and `_fixupOps` matches what we actually commit.

The stash is taken lazily in `$fixup()` rather than unconditionally in `apply()`, so submits that don't fix anything up don't pay to clone their op. Cloning is necessary because OT types make no guarantees about mutating their inputs. It's wrapped in an object so that its own truthiness tells us whether this attempt has stashed yet: the stashed value can legitimately be falsy, since `checkOp()` only requires that `op.op` is present, and `op.create.data` is whatever the type's `create()` returned.

Note the `delete op.m.fixup` is *not* made redundant by this: `commit()` only sets `m.fixup` when `_fixupOps` is non-empty, so an attempt that fixes nothing up would otherwise commit an op still carrying the previous attempt's `m.fixup`. Resetting in `retry()` rather than `apply()` also covers the retry that finds our own op already committed, since that bails out during transform without ever reaching `apply()`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/lib/submit-request.js
+++ b/lib/submit-request.js
@@ -3,6 +3,7 @@ var projections = require('./projections');
 var ShareDBError = require('./error');
 var types = require('./types');
 var protocol = require('./protocol');
+var util = require('./util');
 
 var ERROR_CODE = ShareDBError.CODES;
 
@@ -43,6 +44,7 @@ function SubmitRequest(backend, agent, index, id, op, options) {
   this.ops = [];
   this.channels = null;
   this._fixupOps = [];
+  this._opBeforeFixups = null;
 }
 module.exports = SubmitRequest;
 
@@ -68,6 +70,12 @@ SubmitRequest.prototype.$fixup = function(op) {
       ERROR_CODE.ERR_TYPE_DOES_NOT_SUPPORT_COMPOSE,
       typeId + ' does not support compose'
     );
+  }
+
+  if (!this._opBeforeFixups) {
+    this._opBeforeFixups = this.op.create ?
+      {data: util.clone(this.op.create.data)} :
+      {op: util.clone(this.op.op)};
   }
 
   if (this.op.create) this.op.create.data = type.apply(this.op.create.data, op);
@@ -186,8 +194,6 @@ SubmitRequest.prototype.apply = function(callback) {
   // Always set the channels before each attempt to apply. If the channels are
   // modified in a middleware and we retry, we want to reset to a new array
   this.channels = this.backend.getChannels(this.collection, this.id);
-  this._fixupOps = [];
-  delete this.op.m.fixup;
 
   var request = this;
   this.backend.trigger(this.backend.MIDDLEWARE_ACTIONS.apply, this.agent, this, function(err) {
@@ -254,8 +260,19 @@ SubmitRequest.prototype.retry = function(callback) {
   if (this.maxRetries != null && this.retries > this.maxRetries) {
     return callback(this.maxRetriesError());
   }
+  this._resetFixups();
   this.backend.emit('timing', 'submit.retry', Date.now() - this.start, this);
   this.submit(callback);
+};
+
+// The fixups must be undone before the op is transformed forward again
+SubmitRequest.prototype._resetFixups = function() {
+  this._fixupOps = [];
+  delete this.op.m.fixup;
+  if (!this._opBeforeFixups) return;
+  if (this.op.create) this.op.create.data = this._opBeforeFixups.data;
+  else this.op.op = this._opBeforeFixups.op;
+  this._opBeforeFixups = null;
 };
 
 SubmitRequest.prototype._transformOp = function(ops) {
--- a/docs/middleware/op-submission.md
+++ b/docs/middleware/op-submission.md
@@ -65,6 +65,9 @@ backend.use('apply', (context, next) => {
 })
 ```
 
+{: .warn :}
+The `'apply'` hook may be triggered more than once for a single submission. If another client wins the race to commit, the op is transformed over the winning op and applied again to the newer snapshot, re-triggering the hook. Be careful, therefore, with any side effects that assume the hook only runs once per op.
+
 ### Commit
 
 The [`commit`]({{ site.baseurl }}{% link middleware/actions.md %}#commit) hook is triggered after the op has been applied to the snapshot in memory, and both the op and snapshot are about to be written to the database.
@@ -157,6 +160,9 @@ backend.use('apply', (request, next) => {
 
 {: .warn :}
 The `request.$fixup()` method may throw an error, which should be handled appropriately, usually by passing directly to the `next()` callback.
+
+{: .info :}
+Since the [`'apply'`](#apply) hook can be triggered more than once for a single submission, `request.$fixup()` may also be called more than once. Fixups from an abandoned attempt are discarded along with it, and the hook is expected to fix the op up again against the newer snapshot. Only the fixups from the attempt that is actually committed are sent to the client.
 
 ## Comparing old snapshot version with new version
 
--- a/test/middleware.js
+++ b/test/middleware.js
@@ -702,6 +702,79 @@ describe('middleware', function() {
       });
     });
 
+    it('applies the fixup once when the submit is retried', function(done) {
+      var remoteDoc = backend.connect().get('dogs', 'fido');
+
+      doc.submitOp([{p: ['fixups'], oi: 0}], function(error) {
+        if (error) return done(error);
+        remoteDoc.fetch(function(error) {
+          if (error) return done(error);
+
+          backend.use('apply', function(request, next) {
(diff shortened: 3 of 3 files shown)


lib/submit-request.js
```diff
@@ -70,6 +72,10 @@ SubmitRequest.prototype.$fixup = function(op) {
     );
   }
 
+  if (!this._fixupOps.length) {
```
alecgibson: I think this reads a bit weird: I guess it's trying to run only the first time `$fixup` is called, but it feels like it would be more idiomatic to check for the truthiness of `this._opBeforeFixups` (and make sure we unset it in `_restoreOpBeforeFixups()`).

lib/submit-request.js
```diff
@@ -254,10 +260,19 @@ SubmitRequest.prototype.retry = function(callback) {
   if (this.maxRetries != null && this.retries > this.maxRetries) {
     return callback(this.maxRetriesError());
   }
+  this._restoreOpBeforeFixups();
```
alecgibson: I think it would be neater if this lived alongside the place where we set `_fixupOps = []`

dawidreedsy (approved):

